PR #2, opened 2018-12-13
Add iterator over Message records

vavrusa:
Iterating through records is tedious with the current interface, as it requires
descending and checking each section individually. This change adds an
iterator interface over Message that scans through all sections in sequence.

### DIFF
--- a/domain-core/src/bits/message.rs
+++ b/domain-core/src/bits/message.rs
@@ -260,6 +260,14 @@ impl Message {
         let additional = authority.clone().next_section()?.unwrap();
         Ok((question, answer, authority, additional))
     }
+
+    /// Returns record iterator for all sections
+    pub fn iter(&self) -> MessageIterator {
+        match self.answer() {
+            Ok(section) => MessageIterator { inner: Some(section) },
+            Err(_) => MessageIterator { inner: None },
+        }
+    }
 }
 
 
@@ -373,6 +381,36 @@ impl AsRef<[u8]> for Message {
     }
 }
 
+//--- Iterator
+
+pub struct MessageIterator {
+    inner: Option<RecordSection>,
+}
+
+impl Iterator for MessageIterator {
+    type Item = (Result<ParsedRecord, ParsedDnameError>, Section);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // Try to get next record from current section
+        match self.inner {
+            Some(ref mut inner) => {
+                let item = inner.next();
+                if let Some(item) = item {
+                    return Some((item, inner.section));
+                }
+            },
+            None => return None,
+        }
+
+        // Advance to next section if possible, and retry
+        self.inner = match self.inner.clone().unwrap().next_section() {
+            Ok(section) => section,
+            Err(_) => None,
+        };
+
+        self.next()
+    }
+}
 
 //------------ QuestionSection ----------------------------------------------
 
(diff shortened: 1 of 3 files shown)


vavrusa:
Yeah, so my use case is I pick up messages from cache, go through records in the cached message, update TTL and stuff and and insert them into response.

There's two parts:

1. Iterator for Message to go through all the records. I think this is fairly straightforward as there's no way to iterate it now without going through sections manually (the method returning all sections returns a tuple that is not iterable).
2. RecordSectionBuilder interface that accepts not only record but also section, but requires the sections to be written in order.

I agree that it would be nice to provide an interface for writing sections out of order. I added traits with default methods, so implementing a buffered RecordSectionBuilder should be relatively easy (StreamBuilder is not that long), it would need to reimplement the push / freeze methods to serialize the buffered records. Originally I wanted to add something like "next_section" to RecordSectionBuilder, but I'm not sure how to do that as you can't write something like `fn next_section(self) -> impl Option<RecordSectionBuilder>` in the trait, so I went with a stream builder which interface allows you write records out of order, but it will trigger an error, so it's prevented. Do you have any good idea how to solve it?

partim:
If you don't need to shift records between sections and don't need to add records to the sections either, how about a method a la

```
impl MessageBuilder {
    fn add_records(self, message: &Message, op: F) -> Result<AdditionalBuilder, ShortBuf>
    where F: Fn(record: Record) -> Option<Record> {
        unimplemented!()
    }
}
```

I.e., you provide a closure which receives a copy of each record in the message, decides whether to keep it and if so can manipulate the record into whatever shape it wants before returning it. We could provide an extra section argument to the closure if that is relevant for the decision.

The method would be called after creating the question section, add all the records returned by the closure to the same section they were in in the message and end up with an `AdditionalBuilder`, so you could still manipulate the header (we might need an FnMut closure for that so you can update some outside variables from the closure, but I am not quite sure whether that is actually necessary).

vavrusa:
Yeah, that would work as well. That's basically what the caller loop is doing, but with you're moving the loop inside the function and adding a callback for each record. It's less powerful because it requires the source to be Message and not any stream of records, but maybe that's better. If you'd rather have this, I can replace `stream()` with it.

partim:
I was thinking about the fact that in an in-memory cache you probably don't want to store the record set as a `Message` but in its decoded form. So maybe it would be better to have this method on message? Then you could have some other type that is basically just three `Vec<Record>`s implement its version of that method for a consistent interface? I think this way it would be easier to implement than with a method on `MessageBuilder` and a trait for which making a safe definition would be tricky.

vavrusa:
> I was thinking about the fact that in an in-memory cache you probably don't want to store the record set as a Message but in its decoded form. So maybe it would be better to have this method on message?

Can you give me an example? I'm confused since Message doesn't mutate the underlying bytes, or do you mean something like `Message::from_records(...) -> Self` ?

vavrusa:
I see, this would copy just records, not the whole message (header, question) right?

partim:
Yes. I believe that’s the right thing what with an answer needing the question copied from the query. Plus, since you’ll end up with an `AdditionalBuilder`, you can also add an OPT record etc.

partim:
@vavrusa, do you want to take a stab at this or should I?

vavrusa:
I reverted the `StreamBuilder` commit and added `copy_records()` method on Message. Feel free to change, it's a bit verbose, but I tried to avoid macros.

partim:
Looks good! Once again, thanks for the PR!

PR #6, opened 2019-02-08
Fix for unwrap in copy_records and a convenience method for OPT

vavrusa:


partim:
Thanks for the PR! It looks good to me.

PR #17, opened 2019-04-04
bits::message_builder bugfix and optimization

vavrusa:
The 58ecd8b fixes appending to additional section.
The 89ef001 allows pushing records by reference.

### DIFF
--- a/domain-core/Cargo.toml
+++ b/domain-core/Cargo.toml
@@ -14,12 +14,18 @@ license = "BSD-3-Clause"
 name = "domain_core"
 path = "src/lib.rs"
 
+[features]
+tsig = ["ring"]
+default = ["tsig"]
+
 [dependencies]
 bytes          = "^0.4"
 chrono         = "^0.4"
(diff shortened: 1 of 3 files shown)


partim:
Thanks for the PR!

I kind of want to redesign the message builder on one of the next versions, now that we have a bit more experience with it. One idea for this particular issue is to create a pair of traits `RecordBuilder` and `RecordDataBuilder` to cover all the cases with a single `push` method.

We could do that now or leave it for later. I am fine with just adding `push_ref` for now.

vavrusa:
@partim I forgot the comment, sorry! Adding a different trait works too, it'd be probably nicer, but perhaps it's not as obvious what's happening when you write something like `push(rr.into())` vs `push_ref(&rr)`. Perhaps we could have two different message builders, one buffered (that allows pushing in any order) and one unbuffered (that serializes directly to wire), each using a different trait (as the buffered builder needs to own the record anyway).

PR #21, opened 2019-05-05
WIP: Expose a new option for setting (non) recursive queries

hawkinsw:
Interested in your thoughts on this. This adds a feature that lets the user specify whether they want to do a recursive or iterative query. I am still new to rust and using this as a learning exercise so I want to hear your feedback on making this patch better. I hope that this is a feature that seems worthwhile. It's something that I needed on a teaching application!

Thanks for making a great, usable library!

### DIFF
--- a/domain-resolv/src/lookup/host.rs
+++ b/domain-resolv/src/lookup/host.rs
@@ -27,11 +27,12 @@ use crate::resolver::{Resolver, SearchNames};
 /// return the canonical name.
 pub fn lookup_host<R: Resolver, N: ToDname>(
     resolver: &R,
-    name: &N
+    name: &N,
+		recursive: bool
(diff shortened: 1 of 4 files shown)


partim:
Thank you for the pull request!

In libc’s stub resolver, whether or not to set the RD bit can be selected via the `recurse` option in the resolver configuration. We indeed also have `domain_resolv::stub::conf::ResolvOptions::recurse` for this purpose. Apparently, I forgot to pick this up when creating the query, but that can easily be fixed.

Do you have a use case for selecting the RD bit on a per-query basis or would it be fine to do this when you create the resolver? If that is fine, I’d rather go that route as it seems cleaner to me. Happy to be convinces otherwise, though!

PR #25, opened 2019-05-20
Support dnssec query in stub resolver

vendemiat:


partim:
Thank you for the pull request!

I’m afraid I have a few notes, though.

For one, I think the AD bit and DO bits should be set independently. In addition, given that the DO bit requires EDNS support, if it is requested, servers that are known to not support ENDS should probably be skipped.

Independently of that, I don’t think the `use_dnssec` flag shouldn’t be in `QueryBuilder` as that is supposed to be generic. Instead, maybe `ServerInfo::prepare_message` should take a `&ResolvOptions` as an additional parameter.

vendemiat:
Thanks for the feedback, I will try to rework it. 

Although I would argue that we should export only one dnssec flag for the user. Furthermore this patch only set DO bit when EDNS is enabled for the server.

PR #26, opened 2019-05-20
Fix panic in stubresolver when UDP ServerList is empty

vendemiat:
'attempt to calculate the remainder with a divisor of zero', domain-resolv/src/stub/net/mod.rs:202


partim:
Thanks for the pull request!

I was working with the assumption that the server list will never be empty, but that may perhaps be a bad assumption so being safe is a good idea.

I will first merge #27 in order to deal with the Travis failure on nightly and then merge this PR.

PR #28, opened 2019-05-21
rdata/rfc4034: add support for calculating DNSKEY digest

vavrusa:
This commit implements necessary interfaces for RFC4034, 5.2:

```
5.2.  Processing of DS RRs When Validating Responses

   The DS RR links the authentication chain across zone boundaries, so
   the DS RR requires extra care in processing.  The DNSKEY RR referred
   to in the DS RR MUST be a DNSSEC zone key.  The DNSKEY RR Flags MUST
   have Flags bit 7 set.  If the DNSKEY flags do not indicate a DNSSEC
   zone key, the DS RR (and the DNSKEY RR it references) MUST NOT be
   used in the validation process.
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,5 @@
 [workspace]
 members = ["domain", "domain-core", "domain-resolv", "domain-sign", "domain-tsig", "domain-validate", "interop"]
 
+[patch.crates-io]
+ring = { git = "https://github.com/andrewtj/ring.git", rev = "eedb0514fb73e1034eefbec10140af8a51effff3"}
--- a/domain-core/Cargo.toml
+++ b/domain-core/Cargo.toml
@@ -22,4 +22,3 @@ derive_more    = "^0.14"
 rand           = "^0.6"
 unwrap         = "^1.2"
 void           = "^1.0"
-
--- a/domain-core/Cargo.toml
+++ b/domain-core/Cargo.toml
@@ -22,3 +22,4 @@ derive_more    = "^0.14"
 rand           = "^0.6"
 unwrap         = "^1.2"
 void           = "^1.0"
+
--- a/domain-core/Cargo.toml
+++ b/domain-core/Cargo.toml
@@ -20,5 +20,6 @@ bytes          = "^0.4"
 chrono         = "^0.4"
 derive_more    = "^0.14"
 rand           = "^0.6"
+unwrap         = "^1.2"
 void           = "^1.0"
 
--- a/domain-core/Changelog.md
+++ b/domain-core/Changelog.md
@@ -15,14 +15,26 @@ Breaking Changes
 
 New
 
-* `bits::message::Message::opt` returns a messages OPT record if present.
+* `message::Message::opt` returns a messages OPT record if present.
   ([#6], thanks to Marek Vavruša!)
 * unsafe `bits::name::Dname::from_bytes_unchecked` in order to create
   names from well-known sequences. [(#31)]
+* `compose::Compose::compose_canonical` for composing the canonical form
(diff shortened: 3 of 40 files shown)


partim:
Thanks for the pull request!

I had the intention to move everything DNSSEC related (i.e., signing and validation) into a separate crate. The main reason was to keep `domain-core` free of features. Since it is a rather fundamental crate, it may be pulled in at several places and doing so with differing feature sets might be confusing (or may even lead to breakage, not sure).

Would you mind if I started `domain-dnssec` and we’d move your code into a validation module there? (I had intentions to do that soon anyway to start working on a signer.)

vavrusa:
Yeah, sounds good! My thinking was to keep primitives (digest calculation and rrsig verification) close to their respective types for convenience, and move higher-level logic to a separate crate, but perhaps it can all be in a separate crate.

I have one more change related to this that adds Ord and changes PartialOrd of RDATA to RFC4034 canonical instead of arbitrary to facilitate sorting of RRSETs for validation later. Currently Soa and Rrsig RDATA is ordered wrong because of Serial ordering semantics. So I'll hold off then until you add a separate crate and split it into core changes + dnssec changes.

partim:
Sorry for the long silence. I was busy with other stuff.

I have now created two separate crates, `domain-sign` and `domain-validate`, for DNSSEC. The reasoning behind splitting them was that for validation we can perhaps get away with just _ring_ while for signing we may need to support a generic signer interface (something, something, HSM).

I think most of the code can remain in `domain-core`, though. Unless I missed something, only `DnsKey::digest` and `Rsig::verify_signed_data` need to move. My proposal would be to add `DnsKeyExt` and `RsigExt` traits in `domain-validate` that have these methods.

partim:
Oh, and the tests fail because it wants to compile your quotes from the RFC. I think you need to stick a `txt` after the opening triple-backticks or possibly `ignore`.

vavrusa:
> I think most of the code can remain in domain-core, though. Unless I missed something, only DnsKey::digest and Rsig::verify_signed_data need to move. My proposal would be to add DnsKeyExt and RsigExt traits in domain-validate that have these methods.

Yeah, sounds good! I'm still testing things, when it's about done I'll refactor the PR to split out the proposed traits and fix up the doc tests.

partim:
I pushed the first version of my signing branch. There’s some overlap with your work. I should probably have mentioned these earlier, but it took me quite some time to figure the design out. My apologies for that!

I think the most important thing is the new trait `domain-core::cmp::CanonicalOrd` which orders record data and records in canonical order. Additionally, I added `Compose::compose_canonical` for creating the data that goes into the signatures.

vavrusa:
Alright! I'll try to rebase this and put the ext traits for Ds and Rrsig to domain-validate. I think we'll also need to bump the ring version to at least 0.15alpha as the API has changed from 0.14. We'll need 0.15 because of https://github.com/briansmith/ring/issues/870 (I currently use my fork with 0.14 but added support for 1024b RSA keys, but that's not sustainable).

partim:
I guess this is ready to be merged, then. There’s going to be some reshuffling `domain-sign` as it moves along further, but that can happen in follow-up PRs.

Just to double-check: Are you okay with merging?

vavrusa:
I think we can merge it and start separate PRs for followup work if you're okay with the updated ring version. I hope that by the time domain gets a new release tag, ring will have a new crate published.

(I noticed there's an abstraction for sorted record sets, which might be useful for validation too)

partim:
I am considering making the abstractions for sorted records, record families (i.e., all records for a name/class pair), and all RRsets into traits since an authoritative may want to keep them in a tree rather than a vec. But let’s see where this all goes before making decisions.

PR #38, opened 2019-11-04
Fixed crash in NSEC parser, infinite loop in CNAME unrolling, various validator bugfixes

vavrusa:


partim:
Apologies for the delayed processing of the PR. I have been busy on getting what is now #39 into a shape that works and is usable. I will have merge all your changes into that as well – and wanted to get to a point where I can do that before merging the PR so that nothing gets lost. I shall do that within the next few days and then merge the PR.

Unless you expect to have more commits that you would want to process via this one?

vavrusa:
@partim no problem, thanks for update! I don't expect to push anything soon except maybe random bugfixes, in any case I'll check if there's any commits to master and rebase.

vavrusa:
There's one change that will cause a conflict - I changed some methods like `inc_*count()` to return a Result instead of panicking, which makes it easy to propagate the error to public API for pushing records as opposed to checking header counts before each record push and catching panics. Maybe also adding an implicit MessageBuilder size limit of 64k would be a good idea.

partim:
The message builder has been redesigned for #39. You can now choose whatever type you want to build your message into – that type can then enforce any kind of restrictions. Have a [look at the branch](https://github.com/NLnetLabs/domain/blob/generic-bits/domain-core/src/message_builder.rs). It currently lacks compression, I’ll add that back in again soon.

One request for the future, if I may: Maybe you can place proposals for changing APIs in separate PRs? It makes it easier to track when stuff changes and also makes sure that we can properly review them.

vavrusa:
> The message builder has been redesigned for #39. You can now choose whatever type you want to build your message into – that type can then enforce any kind of restrictions.

That's awesome!

> One request for the future, if I may: Maybe you can place proposals for changing APIs in separate PRs?

Yeah totally reasonable, I used my master branch for PR and then added up bunch of stuff over time. I'm happy to split it into separate PRs when you're ready to review.

partim:
> I'm happy to split it into separate PRs when you're ready to review.

I think we can leave this one as it is. The transition from 0.4 to 0.5 will be a bit messy, anyway. I wonder, though, if it is worthwhile to port your bug fixes back to 0.4 and release 0.4.1 before moving on.

vavrusa:
I added another bugfix in 19c1346

vavrusa:
Added another bugfix in 0553b4f. @partim do you have any plans for fuzzing? The NSEC/NSEC3 parsing code seems like a good candidate, I can add a few fuzzer harnesses if you let me know how would you like to structure it etc.

partim:
I haven’t thought about fuzzing at all. Not sure what a good structure is, either. So maybe just go ahead if you want and we’ll figure it out as we go?

vavrusa:
@partim related to e28478e, do you want to rather deprecate WKS instead of fixing it up? (it's likely going to be [obsoleted soon](https://tools.ietf.org/html/draft-sury-deprecate-obsolete-resource-records-00))

partim:
That’s probably a good idea. The representation format parser always annoyed me because it needs access to `/etc/services`.

vavrusa:
@partim deprecated the WKS in 9ea3f43

vavrusa:
Also added two more commits to fix parsing OPT options.

partim:
@vavrusa, how do you want to proceed with getting your changes into the new 0.5 world? I am happy to go over your commits and add them to #39. Alternatively we could also merge #39 and then either rework this PR or make a new one (which might be easier than fixing all the conflicts).

An option is also to merge this PR into a new `series-0.4` branch as the last changes to that series and make a final round of releases so that code on 0.4 at least receives all your fixes.

vavrusa:
> I am happy to go over your commits and add them to #39.
I think commit-by-commit is going to be easiest. 

> An option is also to merge this PR into a new series-0.4 branch
Yes that sounds good 👍

partim:
For porting to series-0.5, I was wondering what best to do regarding “infinite loop in `Message::canonical_name()`”. Instead of using a hash map (which we can’t in no-std world anyway), perhaps a simple counter and breaking out if it reaches 10 or so would be good enough? If you prefer the hash map approach, maybe we can have it as a separate method to make clear that it is rather more expensive?

vavrusa:
Counting is feasible, 10 will trigger some false positives. 50 is safer. At some point checking N names is going to be more expensive than checking a hashset once, but I don't think this code path performance is critical enough to warrant an alternative implementation (or at least yet).

partim:
It just occurred to me that since we are looking at an individual message, the ANCOUNT is a perfect, easy to retrieve upper limit.

vavrusa:
Yeah that's true, although a query or response can be crafted in a way that ancount is 64k but there's only a few records and you wouldn't know before parsing them all.

vavrusa:
I'm going to keep the branch open until https://github.com/briansmith/ring/pull/952 is merged to unblock the validator.

partim:
It just occurred to me that this PR contains quite a few breaking changes, so we can’t really turn it into another 0.4 series release. We could apply the bug fix commits, though, in a separate PR?

partim:
I believe everything from the original PR was handled elsewhere and this should have been closed a long time ago …

PR #39, opened 2019-11-18
Make core DNS types generic over the type of octet sequences.

partim:
This PR heavily modifies the API of `domain-core` to let all types dealing with variable length binary data be generic over how that data is stored internally. Currently, we use `bytes::Bytes` everywhere but this depends on having the `alloc` crate. With this modification, we can also operate on stack-allocated fixed-size buffers.

The PR is currently work-in-progress with the following things left to do:

* [x] convert master file parsing to the new world,
* [x] convert `domain-resolv` to the new core (this one also serves as the proof of concept),
* [x] convert `domain-tsig` to the new core,
* [x] convert `domain-sign` to the new core,
* [x] convert `domain-validate` to the new core,
* [x] document the concepts and traits for generic octet sequences,
* [ ] update the documentation of `domain-core`.

Feedback etc. is, of course, welcome already!


vavrusa:
Hi 👋 any progress here? I wanted to upgrade bytes crate to 0.5 as tokio and most of newer crates are using it and it makes it incompatible with Message requiring a 0.4 bytes::Bytes, but seems like it would clash with changes in this PR so I'm holding off.

partim:
After a bunch of other deadlines and vacations, I actually started working on this again this week. If there’s no sudden pitfalls, the code should be complete by the end of next week. We can than merge it and do the documentation separately.

vavrusa:
That's great! Thanks for the update, I'll keep an eye on it and rebase my PR when it's merged. I can probably help to port domain-validate over if there's any changes necessary.

partim:
I’ve converted `domain-validate` as well. The only relevant difference should be that the methods that actually compose some data now can fail since they are generic over the octet builder type. The tests seem to be succeeding.

Note that this is still based on the version before #38. I’ll merge that in soon.

partim:
Note: domain-resolv is still on tokio 0.1. This will be fixed in a separate PR after this one got merged.

partim:
I am going to merge this PR into the new `series-0.5` branch which will receive further changes until the structure and API has stabilized and can be merged into master.

partim:
I’ll leave this as a note here: The `series-0.5` branch should now be fairly stable for everything except the resolver which I am currently converting to new futures. Unlike how it appears in the branch, it will also remain in `domain-resolv` for now, since you can’t do async functions in a no-std crate.

PR #42, opened 2020-03-19
Port changes from #38 to series-0.5.

partim:
This PR takes all the changes proposed in #38 and ports them to the brave new world of series-0.5.

I’ve merged some commits. Some other commits have already been considered during the initial refactoring for 0.5. A few will be discussed on #38 before coming here.


vavrusa:
I see you squashed most of the commits, is there a way to check which commits are applied just to make sure nothing was forgotten?

vavrusa:
I'm going to start testing the 0.5 branch now.

partim:
I actually re-implemented your changes atop the new branch. Seemed easier. I believe I included all of your changes.

There is one more breaking change coming up with the resolver re-implementation: I made the `OptBuilder` take a `&mut AdditionalBuilder<_>` and only accessible via a closure. Makes it easier to manipulate only the OPT record on an existing message. I should get to commit this tomorrow.

vavrusa:
Ok, makes sense. Are you planning to add something like `prelude::*`? It's maybe a bit cumbersome now as `domain_core::Message` becomes `domain::base::message::Message` and the traits are not all in one place anymore.

partim:
There used to be re-exports at the `core` (now `base`) level. We should probably bring those back, i.e., we’d have `domain::base::Message`.

Personally, I prefer explicit imports, but I don’t mind making a prelude for people who prefer that.

vavrusa:
> There used to be re-exports at the core (now base) level. We should probably bring those back

That's a good point, I used it like this on the 0.4 branch. If
`domain::base` reexports common types and traits that's perfect for
me.

partim:
I’ll make a PR for this tomorrow, too.

PR #45, opened 2020-03-23
Changes for 0.5.

partim:
This PR contains the major changes for the upcoming 0.5.0 release of both `domain` and `domain-resolv`.


partim:
Before merging, I want to do one more change to `domain-resolv`, which is to swap out `io::Error` for dedicated query and search error types. Main reason is that I want to have an easy and clear way to signal ServFail.

partim:
Ah feck, now I actually merge this into master via a different PR. Should be fine though. Closing this one, then.

PR #57, opened 2020-04-03
rdata/rfc1035: fixed TxtBuilder

vavrusa:
The TxtBuilder::finish() used remaining free space to calculate size instead of the last slice size.


vavrusa:
Congrats on the new version! Do you use a custom `rustfmt.toml`?

partim:
Thanks for the fix!

I don’t have a custom `rustfmt.toml` yet; should probably look into that. I am rather relaxed about formatting. I only have one hard rule: line length is 78 characters.

vavrusa:
It makes it easier to fix up formatting before commit with `cargo fmt`. Right now it generates a lot of diffs.

PR #58, opened 2020-04-16
rdata/rfc6672: added Dname type support

vavrusa:


domain/src/rdata/rfc1035.rs
```diff
@@ -32,6 +32,7 @@ use crate::base::serial::Serial;
 ///
 /// Implements some basic methods plus the `RecordData`, `FlatRecordData`,
 /// and `Display` traits.
+#[macro_export]
```
partim, line 35: Instead of exporting the macro here (because I think that makes it public), perhaps move it to `rdata::macros`? It’s the place for module-wide macros …

domain/src/rdata/rfc6672.rs
```diff
@@ -0,0 +1,41 @@
+use crate::base::cmp::CanonicalOrd;
+use crate::base::iana::Rtype;
+use crate::base::name::{ParsedDname, ToDname};
+use crate::base::octets::{Compose, OctetsBuilder, OctetsRef, Parse, ParseError, Parser, ShortBuf};
+use crate::base::rdata::RtypeRecordData;
+#[cfg(feature = "master")]
+use crate::master::scan::{CharSource, Scan, ScanError, Scanner, SyntaxError};
+#[cfg(feature = "master")]
```
partim, line 8: This should probably be for feature `"bytes"`.

domain/src/rdata/rfc1035.rs
vavrusa: Ah you're right, I'll move it to macros!

PR #63, opened 2020-06-17
bump the revision number

vavrusa:


partim:
I tend to only use the ‘proper’ version number when doing an actual release. Is there a reason why the -bis thing doesn’t work for you?

That said, while there isn’t many changes, I’m happy to do a 0.5.2 if this helps you.

vavrusa:
Yeah, I would love to just put a package version in Cargo, instead of `[patch.crates-io]` to specific git revision, but I can wait until you're ready to do a release!

partim:
Version numbers are free, so I don’t mind making a single-issue release. But just in case: do you have any upcoming additions we might throw in?

vavrusa:
Not right now, I'm converting parts of some libraries to the new domain so I'll submit PRs as I test/discover problems.

partim:
I’ll make a proper release, then, via a separate PR – and close this one.

PR #65, opened 2020-06-29
Fix StubResolver::run and add example.

partim:


glts:
@partim Thank you. The example works, but replacing the line

```
stub.query((name, rtype)).await
```

with

```
stub.lookup_addr(Ipv4Addr::new(1, 2, 3, 4).into()).await
```

brings back the compilation error. `lookup_addr` should work this way, shouldn’t it?

partim:
I see.

Removing the `R: Send + 'static` bounds on `StubResolver::run` solves the issue but there’s some new lifetime shenanigans. This may need some re-thinking to work.

In the mean time, you can do what `StubResolver::run_with_conf` does manually:

```
    let stub = StubResolver::new();
    let mut runtime = tokio::runtime::Builder::new()
        .basic_scheduler()
        .enable_all()
        .build().unwrap();
    
    let _res = runtime.block_on(
        stub.lookup_addr(Ipv4Addr::new(1, 2, 3, 4).into())
    );
```

glts:
No worries, I found that `query` covers all of my requirements, everything is working perfectly. Thanks again.

partim:
Good to hear it’s not too convoluted! I’ll still look into making the `run` shortcut work without having to resort to async for the most common cases – that would make it more useful.

glts:
@partim May I ask: is `StubResolver::run` appropriate (ie cheap to call) in a long-running application that does repeated DNS queries? Or would I anyway have to create and manage my own runtime instance, and use `block_on` as you showed above?

And yes, support for synchronous operation is certainly welcome by people like me who don’t (yet) need the async capability.

partim:
Under the hood, it does exactly what the code snippet does, i.e., create a runtime, a stub resolver, and a future from the provided closure and then block on it.

The reason is that in many cases – even relatively simple ones like `lookup_host` – multiple queries need to be fired off in parallel which means some kind of async processing is necessary.

Issue #68, opened 2020-07-22
Random things discovered when porting to domain=0.5


vavrusa:
Hi, I ended up adding a bunch of helper functions over time as I was updating to the 0.5 domain library, and I thought some of it might be useful in the base library itself. Maybe there's a nicer way of doing this stuff, so I thought I would open this ticket first to see what would make sense / when there's a better way to do things. I'm happy to open PRs for things you thing would be good to add.

1. There are some convenience functions missing for converting names and records from one octets to another, so I created a bunch of helpers:

```
/// Creates a root domain over any octets.
pub fn root_name<O>() -> Result<Dname<O>> where O: FromBuilder;

/// Convert name to canonical format (equivalent to "to_dname()" but also converts to canonical format)
pub fn to_canonical_name<O>(name: impl ToDname) -> Dname<O> where O: FromBuilder;

/// Convert OctetsRef into owned Octets.
pub fn to_octets<O, R>(o: R) -> Result<O> O: FromBuilder;
```

2. Some RecordData types (I think Ds or Dnskey) have "convert()" method which makes this easier, but most don't so they need to be recreated. This is useful when for example recreating trust anchors from static data into a Record backed by any other storage type, or storing records from `Message<&[u8]>` in cache with a  different storage type. I've added this convenience conversion function like this, but it's still a bit tedious. Perhaps some sort of "convert()" method would be useful for all types?

```
/// Convert Record into another Record with different RDATA.
pub fn from_record<N: ToDname, O: FromBuilder, D, DD, F: FnOnce(D) -> Result<DD>>(
    rr: Record<N, D>,
    f: F,
) -> Result<Record<Dname<O>, DD>>;
``` 

3. In most projects I have a need to print or debug messages in some sort of text format. I've implemented an extension trait for various wrapper structures that implement `std::fmt::Display`. Perhaps that would be useful in the domain library as well?

```
/// Extension for `domain::Message` for pretty printing.
pub trait MessageExt<O> {
    /// Formats the message question as a shorthand "<query name> <query_type>".
    fn as_formatted_query(&self) -> FmtQuery<O>;
    /// Formats the message and records using a dig-like text format.
    fn as_formatted_text(&self) -> FmtMessageText<O>;
    /// Formats the message and records using a short text format.
    fn as_formatted_short(&self) -> FmtMessageShort<O>;
}
```

vavrusa:
[...]

PR #69, opened 2020-08-07
domain/octets: implement Convert trait to allow converting octet types

vavrusa:
This trait forms a basis for converting structured types built on top
of octets such as CharStr, Dname, RDATA, and Record types.

#68

### DIFF
--- a/src/base/charstr.rs
+++ b/src/base/charstr.rs
@@ -33,7 +33,7 @@ use core::{cmp, fmt, hash, ops, str};
 };
 use super::cmp::CanonicalOrd;
 use super::octets::{
-    Compose, Convert, EmptyBuilder, FromBuilder, IntoBuilder, OctetsBuilder, OctetsRef,
+    Compose, ConvertOctets, EmptyBuilder, FromBuilder, IntoBuilder, OctetsBuilder, OctetsRef,
     Parse, ParseError, Parser, ShortBuf
 };
 use super::str::{BadSymbol, Symbol, SymbolError};
@@ -163,10 +163,10 @@ impl CharStr<[u8]> {
 }
 
 
-//--- Convert
+//--- ConvertOctets
 
-impl<Octets, Other> Convert<CharStr<Other>> for CharStr<Octets>
-where Octets: AsRef<[u8]> + Convert<Other> {
+impl<Octets, Other> ConvertOctets<CharStr<Other>> for CharStr<Octets>
+where Octets: AsRef<[u8]> + ConvertOctets<Other> {
     fn convert(&self) -> Result<CharStr<Other>, ShortBuf> {
         unsafe { Ok(CharStr::from_octets_unchecked(self.0.convert()?)) }
     }
--- a/src/base/charstr.rs
+++ b/src/base/charstr.rs
@@ -33,7 +33,7 @@ use core::{cmp, fmt, hash, ops, str};
 };
 use super::cmp::CanonicalOrd;
 use super::octets::{
-    Compose, EmptyBuilder, FromBuilder, IntoBuilder, OctetsBuilder, OctetsRef,
+    Compose, Convert, EmptyBuilder, FromBuilder, IntoBuilder, OctetsBuilder, OctetsRef,
     Parse, ParseError, Parser, ShortBuf
 };
 use super::str::{BadSymbol, Symbol, SymbolError};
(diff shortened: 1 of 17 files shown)


partim:
Could we get away with using `TryFrom` with references as sources instead of having a new trait? The trait bounds might get a bit messy, but so far all those mad bounds for octets types have been surprisingly invisible for users.

vavrusa:
I can add it as a convenience trait, would that work for you? Having the implementation for octets conversion in a specific method is good as it's clear in what it does (octets conversion), whereas TryFrom is a generic conversion from anything to anything.

partim:
After thinking about it properly (I blame the heat), you are, of course, correct that we want a dedicated trait.

I wonder now, though, if the blanket impl under certain conditions is a good idea or if it (because of specialisation rules) may prohibit optimizations for certain pairs or even block out-of-crate implementation completely.

vavrusa:
Right, this:

```
impl<Octets, Target> Convert<Target> for Octets
where
    Octets: AsRef<[u8]>,
    Target: FromBuilder,
    <Target as FromBuilder>::Builder: EmptyBuilder
{
    fn convert(&self) -> Result<Target, ShortBuf> {
        self.as_ref().to_octets()
    }
}
```

I'm not sure how to implement it better. Implement it for combinations of &[u8], Cow<&[u8]>, Vec<u8>, Bytes<u8>, BytesMut<u8>?

partim:
Implementing it specifically for each target type but if possible using generic sources might work best? I.e., have a specific impl for `&[u8]` that accepts anything that is `AsRef<[u8]>` and so on.

It also now occurs to me that the name of the trait should probably be more specific, e.g. `ConvertOctets`, since traits are normally imported by name only and not qualified by their module. So I think we should indicate in the name itself that the trait is specific for converting octet sequences.

partim:
So turns out I was confused about the direction of `ConvertOctets` in my earlier remark. That's a good indication that the name is bad. Perhaps we should call it `IntoOctets` and also have a `FromOctets`?

Looking at your impls for `Dname<_>`, I wonder if we should drop the trait bounds from the type definition. That would make things like `IntoOctets` (and probably more trait impls) much more compact.

vavrusa:
What do you want to do about existing `IntoOctets` trait?

partim:
Sorry, I should look at these things properly before responding.

Do we need to have the current `IntoOctets` as a separate trait from `OctetsBuilder`. Asked differently, is there a use case for an octets builder that cannot impl current `IntoOctets` for any reason? All the octets builders we have right now also implement `IntoOctets`.

So, unless there is a reason to keep this split, I think merging current `IntoOctets` into `OctetsBuilder` and then reusing the name for a new `IntoOctets` trait might be a good path forward. I think this will also bring the interface more in line with common use where the Into traits carry their target as a type parameter and not an associated type.

The merging should probably be in a separate PR. I'm happy to do that if you agree with this approach.

vavrusa:
Ok, that sounds reasonable.

partim:
I’ve merged the trait in #75.

partim:
That last commit should probably be in its own PR.

Since this is all a bit messy now, do you want to continue with this PR or do you want me to take it over?

vavrusa:
Ah sorry this is from my main branch, let me open a separate PR for the bugfixes. If you have time to finish the Convert trait that'd be great, I can get back to it next week otherwise.

vavrusa:
I moved bugfixes to https://github.com/NLnetLabs/domain/pull/76 and rebased this to master.

partim:
I’ll try to get a new PR for the conversion traits submitted by tomorrow CET evening.

partim:
So I drafted a version that, similarly to the standard `From` and `Into` moves the sources into the target. That is fine for converting from a type atop a `Vec<u8>` to, say, `Bytes`, but we’d still need to provide a non-trait `for_ref` and `for_slice`. I don’t quite mind that, but does that work for your use cases?

I also just realized that `from_octets` is commonly used as the name of the associated function that creates a thing from a raw octets sequence. I don’t really want to change that (again), since it is all over the place. So we need a better name for those traits.

partim:
Maybe flip it around? `OctetsFrom` and `OctetsInto`? That would express that we are indeed only converting octets and nothing else?

vavrusa:
> So I drafted a version that, similarly to the standard From and Into moves the sources into the target.

It should be ok either way. I've added some test cases for conversions here so if those pass it'll likely work unless there's some complications propagating the new trait bounds.

> Maybe flip it around? OctetsFrom and OctetsInto?

It's ok, it might be confusing with `from()` and `from_octets()`. I kind of liked the `ConvertOctets` naming because it implies converting from one octets into another and not just from octets to something else. But it doesn't describe direction sufficiently.

partim:
I am not quite happy with the names either but I suspect there is no perfect solution here.

Meanwhile, I suggest we continue in PR #77.

partim:
This PR has been solved in #77 and I believe can be closed.

PR #77, opened 2020-09-09
Introduce OctetsFrom and OctetsInto.

partim:
This continues from PR #69.

I think I have implemented `OctetsFrom` for all relevant types with `OctetsInto` automatically available for them via a blanket impl. @vavrusa, can you have a look and let me know if this works for you or what needs changing?

### DIFF
--- a/src/base/octets.rs
+++ b/src/base/octets.rs
@@ -313,6 +313,86 @@ impl<'a, A: Array<Item = u8>> OctetsRef for &'a SmallVec<A> {
 }
 
 
+//------------ OctetsFrom ----------------------------------------------------
+
+/// Convert a type from one octets type to another.
+///
+/// This trait allows creating a value of a type that is generic over an
+/// octets sequence from an identical value using a different type of octets
+/// sequence.
+///
+/// This is different from just `From` in that the conversion may fail if the
+/// source sequence is longer than the space available for the target type.
+pub trait OctetsFrom<Source>: Sized {
+    /// Performs the conversion.
+    fn octets_from(source: Source) -> Result<Self, ShortBuf>;
+}
+
+
+impl<'a, Source: AsRef<[u8]> + 'a> OctetsFrom<&'a Source> for &'a [u8] {
+    fn octets_from(source: &'a Source) -> Result<Self, ShortBuf> {
+        Ok(source.as_ref())
+    }
+}
+
+#[cfg(feature = "std")]
+impl<Source> OctetsFrom<Source> for Vec<u8>
+where Self: From<Source> {
+    fn octets_from(source: Source) -> Result<Self, ShortBuf> {
+        Ok(From::from(source))
+    }
+}
+
+#[cfg(feature = "bytes")]
+impl<Source> OctetsFrom<Source> for Bytes
+where Self: From<Source> {
+    fn octets_from(source: Source) -> Result<Self, ShortBuf> {
+        Ok(From::from(source))
+    }
+}
+
+#[cfg(feature = "bytes")]
+impl<Source> OctetsFrom<Source> for BytesMut
+where Self: From<Source> {
+    fn octets_from(source: Source) -> Result<Self, ShortBuf> {
+        Ok(From::from(source))
+    }
+}
+
+#[cfg(features = "smallvec")]
+impl<Source, A> OctetsFrom<Source> for SmallVec<A>
+where Source: AsRef<u8>, A: Array<Item = u8> {
+    fn octets_from(source: Source) -> Result<Self, ShortBuf> {
+        Ok(smallvec::ToSmallVec::to_smallvec(source.as_ref()))
+    }
+}
+
+
+//------------ OctetsInto ----------------------------------------------------
+
+/// Convert a type from one octets type to another.
+///
+/// This trait allows trading in a value of a type that is generic over an
+/// octets sequence for an identical value using a different type of octets
+/// sequence.
+///
+/// This is different from just `Into` in that the conversion may fail if the
+/// source sequence is longer than the space available for the target type.
+///
+/// This trait has a blanket implementation for all pairs of types where 
+/// `OctetsFrom` has been implemented.
+pub trait OctetsInto<Target> {
+    /// Performs the conversion.
+    fn octets_into(self) -> Result<Target, ShortBuf>;
+}
+
+impl<Source, Target: OctetsFrom<Source>> OctetsInto<Target> for Source {
+    fn octets_into(self) -> Result<Target, ShortBuf> {
+        Target::octets_from(self)
+    }
+}
+
+
 //------------ OctetsBuilder -------------------------------------------------
 
 /// A buffer to construct an octet sequence.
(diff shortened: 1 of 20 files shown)


vavrusa:
Yeah! Can you add the tests from PR#69? It should make it more obvious whether it works or not. The implementation for Record is missing. I'm trying to figure out how the traits bound would look like now. One example - converting record with a DS into another record with generic octets.

Previously it would look something like this:

```
fn convert_ds<O>() -> Record<Dname<O>, Ds<O>> where
    O: FromBuilder,
    <O as FromBuilder>::Builder: EmptyBuilder {
    use domain::base::iana::{Class, SecAlg, DigestAlg};
let ds: Record<Dname<bytes::Bytes>, Ds<bytes::Bytes>> = Record::new(Dname::root_bytes(), Class::In, 0, Ds::new(20326, SecAlg::RsaSha256, DigestAlg::Sha256, hex::decode("E06D44B80B8F1D39A95C0B0D7C65D08458E880409BBC683457104237C7F8EC8D").unwrap().into()));

    ds.convert().unwrap()
}

fn test_convert() {
    let t1 = convert_ds::<Vec<u8>>();
    let t2 = convert_ds::<bytes::Bytes>();
}
```

Now I have:

```
fn convert_ds<'a, O>() -> Record<Dname<O>, Ds<O>> where
    O: FromBuilder + OctetsFrom<&'a [u8]>,
    <O as FromBuilder>::Builder: EmptyBuilder {
    use domain::base::iana::{Class, SecAlg, DigestAlg};
let ds: Record<Dname<bytes::Bytes>, Ds<bytes::Bytes>> = Record::new(Dname::root_bytes(), Class::In, 0, Ds::new(20326, SecAlg::RsaSha256, DigestAlg::Sha256, hex::decode("E06D44B80B8F1D39A95C0B0D7C65D08458E880409BBC683457104237C7F8EC8D").unwrap().into()));

    // This doesn't compile, it would need something like for_slice() to be able to convert Ds<Bytes> -> Ds<&[u8]> -> Ds<O>
    let rdata_slice: Ds<&[u8]> = ds.data().octets_into().unwrap();
    Record::new(ds.owner().for_slice().octets_into().unwrap(), Class::In, 0, rdata_slice.octets_into().unwrap())
}

fn test_convert() {
    let t1 = convert_ds::<Vec<u8>>();
    let t2 = convert_ds::<bytes::Bytes>();
}
```

partim:
I have added test to `rfc1035`. I think that’s the ones you had?

An impl for `Record` is present. I also added a test for a `Record<_, Ds<_>>` that seems to simply work. You don’t even need a separate function anymore, just call `octets_into` on your source record.

If you do want to use a function, I think just using `OctetsInto` or `OctetsFrom` (depending on your direction) should be enough.

partim:
One thing that I didn’t add is an impl on `ParsedDname`. The reason is that that type is atop a `OctetsRef`, not an octets sequence itself. Converting that would require to convert the octets sequence used by the parser underlying the parsed name, but then you likely end up with lifetime issues because `ParsedDname` may hold a reference to that underlying octets sequence.

But maybe this just works out and I worry too much?

partim:
@vavrusa, does this work for you now or is there still some usability issues to flesh out?

vavrusa:
Hi @partim sorry about the delay, I was distracted with other things. I'll make a note to try it out again this week. If I remember correctly I added Convert implementation on ParsedDname<O> that converted into Dname<OO>. The reason for that was so that I could convert Record<N, D> from Message into something I can store in cache. It's possible without it, just that I can't do `rr.into()` but have to recreate the record and convert owner using ToDname, and data using Convert.

src/base/octets.rs
```diff
@@ -313,6 +313,86 @@ impl<'a, A: Array<Item = u8>> OctetsRef for &'a SmallVec<A> {
...
+/// source sequence is longer than the space available for the target type.
+pub trait OctetsFrom<Source>: Sized {
+    /// Performs the conversion.
+    fn octets_from(source: Source) -> Result<Self, ShortBuf>;
+}
+
+
+impl<'a, Source: AsRef<[u8]> + 'a> OctetsFrom<&'a Source> for &'a [u8] {
+    fn octets_from(source: &'a Source) -> Result<Self, ShortBuf> {
+        Ok(source.as_ref())
```
vavrusa, line 334: This blanket implementation is not enough. Bytes for example only supports conversion from `&'static [u8]`. So converting from `&Record` now requires cloning the record and converting each member using the owned value. Perhaps the blanket implementation could use OctetsBuilder which can convert any `Source: AsRef<[u8]>` into `Octets`?

src/base/record.rs
```diff
@@ -182,6 +183,27 @@ impl<N, D> From<(N, u32, D)> for Record<N, D> {
...
+        source: Record<SrcName, SrcData>
+    ) -> Result<Self, ShortBuf> {
+        Ok(Record {
+            owner: Name::octets_from(source.owner)?,
+            class: source.class,
+            ttl: source.ttl,
+            data: Data::octets_from(source.data)?
+        })
+    }
+}
```
vavrusa, line 204: Maybe add an implementation for `N: SrcName` instead? SrcName should have `ToDname`, so converting an `ParsedRecord<impl OctetsRef>` into `Record<Dname<O>, ...>` should be possible.

src/base/octets.rs
partim: But isn’t that rather a missing `impl OctetsFrom<&'_ Record<...>> for Record<...>`? Or am I misunderstanding you?

src/base/record.rs
partim: I feel that this sort of thing should happen via regular `From` from a `ParsedDname` to its `Dname` given that it actually copies things around. The intention of `OctetsFrom` was really just a type conversion for the particular octet type.

src/base/octets.rs
vavrusa: Ah yes you're right, I misread this, I thought it was implemented for any `AsRef<[u8]>`.

src/base/record.rs
vavrusa: Yes that's fair, however scanning ParsedRecord and converting is common and currently very clunky.

```rust
let rr = rr.into_record::<Dnskey<_>>()?.context("empty record")?;
let rr: Record<Dname<O>, Dnskey<O>> = Record::new(rr.owner().to_dname().context("failed to convert dname")?, rr.class(), rr.ttl(), rr.into_data().octets_into().context("failed to convert data")?);
```

If we add `impl From<ParsedDname<R>> for Dname<O>` that should make `Record::from_record` work, another option is to move this into another method/trait.

partim: I think there indeed lots of `From` impls and other conversion helpers missing. My plan was to add them as they become necessary so as to not add too much stuff that nobody uses and that just increases the code size.

I suggest we look at conversions for records and names in a separate PR.

src/base/octets.rs
partim: It think when doing from a `&Record<O>` to a `Record<OO>`, we should go via `&O: OctetsRef`. So perhaps the trait bound should be `where &O: OctetsRef, OO: OctetsFrom<&O>` (plus lifetimes, of course). There might be a blanket impl in there somewhere, but it probably won’t work because of specialization rules.

Perhaps explore this in a different PR? Just so we get some progress here?

src/base/record.rs
vavrusa: Yes, of course, I didn't mean to hold this up, sorry.

partim: Nono! I really appreciate the feedback. Here is as good as anywhere.

PR #79, opened 2020-11-05
Add support for RFC 8914

xofyarg:
This change adds Extended DNS Error support into opt module.

### DIFF
--- a/src/base/iana/exterr.rs
+++ b/src/base/iana/exterr.rs
@@ -1,0 +1,176 @@
+//! Extended DNS Error
+
+//------------ Extended Error Code ---------------------------------------------------------
+
+int_enum! {
+    /// Extended DNS error codes.
+    ///
+    /// A complementary data can be put in EDNS opt, providing
+    /// additional information about the cause of DNS errors. Defined
+    /// in [RFC 8914]. Current registered values can be found in [IANA
+    /// registry].
+    ///
+    /// [RFC 8914]: https://tools.ietf.org/html/rfc8914
+    /// [IANA registry]: https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#extended-dns-error-codes
+    =>
+    ExtendedErrorCode, u16;
+
+    /// The error in question falls into a category that does not
+    /// match known extended error codes. Implementations SHOULD
+    /// include an EXTRA-TEXT value to augment this error code with
+    /// additional information.
+    (Other => 0, b"Other Error")
+
+    /// The resolver attempted to perform DNSSEC validation, but a DNSKEY
+    /// RRset contained only unsupported DNSSEC algorithms.
+    (UnsupportedDnskeyAlgorithm => 1, b"Unsupported DNSKEY Algorithm")
+
+    /// The resolver attempted to perform DNSSEC validation, but a DS
+    /// RRset contained only unsupported Digest Types.
+    (UnsupportedDsDigestType => 2, b"Unsupported DS Digest Type")
+
+    /// The resolver was unable to resolve the answer within its time
+    /// limits and decided to answer with previously cached data
+    /// instead of answering with an error. This is typically caused
+    /// by problems communicating with an authoritative server,
+    /// possibly as result of a denial of service (DoS) attack against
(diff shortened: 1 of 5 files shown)


partim:
Thank you for the PR! It is looking good.

Perhaps it would be a good idea to move the `Code` enum into the `base::iana` module where all the IANA managed definitions already live, perhaps as `base::iana::exterr::ExtendedError`? That way we could reuse the macros already defined for defining these enums with all the methods and impls. The purpose string from the registry can go where the mnemonics normally live.

xofyarg:
I've separated the definitions and logic, please take another look.

Although the EXTRA-TEXT is not used very often, the `fmt::Display` should include it. So I didn't use any macro on `iana::ExtendedErrorCode`, but implement it on `opt::ExtendedError`.

partim:
Looking great now, I’d be happy to merge it. Thanks again for the PR!

We could probably manually impl `Display` for `ExtendedErrorCode` as well – it just needs to call `to_mnemonic`. Do you want to add that or are you fine the way it is and want me to merge? Either way is fine with me.

xofyarg:
Sure, I can add something like below:
```
int_enum_str_with_prefix!(ExtendedErrorCode, "EDE", b"EDE", u16, "unknown ede code");  
```

BTW, I wonder why the mnemonic is stored as `&'static [u8]` instead of `&'static str`. It makes some functions trying to convert it to utf8 unnecessarily like in `Display`. I had some benchmarks:

```
use super::super::octets::Octets512;
pub struct B1(pub ExtendedError<Octets512>);
impl fmt::Display for B1 {
        match self.0.code.to_mnemonic() {
            Some(m) => {
                for ch in m {
                    f.write_char(*ch as char)?
                }
            }
            None => write!(f, "Undefined({})", self.0.code.to_int())?,
        }

        if let Some(text) = &self.0.text {
            let text = str::from_utf8(text.as_ref()).map_err(|_| fmt::Error)?;
            write!(f, " ({})", text)?;
        }
        Ok(())
    }
}

pub struct B2(pub ExtendedError<Octets512>);
impl fmt::Display for B2 {
        self.0.code.fmt(f)?;

        if let Some(text) = &self.0.text {
            let text = str::from_utf8(text.as_ref()).map_err(|_| fmt::Error)?;
            write!(f, " ({})", text)?;
        }
        Ok(())
    }
}

pub struct B3(pub ExtendedError<Octets512>);
impl fmt::Display for B3 {
        match self.0.code.to_mnemonic() {
            Some(m) => f.write_str(unsafe { str::from_utf8_unchecked(m) })?,
            None => write!(f, "Undefined({})", self.0.code.to_int())?,
        }

        if let Some(text) = &self.0.text {
            let text = str::from_utf8(text.as_ref()).map_err(|_| fmt::Error)?;
            write!(f, " ({})", text)?;
        }
        Ok(())
    }
}
```

```
ede_display             time:   [166.57 ns 167.09 ns 167.66 ns]
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) high mild
  3 (3.00%) high severe

ede_display_via_code    time:   [206.98 ns 207.29 ns 207.62 ns]
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) high mild
  3 (3.00%) high severe

ede_display_fast        time:   [48.441 ns 48.892 ns 49.395 ns]
Found 31 outliers among 100 measurements (31.00%)
  15 (15.00%) low severe
  2 (2.00%) low mild
  4 (4.00%) high mild
  10 (10.00%) high severe
```

partim:
`int_enum_str_with_prefix!` also adds an impl for `FromStr` which, I think, we don’t really want here. Maybe just impl `Display` directly for now?

I think the decision to make mnemonics `[u8]`s and not `str`s was because they are always ASCII. I wouldn’t mind switching them to `str`.

This is a breaking change, though, so would be wait at least until 0.6. I have plans to revisit zone file reading and writing as that is fairly slow today, too. That would be a good occasion.

xofyarg:
OK, using directly implementation for `Display` now.

partim:
Thank you once again for the PR!

PR #80, opened 2020-11-05
Fixed a TXT wireformat parser bug and allow calling rtype() on enum record data types

vavrusa:
There is an error currently which checks remaining bytes of the parent parser, which has already been consumed, so the TXT RDATA elements are not actually checked. This lets malformed data past the parser and crashes later.

### DIFF
--- a/src/rdata/rfc1035.rs
+++ b/src/rdata/rfc1035.rs
@@ -1199,7 +1199,7 @@ impl<Octets: AsRef<[u8]>> Txt<Octets> {
     }
 
     pub fn is_empty(&self) -> bool {
-        false
+        self.0.as_ref().is_empty()
     }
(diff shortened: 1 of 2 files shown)


src/rdata/rfc1035.rs
```diff
@@ -1301,7 +1301,7 @@ impl<Ref: OctetsRef> Parse<Ref> for Txt<Ref::Range> {
         let len = parser.remaining();
         let text = parser.parse_octets(len)?;
         let mut tmp = Parser::from_ref(text.as_ref());
-        while parser.remaining() != 0 {
+        while tmp.remaining() != 0 {
```
vavrusa, line 1304: This is the actual bug.

PR #84, opened 2021-01-13
chore: upgrade deps following tokio 1.0

koivunej:
This was surprisingly easy, nothing special to call out.

I'll post a follow-up to this with all of the rustfmt/whitespace changes.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,7 +10,7 @@ Breaking
   `resolv-sync`. ([#74])
 * The trait `domain::base::octets::IntoOctets` has been merged into
   `domain::base::octets::OctetsBuilder`. It’s method `into_octets` is now
-  available as `freeze` on `OctetsBuilder. ([#75])
+  available as `freeze` on `OctetsBuilder`. ([#75])
 
 * Supprt for extended errors defined in [RFC 8914]. ([#79] by [@xofyarg])
 * New traits `domain::base::octets::OctetsFrom` and `OctetsInto` to
--- a/Changelog.md
+++ b/Changelog.md
@@ -17,6 +17,8 @@ Breaking
   convert types that are generic over octets sequences between different
   octets sequences. ([#77])
 
+* Upgrade to tokio 1.0, bytes 1.0, and latest of other dependencies ([#84] by [@koivunej])
+
 Bug Fixes
 
 * Fix domain name compressors when giving a root label only. ([#76]
@@ -38,8 +40,10 @@ Other Changes
 [#77]: https://github.com/NLnetLabs/domain/pull/77
 [#79]: https://github.com/NLnetLabs/domain/pull/79
 [#80]: https://github.com/NLnetLabs/domain/pull/80
+[#84]: https://github.com/NLnetLabs/domain/pull/84
 [@vavrusa]: https://github.com/vavrusa
 [@xofyarg]: https://github.com/xofyarg
+[@koivunej]: https://github.com/koivunej
 [RFC 8914]: https://tools.ietf.org/html/rfc8914
 
 
--- a/src/resolv/stub/conf.rs
+++ b/src/resolv/stub/conf.rs
@@ -392,14 +392,14 @@ impl ResolvConf {
                                   line.starts_with('#') {
                 continue
             }
-            
+
             let mut words = line.split_whitespace();
             let keyword = words.next();
             match keyword {
                 Some("nameserver") => self.parse_nameserver(words)?,
                 Some("domain") => self.parse_domain(words)?,
                 Some("search") => self.parse_search(words)?,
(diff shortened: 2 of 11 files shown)


koivunej:
Strange, once again this happened that local clippy reported nothing but the ci fails. Lets see which of these are issues for 1.45 and latest stable...

koivunej:
Next up: beta clippy chores.

koivunej:
So the `beta` and `nightly` clippy lints now trigger an error on stable because there is a similar lint. This can most likely be worked around, but also the streamlining of the CI around the MSRV (1.45.0) starts to sound tempting.

koivunej:
Ok, trying another way around, I dont think there's a way to adjust the 1.45.0 clippys handling of future lint names.

partim:
Changing `master::scan::Scanner::ok` to just `fn ok(&mut self)` and then changing the calls into `{ self.ok(); Ok(()) }` might be easier at this point. (If I didn’t miss anything, all the calls are for `Scanner::ok::<()>`).

koivunej:
@parlin agreed as I am currently looking at what happens with `Ok(self.ok(()))` which triggers another "passing unit value to function" lint from clippy or rustc.

partim:
This function certainly is, er, creative. Apologies for turning a seemingly straightforward task into a wild goose chase!

koivunej:
Lets see if it works now... I'll squash that back and forth once we have a next issue.

koivunej:
Perhaps with the NOOP TODO function, it would be better to have `todo!()`.

koivunej:
Yeah, I think this is now ready for review. I am thinking all of the refactorings are quite clear, at least from the single-commit view. Nothing really perilous except for the `!(0x20...0x7f).contains(&ch)` which I was sure to copy paste (and in the other case change `b' '` into 0x20 for "uniform").

src/resolv/stub/conf.rs
```diff
@@ -452,10 +452,16 @@ impl ResolvConf {
         words: SplitWhitespace
     ) -> Result<(), Error> {
         // XXX TODO
-        let _ = words; 
-        Ok(())
+        let _ = words;
+        let ret = Ok(());
+
+        if ret.is_err() {
```
partim, line 458: Maybe it is better to comment out `parse_sortlist` entirely and replace its call in the match with a noop.

Changelog.md
```diff
@@ -10,13 +10,15 @@ Breaking
...
   `domain::base::octets::OctetsBuilder`. It’s method `into_octets` is now
-  available as `freeze` on `OctetsBuilder. ([#75])
+  available as `freeze` on `OctetsBuilder`. ([#75])
 
 * Supprt for extended errors defined in [RFC 8914]. ([#79] by [@xofyarg])
 * New traits `domain::base::octets::OctetsFrom` and `OctetsInto` to
   convert types that are generic over octets sequences between different
   octets sequences. ([#77])
 
+* Upgrade to tokio 1.0, bytes 1.0, and latest of other dependencies ([#84])
```
partim, line 20: You may or may not want to name-check yourself here. So far, I’ve always did that for ‘outside contributions.’

src/resolv/stub/conf.rs
koivunej: My preference would be to replace the match or the fn body with `todo!()` but I'll comment it out.

partim:
I forgot to say this over all the Clippy kerfuffle: Thank you for the PR!

Changelog.md
koivunej: This sounds like a good way. I'll add myself here.

koivunej:
> Thank you for the PR!

You're welcome. We are using `domain` in [`rust-ipfs`](https://github.com/rs-ipfs/rust-ipfs) in dnslink resolution; I didn't yet check if there are any other changes needed for us in the breaking but surely they can be handled.

partim:
Cool! If there is anything you need or want or wish you had, let us know!

koivunej:
@partim: I may have done a mistake by raising the smallvec version requirement needlessly, and while working on this, it seems that futures = "0.3.9" has been yanked replaced by "0.3.10". I'll see what would be a good way forward with these.

partim:
Futures should be fine. Cargo will just resolve to 0.3.10, anyway.

I don’t think raising smallvec is an issue either. As far as semver goes, anyone requiring 1.2 should be fine with 1.6, too.

koivunej:
> anyone requiring 1.2 should be fine with 1.6, too

Now that I double checked I may have misunderstood the conflict given to me by cargo while upgrading libp2p. Still, the #86 fixes the issue, while adding the additional minimal-version checks which should make the build more future proof, and need to think about this less in the future.

PR #85, opened 2021-01-13
cargo fmt

koivunej:
This PR adds a ci step to check that everything is formatted, followed up by formatting of all files. 

Rationale for using rustfmt: This helps folks (including myself) who have rustmt by default run on every save to avoid some work of undoing the formatting. It'll take care of other whitespace errors (as reported by git) I believe as well.

Rationale for using the `stable` as formatter: I'd assume most devs looking to make changes in future will be using the latest stable locally. While there have been minor regressions in the past for rustfmt releases, I've found it to be quite stable, most likely just reformatting previously bad looking sections.

I do not really recommend merging this now, since it might make it quite difficult for anyone with open changes to merge/rebase them in. Feel free to just take the ci check, modify it to be based on other version and/or close this at any time :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,8 @@ jobs:
       run: rustup component add clippy
     - if: matrix.rust != 'nightly'
       run: cargo clippy --all-features -- -D warnings
+    - if: matrix.rust == 'stable' && matrix.os == 'ubuntu-latest'
+      run: cargo fmt --all -- --check
     - run: cargo build --no-default-features --verbose
     - run: cargo test --features="master resolv resolv-sync sign tsig validate" --verbose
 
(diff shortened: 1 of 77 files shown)


koivunej:
Whoops, botched the workflow yaml.

koivunej:
Quite interesting; did `rustfmt` allow a clippy lint to start triggering? https://github.com/NLnetLabs/domain/runs/1695335298?check_suite_focus=true#step:5:96

partim:
I am willing to add this, but I have one hard requirement: maximum line length of 78 characters. I guess that can be chosen via rustfmt configuration?

koivunej:
To be honest I have no idea, but I would assume so. I have never looked into configuring rustfmt, my preference is just to have a tool to handle the formatting consistently.

koivunej:
Found it, and applied: https://rust-lang.github.io/rustfmt/?version=master&search=#max_width

partim:
Thanks!

I might have to do some more tweaking of the config along the way, but that’s not a stopper for now.

PR #86, opened 2021-01-13
More lenient dependency requirements

koivunej:
This fixes my over-eager minimum version raising for smallvec. Since the earliest of 1.0.0 seems to work, it should be required, and the same logic is followed for other deps. I found this by trying to update the crate as the rust-ipfs dep and there was at least one conflict with a smallvec 1.4.2 requirement coming from libp2p-core. With "1" domain is compatible with that as well.

In addition to tweaking the versions, this adds a `cargo +nightly update -Z minimal-versions` following https://github.com/rust-lang/cargo/issues/5657. This is how I found that `chrono 0.4.0` was not good, but `0.4.6` was. Later on the dependency minimal versions should be raised if the last test starts to fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,4 +21,9 @@ jobs:
       run: cargo clippy --all-features -- -D warnings
     - run: cargo build --no-default-features --verbose
     - run: cargo test --features="master resolv resolv-sync sign tsig validate" --verbose
-
+    - if: matrix.rust == 'nightly'
+      run: |
+        cargo +nightly update -Z minimal-versions
+        cargo check --features="master resolv resolv-sync sign tsig validate" --verbose --all-targets
+        cargo test --features="master resolv resolv-sync sign tsig validate"
+      name: Check and test with minimal-versions
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ path = "src/lib.rs"
 rand           = "0.7"
 
 bytes          = { version = "1.0", optional = true }
-chrono         = { version = "0.4", optional = true }
+chrono         = { version = "0.4.1", optional = true }
 futures        = { version = "0.3", optional = true }
 #openssl       = { version = "0.10", optional = true }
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -21,4 +21,9 @@ jobs:
       run: cargo clippy --all-features -- -D warnings
     - run: cargo build --no-default-features --verbose
     - run: cargo test --features="master resolv resolv-sync sign tsig validate" --verbose
-
+    - if: matrix.rust == 'nightly'
+      run: |
+        cargo +nightly update -Z minimal-versions
+        cargo check --features="master resolv resolv-sync sign tsig validate" --verbose --all-targets
+        cargo test --features="master resolv resolv-sync sign tsig validate"
```
koivunej, line 28: Question related to these features: I think this is list is the same as `--all-features`, is there a reason to specify them separatedly?

partim: I think `interop` is missing. Which is important because that one wants NSD installed so it won’t work. If there’s a better way for selectively skipping tests than I’d be more than happy to change that.

koivunej: Aah ok, that explains it. I don't think there's a better way!

koivunej:
I am currently asking around for that mio complication error on mac. minimal-versions check passes on linux of course, which I have locally :)

koivunej:
Traced the mac error to https://github.com/tokio-rs/mio/issues/1439. I think macos cannot run the minimal-versions for a while unless they go ahead and yank the 0.7.6 and 0.7.7, unless there is another solution I cannot see. Trying this the check on windows, if just omitting mac from the list would work.

koivunej:
On windows there probably needs to be more recent chrono.

koivunej:
All right, looking good. Will do a squash on the back and forth, and this should be ready. Apologies for the deps issue caused in #84, I got too focused on the clippy lints :)

partim (approved):

PR #88, opened 2021-01-19
fix: create directories under working copy only

### LINKED ISSUE #87: feature = "interop" tests create directories outside of working copy
Just noticed the extra `target` dir when cleaning up after my PRs.

The files are (relative to `Cargo.toml` when executing `cargo test --all-features`):
- `../target/test/tsig_client_nsd/nsd.conf`
- `../target/test/tsig_client_sequence_nsd/nsd.conf`

I wonder if this is because I do not have the `nsd` binary installed?

koivunej:
Closes #87.

### DIFF
--- a/src/tsig/interop.rs
+++ b/src/tsig/interop.rs
@@ -71,7 +71,7 @@ fn tsig_client_nsd() {
     let mut nsd = Command::new("/usr/sbin/nsd")
         .args(&["-c", &format!("{}", nsdconfpath.display()), "-d"])
         .spawn()
-        .unwrap();
+        .expect("failed to start nsd");
     thread::sleep(Duration::from_secs(1));
     if nsd.try_wait().unwrap().is_some() {
         panic!("NSD didn't start.");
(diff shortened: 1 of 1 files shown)


src/tsig/interop.rs
```diff
@@ -36,7 +36,7 @@ fn tsig_client_nsd() {
     let rng = SystemRandom::new();
 
     let cur_dir = env::current_dir().unwrap();
```
koivunej, line 38: Just re-tested this, it would seem that this is set to the directory of `Cargo.toml` before executing the tests. However [rust book](https://doc.rust-lang.org/book/ch11-02-running-tests.html#running-tests-in-parallel-or-consecutively) doesn't consider the working directory to be relied on, so I'll add a PR to do this relative to CARGO_MANIFEST_DIR per [shepmasters SO answer](https://stackoverflow.com/a/30004252).

partim:
Thanks again for the fixes!

PR #89, opened 2021-01-29
change OctetsFrom blanket implementation

xofyarg:
Previous implementation uses "From" trait to tie both octets together,
makes a direct conversion. But the trait implementation maybe missing
for some common foreign types, like "impl From<Bytes> for Vec<u8>".

This change connects the two octets by indirect traits, it should work
for most common types.


xofyarg:
Hi @partim,

Apologize for the cold PR, I feel it's easier to discuss with it. This is a followup of #77.

As I adopting the new trait `OctetsFrom`, I found it less convenient for types which don't have `From` implemented in between, and I cannot do much about it because they are foreign types. I also added some use cases as tests, please let me know what you think, thank you.

partim:
The reason why we don’t do the blanket implementation is that it blocks some cheap conversions. For instance, you may be able to convert a `Cow<[u8]>` into a `Vec<u8>` without any allocations at all. But with the blanket impl you always would have to allocate. Not having the blanket is indeed cumbersome but I think the potential performance gains are worth it.

PR #101, opened 2021-04-19
Add extra bit truncation to ECS

xofyarg:
More bits need to be considered if the prefix len does not stop at
byte boundary. This change adds an extra bits mask to deal with it.


partim:
Thank you for the PR and sorry for the delay – spring vacation time sort of situation.

The PR looks good! Let me know if you would like to have a point release for it.

xofyarg:
Thanks for merging. No need to release at this point, I'm still catching up the 0.6 release.

PR #102, opened 2021-04-30
rfc7871: fix ECS addr truncation and add strict check

xofyarg:
The code panics when the prefix length is 0(minus overflow) or at
byte boundary(shift overflow), fixed it and added more tests.

Added a strict check in parser, according to RFC, it should return
form error when there are non-zero bits beyond the prefix length.

### DIFF
--- a/src/base/opt/rfc7871.rs
+++ b/src/base/opt/rfc7871.rs
@@ -4,12 +4,13 @@ use super::super::iana::OptionCode;
 use super::super::message_builder::OptBuilder;
 use super::super::net::IpAddr;
 use super::super::octets::{
-    Compose, FormError, OctetsBuilder, Parse, ParseError, Parser, ShortBuf
+    Compose, FormError, OctetsBuilder, Parse, ParseError, Parser, ShortBuf,
 };
 use super::CodeOptData;
 
-
 //------------ ClientSubnet --------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


xofyarg:
@partim sorry about the bad commit #101. Do you think the opt constructor should enforce the address with correct bits, even before composing it into bytes?

The commit also includes some changes due to rustfmt, I can remove it if you don't prefer that.

partim:
> Do you think the opt constructor should enforce the address with correct bits, even before composing it into bytes?

I think quietly changing the bits to zero is more user friendly. But doing that in the constructor rather than only when composing might be a better approach as `ClientSubnet::addr` will then always return the ‘proper’ address. So, enforce that the `addr` field always conforms with the RFC as an invariant of the type.

> The commit also includes some changes due to rustfmt, I can remove it if you don't prefer that.

Nah, that’s fine.

xofyarg:
I added some checks inside the constructor.

CI seems has an issue with beta toolchain. stable works fine on my side though.

partim:
Once again, sorry for the delay. Between all the holidays in May and other projects, this PR sadly slipped.

I have fixed Clippy’s warnings and disabled it for Beta in main, so if you could merge main into your branch, we should go green on the actions and then can merge the PR.

xofyarg:
No worries, and glad to hear you got long vacations!

The PR has been rebased.

PR #106, opened 2021-07-23
Introduce Scanner trait

### LINKED ISSUE #105: More flexible API for record data deserialization
I would love to use this library for a small educational name server I am working on, however the API does not quite seem to cover my use case (yet? :). I want to provide a command-line interface to add records while the server is running, e.g:

```
myclient example.com add mx 10 alt1.example.com.
```

The only provided API to parse `MasterRecordData` from a human friendly string seems to be [`MasterRecordData::scan()`](https://docs.rs/domain/0.6.1/domain/rdata/enum.MasterRecordData.html), powered by the [`Scan`](https://docs.rs/domain/0.6.1/domain/master/scan/trait.Scan.html) trait of the experimental `master` module. While attempting to use this API, I stumbled upon several issues:

* unhelpful error messages
  * attempting to parse `x` as MX data yields `unexpected 'x'`, it would be nice to tell the user what was expected (numeric MX priority)
  * attempting to parse `99999` as MX data yields `illegal integer`, it would be nice to tell the user why the integer is illegal (priority may not be greater than 65535)
  * `unespected end of file` doesn't tell you what was expected
* The `Scan` trait isn't suited to parse command-line arguments since Scanner assumes the context of a Zone file.
   * attempting to parse `10 foo` as MX data yields `owner @ without preceding $ORIGIN` (which I'd argue is confusing even in a zone file but is of course completely out of place for command-line arguments)
* I would like users to be able to directly provide binary strings without needing to escape them, that's not possible with Scanner.
   * For example `TXT` records can contain arbitrary binary data, I want users to be able to directly provide an arbitrary TXT value as a command-line argument (which don't need to be UTF-8).

So I think it would be great if this library could somehow support both parsing records from zone files as well as parsing them from e.g. an `Iterator<Item=OsString>` (where each item represents one part of the record data).

I guess this could be achieved by changing the `Scan` trait, that record data types implement, to take a new `Scanner` trait (instead of a zone-specific scanner). I am not quite sure yet how this `Scanner` trait would look like exactly but I think it would need to abstract over the zone-file specifics like `$ORIGIN`, by e.g. having a `scan_absolute_name(&mut self) -> Result<Dname, ScanError>` method.

Do you think this is worth pursuing?

not-my-profile:
PR to resolve #105.

Todos:

* [ ] rewrite parser with nom
* [ ] add default implementation for `scan_domain_name()`
* [ ] consider always allowing double quoting, e.g. the NSD zone file parser allows everything to be double-quoted
      (replacing all `_word` uses by `_phrase` uses would simplify the trait)
* [ ] update documentation (fix links, add module & trait description)


partim:
Thank you for the PR! It’s looking quite good.

One thing I would want to accomplish with this redesign is to make it possible to use other octets types than `Bytes`. I’m thinking that the `Scanner` trait should gain an associated type `Octets` that represents the native octets type of the scanner. `Scan` would then be generic over the source octets type.

I would also like to get away from the source basically being an iterator over characters, as this slows down parsing for the usual case of ASCII files. Essentially, I want to be able to memory map large files and then works with `&[u8]` on top of that as much as possible, avoiding allocations all together.

I’m thinking that maybe we should introduce a new type that represents a word as a kind of cow: either a simple octets sequence of the source’s native octets type or a more complex variant that allows iterating over a sequence of octets sequences and escape sequences.

Overall, those two changes would make `scan_word` much simpler as we only need one closure. It would receive one of those fancy new word values and returns the final result or an error.

So, yeah, one “tiny” change that is really quite changing everything …

partim:
Sorry, wrong button.

not-my-profile:
Does the following match what you have in mind?

```rust
enum Chunk<'a, B> {
    Slice(&'a B),
    Escaped(u8),
}

struct RdataError {}

trait Scanner<B: AsRef<[u8]>> {
    type Err;

    fn scan_word<F, U>(&mut self, func: F) -> Result<U, Self::Err>
    where
        F: FnOnce(Vec<Chunk<B>>) -> Result<U, RdataError>;
}
```

If that's the case, I think we should probably also introduce a newtype e.g. `struct Chunks<B>(Vec<Chunk<B>>)`, so that we can add convenience functions. I think this would also enable us to get rid of `scan_string_word()`: if we have  `Chunks::to_str(self) -> Result<Cow<str>, Utf8Error>`, by making the closure return `Result<U, Into<RdataError>>` it should be possible to do e.g. `scanner.scan_word(Chunks::to_str)?`. What do you think?

partim:
I am now wondering whether I am overthinking this. At the end of the day, you don’t really care about escape sequences. I _think_ the only time you need to know whether a certain octet was escaped or not is for the dot in domain names.

So, perhaps this is as easy as having two associated types with the `Scanner` trait for the two kinds of sequences you really want from the scanner. So, something like:

```rust
pub trait Scanner {
    type Word: AsRef<[u8]>;
    type Dname: ToDname;
    type Err: Error;

    fn scan_word(&mut self) -> Result<Self::Word, Self::Err>;
    fn scan_dname(&mut self) -> Result<Self::Dname, Self::Err>;

    // ... plus a few more convenience methods.
}
```

The `Error` trait can be similar to serde’s, except that I kind of want to get away without allocations, so maybe require a `&'static str` for `Error::custom`’s message argument.

not-my-profile:
I think with `type Word: AsRef<[u8]>` we would need to copy the whole word every time a word contains an escape sequence. Do we accept allocations for such words since we expect escape sequences to be rare anyway?

I have also thought about getting rid of the closures, since the `Scan` impls don't seem to require backtracking ... although I am not completely sure.

partim:
Resolving escape sequences via allocations should indeed be fine. I think all escape sequences are longer than their content, so the scanner could even cheat and quietly change the buffer without needing to allocate.

I think there was a case somewhere where the parser needed to backtrack (it’s been a while, but I surely wouldn’t have come up with this if it wasn’t necessary), so maybe the closures need to stay.

not-my-profile:
Yes the zone file parser currently needs backtracking e.g. `domain::master::entry::Entry::scan()`. I meant that maybe it's only needed by the zone-file-specific code and not needed in the `Scanner` trait.

partim:
It is entirely possible that backtracking is necessary for one of the record types. But in any case, I don’t think having separate facilities for record data parsing and zone file parsing (of which record data parsing is a part) makes sense. So this new scanner should support parsing complete records and entire zone files.

An option to support backtracking would be to enable it explicitly through a separate method that takes a closure, i.e.,

```rust
fn try_scan<T, E, F: FnOnce(&mut Self) -> Result<Option<T>, E>>(&mut self, op: F) -> Result<Option<T>, E>;
```

The problem with peeking is that it limits you to look at the next word only whereas the closure can both try and grab multiple words and can also be nested for even more complex cases.

not-my-profile:
If we are to support both streaming parsing and complete parsing with memory map, wouldn't we need to write the zone file parsing logic twice? Or is there a way to abstract over this?

partim:
I honestly haven’t thought that far. But the idea of the trait (at least in my head) is that you can have multiple implementations that are tailored for different needs. I suspect part of the logic can be reused, but I don’t mind having to have some duplication if there are different use cases.

not-my-profile:
>I don’t think having separate facilities for record data parsing and zone file parsing (of which record data parsing is a part) makes sense.

I think it does make sense. The entire reason I am working on this is that I want to deserialize record data without having to bother with a zone file parser, which I don't need. I think I made that very clear. Do you no longer consider my use case valid or am I misunderstanding something?

Note that my PR already separates the zone-file-specific errors from the Rdata-specific errors.

partim:
But your command line format uses zonefile format. The only difference is that the “tokenization“ is done through the `env::arg` mechanism rather than an internal lexer. Everything above that will be the same. The `Scanner`/`Scan` pair of traits represents this above layer. Why have two different mechanism that do essentially the same?

As a side note: I have been thinking about providing serde serialization all these types. Primary use case would be to gain a JSON exchange format for APIs, but serde can deserialize from a sequence of elements as well. Perhaps this is a better fit for what you have in mind?

not-my-profile:
>The only difference is [...] the “tokenization“ ... Why have two different mechanism that do essentially the same?

The zone file parser currently does not just operate on "word tokens" but also uses Scanner functions like `scan_space()` and `scan_newline()` which don't make any sense for command-line arguments.

>Perhaps [serde] is a better fit for what you have in mind?

I don't think so, since I want to deserialize record data from a list of `u8` slices.

partim:
I think `Entry` uses the space thing to determine whether a line is indented as that signifies using the owner name from the last entry. This could become a simple check function that a scanner could always respond to with ‘no’ if it doesn’t care. `scan_terminator` sounds good. You probably want that for your use case as well to be able to detect trailing garbage.

not-my-profile:
My latest commit decouples `Entry::scan` from `CharSource` by introducing the following trait:

```rust
pub trait ZoneScanner: Scanner {
    fn eof_reached(&mut self) -> bool;
    fn scan_terminator(&mut self) -> Result<(), Self::Err>;
    fn scan_blank(&mut self) -> Result<(), Self::Err>;
    fn skip_after_terminator(&mut self) -> Result<(), Self::Err>;
    fn use_last_owner(&mut self) -> bool;
    fn origin(&self) -> &Option<Dname<Bytes>>;
}
```

I consider extending the trait to result in a more descriptive API rather than having a buch of no-op methods for when you only need record data deserialization.

The pipeline fails because apparently Rust < 1.49.0 isn't as smart about inferring generic constraints. Would it be acceptable to bump the minimum supported version to 1.49.0 (2020-12-31)?

partim:
If we do this break and declare `Scanner` to only supply a sequence of tokens, then I don’t think we need to introduce the `ZoneScanner` trait just yet and rather just leave that as a struct for now.

However, in this case I still would want the error reporting to change along the lines of what serde does. I.e., instead of having an enum, this should be a trait along the lines of this:

```rust
pub trait Error {
    fn custom(err: &'static str) -> Self;
}
```

The reasoning for not using `impl fmt::Display` as the argument for `custom` is that formatting requires allocation and we don’t necessarily have that. In most cases, I feel a static error message is good enough anyway.

partim:
The minimum Rust version has been increased to 1.49 in the main branch.

not-my-profile:
>If we do this break and declare Scanner to only supply a sequence of tokens, then I don’t think we need to introduce the ZoneScanner trait just yet and rather just leave that as a struct for now.

I don't quite follow. I think ideally we would have two Scanner implementations one for streaming parsing and one for complete parsing. With the ZoneScanner trait both implementations could implement it and be used with `Entry` (without having to duplicate the Entry logic).

not-my-profile:
I have started rewriting the zone file parser with nom ... it's looking good so far. Currently the `Scan` implementations are for `bytes::Bytes` only, which we obviously want to change so that you can avoid allocations. I have been thinking about how we can support both borrowed and owned deserialization of record data. I am not sure how to implement this, since with generics there are many ways. One idea I had is introducing a `BytesDealer` trait as a `Scan` generic parameter:

```rust
pub trait BytesDealer {
    type Bytes: AsRef<[u8]>;
    type Builder: OctetsBuilder<Octets=Self::Bytes>;
    
    fn from_cow(data: Cow<[u8]>) -> Self::Bytes;

    fn builder() -> Self::Builder;
}


pub trait Scanner {
    fn scan_bytes(&mut self) -> Result<Cow<[u8]>, RdataError>;
    // etc.
}

pub trait Scan<B: BytesDealer>: Sized {
    fn scan<S: Scanner>(scanner: &mut S) -> Result<Self, RdataError>;
}

impl<B: BytesDealer> Scan<B> for Txt<B::Bytes> { ... }
```

Does that make sense? Do you have a better idea?

partim:
My idea was to have the octets type be an associated type of the `Scanner` trait and have `Scan` be generic over the `Scanner`. Isn’t that good enough?

not-my-profile:
Some scanner impls (e.g. UnknownRecordData) need to create a new mutable buffer of the octets type.

partim:
The ideas from this PR have been integrated into #142 which has now been merged.

PR #109, opened 2021-08-23
base/flags: add utility type Flags

### LINKED ISSUE #108: utility struct for flags
I'd find it useful if there was a single struct containing the flags from the message header. It'd also be useful if it had a text representation, e.g. "QR AA" and if the struct could be constructed from this string representation as well.

I'd use the same string representation as the dnspython library: https://github.com/rthalley/dnspython/blob/master/dns/flags.py

For now, I'd stick with standard flags only (ignoring EDNS flags), which can be easily constructed from header.

Would this feature make sense to you? If so, I could contribute the code.

tomaskrizek:
The Flags type simplifies working with all flags, e.g. making it easier
to compare, hash or represent them as string.

The string notation is consistent with dnspython library
(https://github.com/rthalley/dnspython/blob/master/dns/flags.py).

Fixes #108

### DIFF
--- a/src/base/header.rs
+++ b/src/base/header.rs
@@ -20,7 +20,7 @@ use super::octets::{
     Compose, OctetsBuilder, Parse, ParseError, Parser, ShortBuf,
 };
 use core::convert::TryInto;
-use core::mem;
+use core::{fmt, mem, str::FromStr};
 
 //------------ Header --------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


partim:
Thanks for the PR. It is looking good!

However, I think I’d prefer to have the `Flags` type be part of the `headers` module instead of having a separate module.

tomaskrizek:
I had the `Flags` as part of `header` module at first, but when I had to create the extra `FromStrError` type, it seemed better to put it in a separate namespace. Wouldn't it be confusing to have this type in the `header` module as well?

partim:
I think `FlagsFromStrError` is perfectly fine. It’s not like people will have to type the name all that often.

tomaskrizek:
Ok, thanks for the suggestion. I've renamed the error type and moved everything into the `header` module.

PR #110, opened 2021-09-15
validate: check for the correct public key size and not infer from RRSIG length

vavrusa:
The public key length should be checked instead of the signature hash length,
which in theory should match, but a shorter signature should be treated as bogus,
while a shorter key should be treated as unsupported.


partim:
Oops, sorry, forgot to merge.

PR #115, opened 2021-10-22
add support for SVCB record

xofyarg:
This commit add initial support for SVCB and HTTPS record defined in
[draft-ietf-dnsop-svcb-https][1]. It includes basic interaction with
the rdata, convert from/to the wire format.

[1]: https://datatracker.ietf.org/doc/draft-ietf-dnsop-svcb-https/

### DIFF
--- a/src/base/iana/mod.rs
+++ b/src/base/iana/mod.rs
@@ -34,6 +34,7 @@ pub use self::opt::OptionCode;
 pub use self::rcode::{OptRcode, Rcode, TsigRcode};
 pub use self::rtype::Rtype;
 pub use self::secalg::SecAlg;
+pub use self::svcb::SvcbParamKey;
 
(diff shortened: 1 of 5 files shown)


partim:
Thanks for the PR and sorry for the long silence! The PR looks great!

partim:
I was a bit quick. Is there a reason you added them as pseudo records rather than zone records?

xofyarg:
Thank you for taking a look. This module will be improved as we involve more into SVCB. Support for zone record will definitely be added later. I'm waiting for the draft to settle down.

partim:
We’re also in the process of fundamentally re-thinking zone file parsing. So it may be good to wait until after that is done, anyway, given that the presentation format isn’t entirely trivial. We’ll definitely keep SVCB in mind for the new architecture.

PR #117, opened 2021-11-22
Enable no_std support.

Jezza:
`domain`, while offering a std feature, doesn't compile without std, regardless of the feature.

I've marked this PR as a WIP, as there's still some open questions.

We do seemingly need rng in a couple of places.
There is some support for non-os based rng in `rand`, but we'll probably need to use some static variable to hold the value.
There's also the question of seeding the thing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,8 +16,7 @@ name = "domain"
 path = "src/lib.rs"
 
 [dependencies]
-rand           = "0.8"
-
+rand           = { version = "0.8", optional = true }
 bytes          = { version = "1.0", optional = true }
 chrono         = { version = "0.4.6", optional = true }
 futures        = { version = "0.3", optional = true }
@@ -32,13 +31,14 @@ tokio          = { version = "1.0", optional = true, features = ["io-util", "mac
 libc = { version = "0.2.71", default-features = false, optional = true }
 
 [features]
-default     = ["std"]
+default     = ["std", "random"]
 interop     = ["bytes", "ring"]
 master      = ["std", "bytes", "chrono"]
-resolv      = ["bytes", "futures", "smallvec", "std", "tokio", "libc"]
+resolv      = ["bytes", "futures", "smallvec", "std", "tokio", "libc", "random"]
 resolv-sync = ["resolv", "tokio/rt"]
 sign        = ["std"]
 std         = []
+random      = ["rand"]
 tsig        = ["bytes", "ring", "smallvec"]
 validate    = ["std", "ring"]
 
--- a/src/base/header.rs
+++ b/src/base/header.rs
@@ -126,6 +126,7 @@ impl Header {
     }
 
     /// Sets the value of the ID field to a randomly chosen number.
+    #[cfg(feature = "random")]
     pub fn set_random_id(&mut self) {
         self.set_id(::rand::random())
     }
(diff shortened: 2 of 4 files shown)


src/base/header.rs
```diff
@@ -407,15 +407,15 @@ impl FromStr for Flags {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let mut flags = Flags::new();
-        for token in s.to_uppercase().split(' ') {
+        for token in s.split(' ') {
```
Jezza, line 410: Not exactly the prettiest thing, but to_uppercase() does allocate, and we only need to compare some basic ASCII characters.
We could probably be smarter about it, and just uppercase the ascii range, which is doable, but this seems to just be simple enough to not have to worry about it.

Jezza:
There's two remaining spots that are currently blocking me.
I do see the try operator in use, and I'm pretty sure that's also not going to work.
So, we might have to build a `try` macro.

`src/base/header.rs:130:29`
image

`src/base/opt/rfc7830.rs:85:13`
image

Jezza:
Almost done.

There's one last challenge.
How to pick a seed for RNG...

partim:
Thanks for the PR!

On the random generator issue: Given that `Header::set_random_id` is only used by the resolver which requires the std feature, anyway, and a convenience method in `MessageBuilder`, I think it is okay to simply only include the method if std is set and make rand an optional dependency.

Jezza:
Awesome, that just leaves `rfc7830.rs`, which uses it to generate random bytes for padding.

partim:
The RFC actually says that zeros SHOULD be used. So I think making `PaddingMode` and `Padding::mode` std-only, rename the current `Padding::new` and `Padding::push` to `{new,push}_with_mode`, and add new `new` and `push` methods without mode is probably the best way forward.

Jezza:
> Thanks for the PR!
> 
> On the random generator issue: Given that `Header::set_random_id` is only used by the resolver which requires the std feature, anyway, and a convenience method in `MessageBuilder`, I think it is okay to simply only include the method if std is set and make rand an optional dependency.

Should the whole function `request_axfr` be dependent on std, or should I just remove the "sets random id" part of the function.

partim:
I think making the whole function depending on std is fine, at least for now. We can revisit it later. Perhaps stick on comment on it, stating that this requires std because of the random ID.

Jezza:
Alrighty, changes are done.

I've gone with marking `PaddingMode::Random` as `feature = "std"`, as no point making whole types vanish.  
The `new` and `push` just use `PaddingMode::Zero` by default.  
I've added the `{new,push}_with_mode`, so that should cover any particular use cases remaining.

Cargo.toml
```diff
@@ -39,7 +38,7 @@ master      = ["std", "bytes", "chrono"]
 resolv      = ["bytes", "futures", "smallvec", "std", "tokio", "libc"]
 resolv-sync = ["resolv", "tokio/rt"]
 sign        = ["std"]
-std         = []
+std         = ["rand", "rand/std", "rand/std_rng"]
```
partim, line 41: Do we still want the additional _rand_ features?

partim:
I fixed the Clippy errors in main. Could you merge from main to hopefully make the CI go green?

bugadani:
Without knowing any of the details, would it be possible to separate the rand and std features? rand does not (yet?) support the espidf OS, which means domain can only be used for ESP32 targets without `std` support, which makes working with domain a bit less comfortable.

Jezza:
Hm, yeah, actually, would be easy enough.

Jezza:
Now you can do something like: `default-features = false, features = ["std"]`.

partim:
Note that the resolver needs random. I suspect that’s what you’d want to use on the ESP32? So perhaps we do need a work-around after all.

Jezza:
I don't think so.

I'm using it for an esp32 as well, and `domain` is only used in the `experimental` feature.
After I made these changes, it builds fine, so it's not using any `random` feature.

https://github.com/esp-rs/embedded-svc/blob/master/src/utils/captive.rs#L6-L13

bugadani:
Well it works perfectly for me, too, so I appreciate the change :)

Jezza:
> Note that the resolver needs random. I suspect that’s what you’d want to use on the ESP32? So perhaps we do need a work-around after all.

It does look like `embedded-svc` does offer a DNS server impl, but that's behind its own `std` feature flag.

partim:
Ok. In this case, can you add `random` to the features required by `resolv`? That would be the last thing before the PR can be merged.

Jezza:
Oh, yeah, of course.

partim:
Once again, thank you for the PR! There is likely going to be release 0.7.0 later this week.

PR #122, opened 2021-11-24
Implement `Octets*` for `heapless::Vec`

bugadani:
This PR implements support of `heapless::Vec` as octet buffer, and also fixes what I considered to be typos with regards to `smallvec`.

### DIFF
--- a/src/base/octets.rs
+++ b/src/base/octets.rs
@@ -1237,7 +1237,7 @@ mod serde {
 
         fn visit_bytes<E: serde::de::Error>(
             self,
-            value: &'de [u8],
+            value: &[u8],
         ) -> Result<Self::Value, E> {
             use core::iter::FromIterator;
 
--- a/src/base/octets.rs
+++ b/src/base/octets.rs
@@ -1235,23 +1235,9 @@ mod serde {
             ))
         }
 
-        fn visit_borrowed_bytes<E: serde::de::Error>(
+        fn visit_bytes<E: serde::de::Error>(
             self,
             value: &'de [u8],
-        ) -> Result<Self::Value, E> {
-            use core::iter::FromIterator;
-
-            if value.len() > N {
-                return Err(E::invalid_length(value.len(), &self));
-            }
-
-            Ok(heapless::Vec::from_iter(value.iter().copied()))
-        }
-
-        #[cfg(feature = "std")]
-        fn visit_byte_buf<E: serde::de::Error>(
-            self,
-            value: std::vec::Vec<u8>,
         ) -> Result<Self::Value, E> {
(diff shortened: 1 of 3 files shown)


src/base/octets.rs
```diff
@@ -1073,6 +1190,62 @@ mod serde {
...
+            use core::iter::FromIterator;
+
+            if value.len() > N {
+                return Err(E::invalid_length(value.len(), &self));
+            }
+
+            Ok(heapless::Vec::from_iter(value.iter().copied()))
+        }
+
+        #[cfg(feature = "std")]
```
bugadani, line 1235: Technically speaking, `heapless` doesn't need `std` to support serialization, but right now `serde/std` is always enabled. This whole visitor implementation might not be necessary, but may come in handy if this changes in the future.

I'm not sure if I could write `#[cfg(feature = "serde/std")]` as I believe that would be technically correct, if possible.

partim: I think we can switch to no-std Serde. Let me try that and make a PR.

partim: Turns out you can’t enable features on an optional dependency without also enabling the dependency itself. There is something called [weak dependency features](https://doc.rust-lang.org/cargo/reference/unstable.html#weak-dependency-features), but that’s not stable yet.

But, now that I look at it closer: I think you don’t want to implement `visit_byte_buf` or `visit_borrowed_bytes`, anyway. `heapless::Vec` has its own memory into which you have to copy the content, so you gain nothing from taking the reference or the vec. The default impls for both these methods forward to `visit_bytes` – which is what you should implement when you copy the data into your own memory. So, I think you should just implement that.

bugadani: I guess I should have checked the serde docs instead of pattern-matching from a different impl :) Thanks for the good insight.

src/base/octets.rs
```diff
@@ -374,17 +392,33 @@ where
     }
 }
 
-#[cfg(features = "smallvec")]
+#[cfg(feature = "smallvec")]
```
bugadani, line 395: Interesting. Did the smallvec support work as intended, before? Should I fix it here or revert this change and open a separate PR?

bugadani:
`-Z minimal-versions` selected `smallvec 1.0` which doesn't contain the `ToSmallVec` trait. I've updated the minimum version to the first where the trait was introduced. This is a bit out of scope of this PR, so just tell me if I should split it off to another.

partim:
I think it’s fine to fix here. We’re fixing octets implementations, so its sort of on topic.

partim:
All looking good now. Thank you for the PR!

PR #123, opened 2021-12-09
Fix typos in docstrings

DRiKE:
This is by no means a complete check of all the docstrings/comments, just a couple that caught my eye.

### DIFF
--- a/src/base/header.rs
+++ b/src/base/header.rs
@@ -775,19 +775,17 @@ impl HeaderSection {
     ///
     /// # Panics
     ///
-    /// This function panics if the the octets slice is shorter than 24
-    /// octets.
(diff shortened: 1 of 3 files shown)


src/base/header.rs
```diff
@@ -775,19 +775,17 @@ impl HeaderSection {
     ///
     /// # Panics
     ///
-    /// This function panics if the the octets slice is shorter than 24
-    /// octets.
+    /// This function panics if the octets slice is shorter than 12 octets.
```
bugadani, line 778: Just making a note that this line changes 24 to 12, which change github fails to highlight.

partim: That is correct. Not quite sure where the 24 came from, but the code indeed checks for the size of `HeaderSection` which is basically `[u8; 12]`.

PR #126, opened 2022-04-13
add OctetsFrom implementations for common octet types

vavrusa:
Hi! 👋 This implements OctetsFrom trait for most common octet type conversions instead of providing a blanket implementation as a followup for https://github.com/NLnetLabs/domain/pull/89

### DIFF
--- a/src/base/name/dname.rs
+++ b/src/base/name/dname.rs
@@ -7,6 +7,7 @@ use super::super::octets::{
 use super::super::octets::{DeserializeOctets, SerializeOctets};
 use super::builder::{DnameBuilder, FromStrError};
 use super::label::{Label, LabelTypeError, SplitLabelError};
+use super::parsed::ParsedDname;
 use super::relative::{DnameIter, RelativeDname};
 use super::traits::{ToDname, ToLabelIter};
 #[cfg(feature = "master")]
@@ -621,6 +622,17 @@ where
     fn octets_from(source: Dname<SrcOctets>) -> Result<Self, ShortBuf> {
         Octets::octets_from(source.0)
             .map(|octets| unsafe { Self::from_octets_unchecked(octets) })
+    }
+}
+
+impl<SrcOctets, Octets> OctetsFrom<ParsedDname<SrcOctets>> for Dname<Octets>
+where
+    SrcOctets: AsRef<[u8]>,
+    Octets: FromBuilder,
+    <Octets as FromBuilder>::Builder: EmptyBuilder,
+{
+    fn octets_from(source: ParsedDname<SrcOctets>) -> Result<Self, ShortBuf> {
+        source.to_dname().map_err(|_| ShortBuf)
     }
 }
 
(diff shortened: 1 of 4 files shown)


partim:
Thanks for the PR!

Looks like we are having MSRV issues in dependencies again. Could you merge main so the checks go green, both here and in #127?

src/base/name/dname.rs
```diff
@@ -624,6 +625,17 @@ where
     }
 }
 
+impl<SrcOctets, Octets> OctetsFrom<ParsedDname<SrcOctets>> for Dname<Octets>
+where
+    SrcOctets: AsRef<[u8]>,
+    Octets: FromBuilder,
+    <Octets as FromBuilder>::Builder: EmptyBuilder,
```
vavrusa, line 632: This is actually a blanket implementation from the previous PR. But it also converts `ParsedDname` -> `Dname`. I think it can either work like this, or or something like `to_dname::<SrcOctets>()?.octets_into()`.

partim: Isn’t just `to_dname` enough here? If this is about the error type – not sure why that is `PushError` in `ToDname::to_dname`. The type implementing `ToDname` should always have a valid domain name, so the only error is a `ShortBuf`. Unless I am missing something, I happily change the type in the trait.

partim:
For my own understanding: Is this PR because `Bytes` doesn’t provide all necessary conversions via the `From` trait?

src/base/name/dname.rs
partim: Feels like there should be a better way to do this: if the name isn’t compressed, this can be done very cheaply for certain types. Maybe have a method `flatten` on `ParseDname<_>` and `Record<ParseDname<_>, _>`?

vavrusa: I can do that too! So how is flatten different from ToDname? It should be possible to optimize ToDname the same way (conversion is cheaper when as_flat_slice is some or octets can be shallow copied/referenced).

Should flatten be something like Dname::as_octets but return a an Option? (so for example ParsedDname<&Bytes> should return Option<&Bytes> which *could* be cheap to convert into Dname<Bytes> or Dname<&Bytes> as ParsedDname guarantees that the octets are a valid Dname). Is that what you mean? Or flatten would convert both type (ParsedDname -> Dname) and octets type (like this method)?

For me this is just a shortcut for building records from `&[u8]` and I rarely want to shallow copy Bytes backed records, so I'm also happy to drop this particular change, as it's just a convenience.

partim: The idea would be that if the parsed name is not compressed, `flatten` can simply return with `self.parser.parse_octets(...)` while otherwise it constructs the name via `Dname::to_dname`. I think this should be possible with slightly extravagant trait bounds – you’d need `<Ref as OctetsRef>::Range` to be `FromBuilder` plus some more things to tie it together.

partim: How do you want to proceed? I want to merge #130 for other work I am doing but there’s some overlap, so I would rather merge this PR first to avoid things getting too messy.

vavrusa: I haven't been able to get back to this yet but let me decline it for now. I think I can make do without the nicer interface (or add extension traits to my library for record building with octets conversion) for now and will resubmit this if needed.

partim: Thanks! Feel free to propose or request something at any time! I might implement the two proposed flatten methods anyway – that seems like a useful thing to have.

PR #137, opened 2022-06-18
Drop two unnecessary uses of the futures crate

glts:
The proposed change removes two (as far as I can see) unnecessary uses of the futures crate in the `resolv` module. They can be expressed using the standard library. It seems to me that those are simply two instances that used to be but are no longer necessary (?).

Also, there is a stray module doc that I moved back to the top of the file, using `//!` instead of `///` as the comment marker.


glts:
I also noticed the abandoned file `src/resolv/lookup/records.rs`. It isn’t referenced anywhere and does not compile. Perhaps you’d like to delete it?

partim:
Good find; thank you for the PR!

I fixed the Clippy warnings in main, can you perchance merge main so the CI errors go away?

You are right, the `record.rs` file is indeed obsolet and can be deleted – do you want to do that as part of the PR?

glts:
Thank you for taking a look!

Done, and rebased.

partim (approved):

partim:
Looks good to go! Thank you again for the PR!

PR #139, opened 2022-08-31
Also declare MSRV in Cargo.toml

glts:
Follow-up to #128: Also declare the MSRV in Cargo.toml, for the benefit of dependent apps and libraries.

See https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field.

Thank you

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 name = "domain"
 version = "0.7.0-dev"
 edition = "2018"
+rust-version = "1.56.1"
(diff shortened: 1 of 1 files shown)


partim:
Thank you for the PR! As part of #142 I have to increase the Rust version to 1.59. Do you mind if I just introduce the field as part of that?

glts:
@partim Of course, thank you

partim:
Just to note it: The attribute was now added as part of #143.

PR #149, opened 2022-09-22
cargo: disable default chrono features

vavrusa:
These are not needed for the zone parser, and the newer chrono adds support for webassembly through bindgen + js_sys, which however only works for browsers and runtimes that implement JS runtime.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ path = "src/lib.rs"
 [dependencies]
 rand           = { version = "0.8", optional = true }
 bytes          = { version = "1.0", optional = true }
-chrono         = { version = "0.4.6", optional = true }
+chrono         = { version = "0.4.6", optional = true, default-features = false }
 futures        = { version = "0.3", optional = true }
 heapless       = { version = "0.7", optional = true }
 #openssl       = { version = "0.10", optional = true }


vavrusa:
@partim I noticed you have a seprate PR open to rework zonefile parser (which also drops chrono), so feel free to close this if you prefer

partim:
The new zone parser might be a bit before I’ll merge it, so I’ll just merge this PR to fix the issue you mentioned. Or, I will as soon as I have figured out why that macOS test fails. (On a train right now with limit connectivity …)

hunts:
The minimal chrono version v0.4.0 pulls in num v0.1.0 which does not compile in Rust nightly. 

```
domain v0.7.0
├── bytes v1.0.0
├── chrono v0.4.0
│   ├── num v0.1.0
│   └── time v0.1.36
│       └── libc v0.2.71
├── futures v0.3.11
 ...
```

See also https://github.com/chronotope/chrono/blob/v0.4.0/Cargo.toml

We can try to bump the minimal of chrono to v0.4.1 which [removed](https://github.com/chronotope/chrono/commit/8f90f405d52b220dc04aded26ff21937a1355382) that dependency:

```                            
diff --git a/Cargo.toml b/Cargo.toml
index 436c2f7..1de9df1 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ path = "src/lib.rs"
 [dependencies]
 rand           = { version = "0.8", optional = true }
 bytes          = { version = "1.0", optional = true }
-chrono         = { version = "0.4", optional = true, default-features = false }
+chrono         = { version = "0.4.1", optional = true, default-features = false }
 futures        = { version = "0.3", optional = true }
 heapless       = { version = "0.7", optional = true }
 #openssl       = { version = "0.10", optional = true }
```

partim:
Thank you both!

PR #151, opened 2022-09-30
add trait FlattenInto

xofyarg:
Hi,

This is a followup for #126, could you please let me know what do you think when you got a moment? Thanks.

The FlattenInto trait can help in case of converting ParsedDname into Dname in nested structure like `Record<N1, ZoneRecordData<O, N2>>`, so the user does not need to enumerate all the rdata types.

### DIFF
--- a/src/base/name/parsed.rs
+++ b/src/base/name/parsed.rs
@@ -5,13 +5,14 @@
 
 use super::super::cmp::CanonicalOrd;
 use super::super::octets::{
-    Compose, FormError, OctetsBuilder, OctetsRef, Parse, ParseError, Parser,
-    ShortBuf,
+    Compose, EmptyBuilder, FormError, FromBuilder, OctetsBuilder, OctetsFrom,
+    OctetsRef, Parse, ParseError, Parser, ShortBuf,
 };
 use super::dname::Dname;
 use super::label::{Label, LabelTypeError};
 use super::relative::RelativeDname;
 use super::traits::{ToDname, ToLabelIter};
+use super::PushError;
 use core::{cmp, fmt, hash};
 
 //------------ ParsedDname ---------------------------------------------------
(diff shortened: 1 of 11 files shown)


partim:
Thank you for the PR!

I wonder if we can get away without having a dedicated trait for this in rather offer a `flatten_into` method on the types that use a `ParsedDname`? Looking briefly, the only tricky bit seems to be record data. Maybe we can add `flatten_into` to the `ParseRecordData` trait?

xofyarg:
I think the trait is not really necessary, but easy to enforce that the method is implemented correctly everywhere. I updated the PR, and removed the trait. Although I didn't find a proper way to add a `flatten_into` method to `Record`, but I'm Ok without it. Please let me know what do you think.

xofyarg:
I can drop the `ToDname`, and use `ParsedDname` instead.

> What would then also be good if the result uses Dname<Ref::Range> in the above example. Then you can simply copy out the range in case the name isn’t compressed rather than constructing a new name like ToDname does.

Not quit sure about this though, if we'd like to optimize `flatten_into` for case that the name is not compressed, then the result would be different when compression is used. I think it's situational only in case of converting `ParsedDname<&Bytes>` to `Dname<Bytes>` so that the `ToDname` could be saved in uncompressed case. Or if I missed anything?

partim:
A `Cow<[u8]>` would be another option to keep uncompressed names cheap.

Can we use `OctetsFrom` for this purpose? The method’s name of `flatten_into` suggests that that might be the case. Require the target’s octets type to be `OctetsFrom<Ref::Range> + OctetsBuilder`?

xofyarg:
I think I get your idea now. PR updated.

partim:
Looking great now! Again, thank you for the PR!

PR #176, opened 2023-02-16
Refactor SVCB record data.

partim:
This PR refactors the SVCB record data to work along the same lines as all the other TLV-style types in the crate. The biggest user-facing change creating and building values. Since values are composed with their length prefixed, all value types now ensure that their data is never longer than 65,535 octets. This requires that creating and building produces results rather than happily taking any kind of octets sequence or appending.

This PR does not add scanning of SVCB/HTTPS records – mostly for time reasons.

This PR renames `base::iana::SvcbParamKey` to `base::iana::SvcParamKey` to match with the terminology used in the Internet draft.


partim:
@xofyarg Since you contributed the original SVCB implementation, perhaps you might want to have a look at this new version and check if it contains everything you need and doesn’t break things for you in too bad a way?

xofyarg:
Hi @partim,

Thanks for the heads up, the API looks good to me, only some trivial requests:
- Maybe add a explicit constructor method for octets wrappers like `DohPath`, for instance `DohPath::from_str`? As `DohPath::from_octets` looks like to compose a value from wire data(although they are identical).
- May I have `std::error::Error` implemented(under std feature) for the new error types? So I could use `anyhow::Context`.

Thanks again for the clean up!

partim:
Thanks for the feedback!

I’ll add both. (I actually realized I had forgotten to add all the bling to the new error types when I did the `FromStr` impl for `RelativeDname` yesterday.)

partim:
For the dohpath, I’ll just check that the string isn’t too long and otherwise accept anything. Actually parsing the URI template feels too difficult.

partim:
Turns out the draft uses slightly different terms than what we used in the code before. Most breakingly, the key type is `SvcParamKey` (i.e., starting with just `Svc` rathern than `Svcb`). This goes for most other types as well – but since they are new it doesn’t matter too much.

In any case, I decided to rename types and traits to match with the terms used in the draft to avoid confusion.

PR #180, opened 2023-02-22
Add hint for `Dname::to_string`

WhyNotHugo:


partim:
Could you add ’or by using its `Display` implementation’?

(I find it important to indicate how you can do this without allocations as well …)

partim:
Perfect! Thank you!

Issue #181, opened 2023-02-23
Error types don't implement std::error::Error


WhyNotHugo:
This makes it hard to handle errors externally. E.g.: this doesn't work:

```rs
    let err : Box<dyn std::error::Error> = Box::from(SrvError::LongName);
```

Due to:

```
the trait `core::error::Error` is not implemented for `SrvError`
```

This also breaks usability with libraries like `thiserror` or `anyhow`.

There mainly two ways around this:

1. Manually impl the required traits (mainly `[Display](https://doc.rust-lang.org/nightly/core/fmt/trait.Display.html)` is missing).
2. Use `thiserror` with helper macros to generate these at build time.

It might be a good idea to add [`assert_impl_all(X, std::error::Error)`](https://docs.rs/static_assertions/latest/static_assertions/macro.assert_impl_all.html) for all error types.

Note: in case you want to support no_std as mentioned in #179, then `core::error::Error` would be correct. They are both the same type.

partim:
They are supposed to impl `std::error::Error` if the `std` feature is enabled. If they aren’t, then that is an oversight and should be fixed.

Currently, we can’t use `core::error::Error` because it is unstable.

WhyNotHugo:
> They are supposed to impl std::error::Error if the std feature is enabled. If they aren’t, then that is an oversight and should be fixed.

I'll try and produce a patch for this specific one.

> Currently, we can’t use core::error::Error because it is unstable.

Ah, I missed that. It might be some time before it's stable and reasonable to push the MSRV.

PR #183, opened 2023-02-24
Implement std::error::Error for SrvError

WhyNotHugo:
Fixes: https://github.com/NLnetLabs/domain/issues/181


WhyNotHugo:
Any thoughts on using [`thiserror`](https://docs.rs/thiserror/latest/thiserror/) instead? It could reduce quite some overhead

partim:
I think for a library having fewer dependencies at the price of having the add a few more boilerplate manually is an okay trade-off.

WhyNotHugo:
Updated. Also replaced `std::fmt::Formatter` with just `fmt::Formatter`.

partim:
Thank you! Looks good to go now!

PR #184, opened 2023-02-24
Minor improvements to the lookup_srv docs

WhyNotHugo:
The function with the same name is pretty well documented, but not easily discoverable when looking at the method description.

### DIFF
--- a/src/resolv/lookup/srv.rs
+++ b/src/resolv/lookup/srv.rs
@@ -48,18 +48,20 @@ type OctetsVec = Vec<u8>;
 /// `service`.
 ///
 /// The value returned upon success can be turned into a stream of
-/// `ResolvedSrvItem`s corresponding to the found SRV records, ordered as per
+/// [`ResolvedSrvItem`]s corresponding to the found SRV records, ordered as per
 /// the usage rules defined in [RFC 2782]. If no matching SRV record is found,
 /// A/AAAA queries on the bare domain name `name` will be attempted, yielding
 /// a single element upon success using the port given by `fallback_port`,
 /// typcially the standard port for the service in question.
 ///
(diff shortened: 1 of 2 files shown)


src/resolv/lookup/srv.rs
```diff
@@ -47,18 +47,20 @@ type OctetsVec = Vec<u8>;
...
-/// `ResolvedSrvItem`s corresponding to the found SRV records, ordered as per
+/// [`ResolvedSrvItem`]s corresponding to the found SRV records, ordered as per
 /// the usage rules defined in [RFC 2782]. If no matching SRV record is found,
 /// A/AAAA queries on the bare domain name `name` will be attempted, yielding
 /// a single element upon success using the port given by `fallback_port`,
 /// typcially the standard port for the service in question.
 ///
 /// Each item in the stream can be turned into an iterator over socket
-/// addresses as accepted by, for instance, `TcpStream::connect`.
+/// addresses as accepted by, for instance, [`TcpStream::connect`]
```
partim, line 57: Looks like you accidentally deleted the sentence’s period.

src/resolv/stub/mod.rs
```diff
@@ -141,6 +141,7 @@ impl StubResolver {
         search_host(&self, qname).await
     }
 
+    /// See the documentation for the [`lookup_srv`] function for details.
```
partim, line 144: Could you add a short one-line summary before this note? Something trivial like ‘Performs an SRV lookup using the stub resolver’ will do.

partim (changes requested):
Thank you for the PR!

I’ve got two minor nit-picks inline.

partim (approved):
Looking good now. Thank you again for the PR!

PR #186, opened 2023-02-28
doc: Change some information in the example

ZhouYixun:
This error occurs when running the sample before changing

```
error[E0107]: this enum takes 1 generic argument but 2 generic arguments were supplied
  --> src\main.rs:92:50
   |
92 |     for option in response.opt().unwrap().iter::<AllOptData<_, _>>() {
   |                                                  ^^^^^^^^^^    - help: remove this generic argument
   |                                                  |
   |                                                  expected 1 generic argument
   |
note: enum defined here, with 1 generic parameter: `Octets`
```

I am learning rust. If there is something wrong, thank you for pointing out.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -89,7 +89,7 @@ fn main() {
     }
 
     // Display is only implemented for some OPTs at the time of writing
-    for option in response.opt().unwrap().iter::<AllOptData<_, _>>() {
+    for option in response.opt().unwrap().iter::<AllOptData<_>>() {
         let opt = option.unwrap();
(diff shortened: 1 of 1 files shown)


partim:
Thank you for the PR! Good find!

partim:
Hang on, something is weird now. `AllOptData` has two type arguments. Not sure how you got your original error?

ZhouYixun:
Maybe there is an unknown error in my local environment. I will check it again. Thank you for your open source project. It's great

PR #199, opened 2023-04-17
Implement Debug for Message.

partim:
This PR adds an implementation of `Debug` for `base::message::Message`. The implementation only prints the contents of the message’s header.


Philip-NLnetLabs (approved):
Looks good to me.

PR #200, opened 2023-04-20
Add methods to make or convert into canonical domain names.

### LINKED ISSUE #182: converting `Dname` to lowercase
I have an instance of possibly mixed-case Dname and I would like to get the lowercase version of it. Ideally, I would call some method like `Dname::into_lowercase()` to obtain it. 

I have not found an easy way to do that with the existing API, but I might be missing something. So far, I just do the conversion like this:

```rust
let dname = Dname::bytes_from_str("eXaMPLe.Com").unwrap();
let dname_lowercase: Dname<Vec<u8>> = Dname::from_octets(
    dname
        .as_octets()
        .iter()
        .map(u8::to_ascii_lowercase)
        .collect(),
).unwrap();
```

It works, but it's a little cumbersome. Obtaining a lowercase version of a Dname seems like a common enough use-case to warrant some library function.

partim:
This PR adds `Dname::make_canonical` and `RelativeDname::make_canonical` to convert a domain name into its canonical form in place as well as `ToDname::to_canonical_dname` and `ToRelativeDname::to_canonical_relative_dname` to produce new canonical domain names.

Fixes #182.

### DIFF
--- a/src/base/name/dname.rs
+++ b/src/base/name/dname.rs
@@ -279,14 +279,7 @@ impl<Octs: ?Sized> Dname<Octs> {
     where
         Octs: AsMut<[u8]>,
     {
-        let mut slice = self.0.as_mut();
-
-        while !slice.is_empty() {
-            let (head, tail) =
-                Label::split_from_mut(slice).expect("invalid dname");
-            head.make_canonical();
-            slice = tail;
-        }
+        Label::make_slice_canonical(self.0.as_mut());
     }
 }
 
--- a/src/base/name/dname.rs
+++ b/src/base/name/dname.rs
@@ -273,6 +273,21 @@ impl<Octs: ?Sized> Dname<Octs> {
     {
         unsafe { Dname::from_slice_unchecked(self.0.as_ref()) }
     }
+
+    /// Converts the domain name into its canonical form.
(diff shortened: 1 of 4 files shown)


Philip-NLnetLabs (commented):
Two comments: make_canonical appears twice. Maybe the core should be a common function? Maybe make_canonical could be implemented using an iterator over the labels?

partim:
It kind of already is by way of the new `Label::split_from_mut`. If it were to be an actual iterator, we’d have to have a new one that has a `&mut Label` as an item which would creep into everything. Not sure that is worth the effort.

Philip-NLnetLabs:
Another use of a &mut Label iteration would be the xor 0x20 trick to add extra randomization

partim:
Ah, yes. So it might make sense to have that iterator then. I shall add it.

partim:
Hm. I can’t quite get the iterator to work do to lifetime issues. I did follow your first suggestion, though, and moved to actual implementation of `make_canonical` to a private `Label::make_slice_canonical` so it only exists once.

Philip-NLnetLabs (commented):
No further comments

Philip-NLnetLabs (approved):

PR #201, opened 2023-04-24
feat: MAX_LENGTH constants

CrabNejonas:
This adds associated constants to all structs that have size limitations such as Dname, Label or CharStr

### DIFF
--- a/src/base/charstr.rs
+++ b/src/base/charstr.rs
@@ -56,6 +56,9 @@ use std::vec::Vec;
 pub struct CharStr<Octs: ?Sized>(Octs);
 
 impl<Octs: ?Sized> CharStr<Octs> {
+    /// Character strings have a maximum length of 255 octets.
+    pub const MAX_LENGTH: usize = 255;
+
     /// Creates a new empty character string.
     pub fn empty() -> Self
     where
@@ -124,7 +127,7 @@ impl CharStr<[u8]> {
 
     /// Checks whether an octets slice contains a correct character string.
     fn check_slice(slice: &[u8]) -> Result<(), CharStrError> {
-        if slice.len() > 255 {
+        if slice.len() > Self::MAX_LENGTH {
             Err(CharStrError)
         } else {
             Ok(())
@@ -283,7 +286,7 @@ where
             );
         let mut chars = s.chars();
         while let Some(symbol) = Symbol::from_chars(&mut chars)? {
-            if builder.len() == 255 {
+            if builder.len() == Self::MAX_LENGTH {
                 return Err(FromStrError::LongString);
             }
             builder.append_slice(&[symbol.into_octet()?])?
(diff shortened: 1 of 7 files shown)


src/base/charstr.rs
```diff
@@ -56,6 +56,9 @@ use std::vec::Vec;
 pub struct CharStr<Octs: ?Sized>(Octs);
 
 impl<Octs: ?Sized> CharStr<Octs> {
+    /// Character strings have a maximum length of 255 octets.
+    pub const MAX_LENGTH: usize = 255;
```
partim, line 60: If you define the constant for a concrete type, for instance, `CharStr<()>`, you avoid having to use the turbofish in some cases and can always refer to it as `CharStr::MAX_LENGTH` – though not `Self::MAX_LENGTH` anymore, but I think that’s fine as these names should be fairly stable.

(This goes for all the other generic types as well.)

src/base/name/builder.rs
```diff
@@ -372,7 +373,7 @@ where
         Builder: FreezeBuilder,
     {
         self.end_label();
-        if self.len() + usize::from(origin.compose_len()) > 255 {
+        if self.len() + origin.compose_len() as usize > Dname::<Builder::Octets>::MAX_LENGTH  {
```
partim, line 376: We try to avoid the `as` operator as much as possible. Can you rewrite all of these using `usize::from`?

src/base/name/chain.rs
```diff
@@ -42,7 +43,7 @@ pub struct Chain<L, R> {
 impl<L: ToLabelIter, R: ToLabelIter> Chain<L, R> {
     /// Creates a new chain from a first and second name.
     pub(super) fn new(left: L, right: R) -> Result<Self, LongChainError> {
-        if left.compose_len() + right.compose_len() > 255 {
+        if left.compose_len() + right.compose_len() > 255 { // TODO can't infer a specific type for Dname here
```
partim, line 46: I think you can use `Dname::MAX_LENGTH` here.

Or, technically, this should be 254 if `R` doesn’t end with the root label so that the root can be appended without errors.

src/base/charstr.rs
partim, line 60: I’m now also wondering if the constant should be called `MAX_LEN`, given that the method is called `len`?

partim (changes requested):
Thank you for the PR! Adding a definition for what maximum length is allowed is an excellent idea!

I have a few minor notes inline.

src/base/charstr.rs
CrabNejonas: ah nice yeah I was wondering how to do that

src/base/name/builder.rs
CrabNejonas: done!

src/base/charstr.rs
CrabNejonas: okay fair enough 👍 done!

partim:
Once again, thank you for the PR!

PR #210, opened 2023-06-05
Fix scanning of the root name.

partim:
This PR fixes scanning (i.e., reading from presentation format) of domain names that are just the root label. It also fixes how they are displayed: previously they were empty, now they are displayed as a dot. It also adds a new method `Dname::fmt_with_dot` that returns a placeholder type that displays the name with a dot at the end.

### DIFF
--- a/src/base/name/builder.rs
+++ b/src/base/name/builder.rs
@@ -195,8 +195,7 @@ where
             }
             self.end_label();
             Ok(())
-        } else if
-            matches!(sym, Symbol::SimpleEscape(b'['))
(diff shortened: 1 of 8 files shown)


src/base/name/builder.rs
```diff
@@ -184,6 +184,31 @@ where
         Ok(())
     }
 
+    /// Pushes a symbold to the end of the domain name.
```
Philip-NLnetLabs, line 187: Spelling error 'symbold'

Philip-NLnetLabs (changes requested):
Looks good except for one spelling error

PR #212, opened 2023-06-06
Fix the Base 64 symbol converter.

partim:
This PR fixes `util::base64::SymbolConverter` to also include the final group in the output if there is padding.

### DIFF
--- a/src/utils/base64.rs
+++ b/src/utils/base64.rs
@@ -368,19 +368,29 @@ impl SymbolConverter {
 
         if self.next == 4 {
             self.output[0] = self.input[0] << 2 | self.input[1] >> 4;
-            if self.input[2] != 0x80 {
-                self.output[1] = self.input[1] << 4 | self.input[2] >> 2;
-            }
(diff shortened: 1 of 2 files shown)


src/utils/base64.rs
```diff
@@ -368,19 +368,29 @@ impl SymbolConverter {
...
         if self.next == 4 {
             self.output[0] = self.input[0] << 2 | self.input[1] >> 4;
-            if self.input[2] != 0x80 {
-                self.output[1] = self.input[1] << 4 | self.input[2] >> 2;
-            }
-            if self.input[3] != 0x80 {
-                if self.input[2] == 0x80 {
-                    return Err(Error::custom("trailing Base 64 data"));
+
+            if self.input[2] == 0x80 {
```
Philip-NLnetLabs, line 372: It would be nice if this could be a constant.

Philip-NLnetLabs (approved):
It would be nice if 0x80 and 0xF0 could be constants

PR #216, opened 2023-07-17
Move `flatten_into` for flattening domain names into a trait.

partim:
This PR moves the `flatten_into` method present on various types that either are or contain domain names into a `FlattenInto` trait.

Turns out that flattening is not only necessary for parsed names as I originally thought but also in chained names such as those produced by the current version of the zonefile parser.

The PR removes the method from all types that aren’t or don’t contain domain names and only had it to be able to make the big record data type enums.

With apologies to @xofyarg who originally proposed this as a trait and, turns out, was right.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ name = "domain"
 path = "src/lib.rs"
 
 [dependencies]
-octseq         = "0.2"
+octseq         = { git = "https://github.com/NLnetLabs/octseq.git", branch = "append-error-alias" }
 time           = "0.3.1"
 
 rand           = { version = "0.8", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ name = "domain"
 path = "src/lib.rs"
 
 [dependencies]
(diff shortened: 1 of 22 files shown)


xofyarg:
Hi @partim,

May I know the plan for merging this PR? We are going to upgrade domain to the 0.8 on our side, and want to potentially reduce the future changes to the downstream code.

partim:
If it helps you, we can do that fairly soon. We can also do an 0.9.0 release with it if that makes your life easier. I’m currently on vacation, though, so I would want to wait with that until the mid-October.

xofyarg:
Sorry to interrupt your vacation, a new version like 0.9.0 would work better for us.

Also what do you think to re-export octseq from domain? As sometimes it's pinned to a "-dev" version, that we have to sync to it at downstream.

partim:
Okay. I’ll merge the branch now and then wait with 0.9.0 until I am back in the office. Looks like we also need an octseq release, so really don’t want to rush this.

Re-exporting octseq can be done, maybe we can make a `dep` module where this can live and not pollute the crate root?

xofyarg:
A submodule to re-export sounds good to me. Thanks for merging this PR.

PR #217, opened 2023-07-19
Introduce generic basis for DNSSEC signing.

partim:
This PR moves the basics for DNSSEC signing to traits so a zone store can implement signer in whatever way is best for it.


bal-e:
Should we close this now?

partim:
This has been superseded with other work … so, closing.

PR #220, opened 2023-08-23
docs: add example for RelativeDname::from_slice

WhyNotHugo:


src/base/name/relative.rs
```diff
@@ -136,6 +136,12 @@ impl RelativeDname<[u8]> {
     }
 
     /// Creates a relative domain name from an octet slice.
+    ///
```
partim, line 139: Since this indeed seems to be a recurring mistake, perhaps we should add a sentence on all of these that the slice needs to in wire format, i.e., no dots but each label prefixed with its length?

partim:
Thank you for the PR! Since this indeed seems to be a recurring mistake, perhaps we should add a sentence on all of these (also `Dname<_>`) that the slice needs to in wire format, i.e., no dots but each label prefixed with its length?

partim (commented):
(Clicked “cancel review” as I replaced it with above comment, but it still posted it …)

WhyNotHugo:
I've included an example for `domain::base::name::Dname::from_slice`. I'm not sure where else this same advise is applicable.

PR #221, opened 2023-08-23
Remove unneeded sub `cfg` with only one condition

WhyNotHugo:
Fixes current failures in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        rust: [1.65.0, stable, beta, nightly]
+        rust: [1.67.0, stable, beta, nightly]
     steps:
     - name: Checkout repository
       uses: actions/checkout@v1
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -109,7 +109,7 @@
 #![allow(clippy::uninlined_format_args)]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
-#[cfg(any(feature = "std"))]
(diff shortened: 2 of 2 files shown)


partim:
> error: package `time v0.3.27` cannot be built because it requires rustc 1.67.0 or newer

Ah, lovely. I wish fundamental libraries wouldn’t do this sort of thing. I guess we’ll have to move to 1.67.0 then.

WhyNotHugo:
> time is guaranteed to compile with the latest stable release of Rust in addition to the two prior minor releases. For example, if the latest stable Rust release is 1.70, then time is guaranteed to compile with Rust 1.68, 1.69, and 1.70.

From: https://github.com/time-rs/time

If you want to keep an MSRV of 1.67.0, you're going to need to pin an older version of the `time` library.

WhyNotHugo:
time v0.3.23 is the last version to support Rust 1.65.0

partim:
Not sure pinning the version is a good idea – you would want to use the newest version if your compiler supports it. For now, upgrading the minimal Rust version to 1.67 is probably better.

WhyNotHugo:
Sure. Note that the latest version is not guaranteed to work with 1.67.0, it does so by mere fortune and this might change any time.

WhyNotHugo:
All green!

image

partim:
> Sure. Note that the latest version is not guaranteed to work with 1.67.0, it does so by mere fortune and this might change any time.

Yeah. I think this is mostly an issue for the CI where we test that the promised MSRV actually works. Maybe we should change that into a `cargo check` with the oldest possible library versions. (I think there was a cargo thing for that, have to dig it up.)

partim:
Anyhow, thanks for the PR! Looking good!

PR #224, opened 2023-09-17
Improve documentation for srv module

WhyNotHugo:


src/resolv/lookup/srv.rs
```diff
@@ -120,6 +121,11 @@ impl FoundSrvs {
     }
 
     /// Converts the value into an iterator over the found SRV records.
+    ///
+    /// If results were found, this returns them in the order prescribed by the SRV records.
+    ///
+    /// If not results were found, the iterator will yield a single entry with the bare host and
+    /// the default fallback port.
```
WhyNotHugo, line 128: Note that the documentation for the **private** member `FoundSrv::items` already mentions this, but this doesn't end up visible anywhere in the documentation.

partim:
Thanks for the PR!

PR #228, opened 2023-09-20
Mark function without side effects as must_use

WhyNotHugo:
This hint will make the compiler warn consumers of this library if they ever call one of these functions without using its return value.

### DIFF
--- a/src/base/charstr.rs
+++ b/src/base/charstr.rs
@@ -62,6 +62,7 @@ impl CharStr<()> {
 
 impl<Octs: ?Sized> CharStr<Octs> {
     /// Creates a new empty character string.
+    #[must_use]
     pub fn empty() -> Self
     where
         Octs: From<&'static [u8]>,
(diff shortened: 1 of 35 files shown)


partim:
Thank you for the PR!

Is this idiomatic use of the `#[must_use]` attribute? I so far only considered it for cases where ignoring the return type likely results in problems not merely waste.

WhyNotHugo:
> I so far only considered it for cases where ignoring the return type likely results in problems not merely waste.

The docs aren't too specific:

https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute

The places where it's used here are functions that have no side effect but construct a new value. Calling them without using the return value is usually a programmer mistake (it should either be assigned or the call deleted entirely).

partim:
Hm. You are certainly correct. And I suppose this isn’t all that different from the unused variables warnings.

I’ll merge it. But, I think this is a breaking change, so I’ll stick the label on it and keep it open until we get closer to a breaking release. (I am trying to avoid the issues we had with having to forward port changes.)

WhyNotHugo:
Exactly! It's more a hint that anything else.

I don't think that it's breaking; at most, consumers will see a warning. And this is in case they're doing something like `CharStr::builder();` without assigning the builder to any variable.

partim:
It might not strictly be breaking, but it has the potential to break CI runs without any changes and I really detest that.

WhyNotHugo:
It's your call, this is trivial to rebase anyway.

PR #229, opened 2023-09-23
add ZONEMD support

xofyarg:
This commit add support for ZONEMD record type, described in RFC 8976.

Since root is rolling it out, there's no reason the standard is missing from the library :-p

### DIFF
--- a/src/rdata/zonemd.rs
+++ b/src/rdata/zonemd.rs
@@ -1,0 +1,449 @@
+//! ZONEMD record data.
+//!
+//! The ZONEMD Resource Record conveys the digest data in the zone itself.
+//!
+//! [RFC 8976]: https://tools.ietf.org/html/rfc8976
+
+use crate::base::cmp::CanonicalOrd;
+use crate::base::iana::Rtype;
+use crate::base::name::PushError;
+use crate::base::rdata::{ComposeRecordData, RecordData};
+use crate::base::scan::{Scan, Scanner};
+use crate::base::serial::Serial;
+use crate::base::wire::{Composer, ParseError};
+use crate::utils::base16;
+use core::cmp::Ordering;
+use core::{fmt, hash};
+use octseq::octets::{Octets, OctetsFrom, OctetsInto};
+use octseq::parse::Parser;
+
+// section 2.2.4
+const DIGEST_MIN_LEN: usize = 12;
+
+/// The ZONEMD Resource Record conveys the digest data in the zone itself.
+#[derive(Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Zonemd<Octs: ?Sized> {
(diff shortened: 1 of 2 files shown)


src/rdata/zonemd.rs
```diff
@@ -0,0 +1,447 @@
...
+        self.digest.as_ref().cmp(other.digest.as_ref())
+    }
+}
+
+/// The Scheme is an 8-bit unsigned integer that identifies the
+/// methods by which data is collated and presented as input to the
+/// hashing function.
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[repr(u8)]
```
partim, line 302: Does that actually work? I would assume the compiler needs to be able to distinguish between the unassigned and private variant both of which can carry a full `u8`.

src/rdata/zonemd.rs
```diff
@@ -0,0 +1,447 @@
...
+        }
+        match self.algo.cmp(&other.algo) {
+            Ordering::Equal => {}
+            other => return other,
+        }
+        self.digest.as_ref().cmp(other.digest.as_ref())
+    }
+}
+
+/// The Scheme is an 8-bit unsigned integer that identifies the
```
partim, line 297: Could you start the docstring with a single one-line summary? This will then show in the index pages of the module and gives a hint as to what the type is for …

src/rdata/zonemd.rs
```diff
@@ -0,0 +1,447 @@
...
+        match n {
+            0 | 255 => Self::Reserved,
+            1 => Self::Simple,
+            2..=239 => Self::Unassigned(n),
+            240..=254 => Self::Private(n),
+        }
+    }
+}
+
+/// The Hash Algorithm is an 8-bit unsigned integer that identifies
```
partim, line 332: One-line summary here as well.

partim (commented):
Thank you for the PR! It looks good with only a few minor comments.

src/rdata/zonemd.rs
xofyarg: It only works with field-less enums. Removed.

xofyarg:
Thanks for the review, @partim. I've addressed all the comments, please take another look.

src/rdata/zonemd.rs
partim: I don’t think it makes too much sense to add the value to `Simple`, given that you convert explicitly in the `From<_>` impls.

xofyarg: You are right. These are removed.

partim (approved):

partim:
Thank you for the changes! Looking good now.

Given that the record type enums are marked as non-exhaustive, I think this isn’t a breaking change and can be merged right away.

PR #231, opened 2023-10-09
add Composer implementation for reference types

xofyarg:
The use case is to share the underlying Composer buffer between builders to avoid allocation, via a thread local variable for example. Without the reference type implements, the ownership could be transferred by memory swap operation, but that's not quite ideal. Plus some methods like MessageBuilder::from_target would drop the Composer on error, which makes it even difficult to share.


partim:
Could this just be a blanket impl for any mut ref to a composer? The compiler seems to accept that.

xofyarg:
Even better. The PR has been updated.

partim:
This isn’t quite right – the trait functions should defer to the functions in the referenced type and not use the default impls. In my quick test, I had something like this:

```
    fn append_compressed_dname<N: ToDname + ?Sized>(
        &mut self,
        name: &N,
    ) -> Result<(), Self::AppendError> {
        Composer::append_compressed_dname(*self, name)
    }
```

xofyarg:
My bad, I got up early this morning, with my brain left behind. Thought the `can_compress` could use the default implementation, but it's better to associate it with the type being referenced. Thanks for the gatekeeping.

PR #236, opened 2023-10-24
Introduce a `ComposeMessage` trait.

partim:
This PR proposes a more flexible way for handing request messages to the client transport layer (and probably other things as well). The new `ComposeMessage` has methods that provide separate access to those parts of a message that a transport needs to build its own final message around: the header and what it calls the data sections (i.e., the question section and the record sections).

This doesn’t provide a means to signal that there isn’t an OPT or TSIG record, but I think it’s fine to have that as an documented part of the contract rather than have separate types (I tried that and things get unwieldy real quick).

This PR doesn’t yet include a way for a client transport user to pass OPT options to the transport. I think we should do an inventory on the existing options and which are user-generated and which are transport-generated first. In addition, we should also consider that there might have to be TSIG (and SIG(0)?) shenanigans which, if we could support them as well, would be nice.

Perhaps we need an approach where there is a basic message from the user, then the transport builds its message around that, and finally hands that back to the user so they can turn it into the final message to be sent out. `ComposeMessage` could be used in all three steps, but maybe it needs dedicated OPT handling added?

### DIFF
--- a/src/base/message.rs
+++ b/src/base/message.rs
@@ -202,7 +202,7 @@ impl Message<[u8]> {
     /// The methods for header access rely on the octets being at least as
     /// long as a header. If the sequence is shorter, the behavior is
     /// undefined.
-    unsafe fn from_slice_unchecked(slice: &[u8]) -> &Self {
+    pub(super) unsafe fn from_slice_unchecked(slice: &[u8]) -> &Self {
         &*(slice as *const [u8] as *const Self)
     }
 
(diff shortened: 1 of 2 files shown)


partim:
Hm, or maybe this is all wrong? Maybe the transport should just hand its desired OPT content as input to a method in the `ComposeMessage` trait and let the trait implementer take care of including it in its own message?

PR #237, opened 2023-10-25
Support not sized octets in ZoneRecordData and UnknownRecordData

hunts:
These two record data impls shall also support not sized octets like the other implementations of concrete record types.

To allow for trading the parsed record that is not sized for a resource record. For example:

```
let msg: &Message<[u8]>: ...
let answer = msg.answer()?;

for parsed_rr in answer.filter_map(|r| r.ok()) {
    if let Ok(Some(rr)) = parsed_rr.into_record::<ZoneRecordData<_, _>>() {
        ...
    }
}
```

### DIFF
--- a/src/base/rdata.rs
+++ b/src/base/rdata.rs
@@ -404,7 +404,7 @@ impl<Octs: AsRef<[u8]>> RecordData for UnknownRecordData<Octs> {
     }
 }
 
-impl<'a, Octs: Octets> ParseRecordData<'a, Octs>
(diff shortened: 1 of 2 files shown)


partim:
Thank you for the PR!

You are, of course, correct. This looks like an oversight when we refactored this code.

hunts:
Thank you Sir. Hope we'll have a 0.9.1 soon :)

partim:
That certainly can be arranged.

PR #241, opened 2023-11-12
Remove Sized constraint from tsig module types.

torin-carey:
Currently the requirement of sized types means that `Message<[u8]>` cannot be used in TSIG operations using the `tsig` module.

These constraints can be relaxed.

### DIFF
--- a/src/tsig/mod.rs
+++ b/src/tsig/mod.rs
@@ -262,7 +262,7 @@ impl Key {
     }
 
     /// Checks whether the key in the record is this key.
-    fn check_tsig<Octs: Octets>(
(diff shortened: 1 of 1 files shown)


partim:
Thank you for the PR!

You are, of course, correct that this module should also work on `Message<[u8]>` – this was missed when changing the octets traits recently.

Please let us know if you need a release soon or if we can can wait a wee bit to collect a few more items.

torin-carey:
Thank you for such a fast review and pull!

I'm not in any rush to see a release, I'm happy to let some other changes pool

PR #242, opened 2023-11-12
Add AsRef<Message<[u8]>> implementation for Message<Octs>

torin-carey:
The reference type `&Message<[u8]>` makes for a suitable borrowed form for the `Message<Octs>` type.

This conversion requires unsafe to avoid the (in this case) unnecessary length check.


partim:
Thank you for the PR!

This look vaguely familiar – I actually added the same impl in #236. But no need to wait for that.

PR #245, opened 2023-12-05
Move FoundSrvs::into_stream into feature "stream"

WhyNotHugo:
This function (and only this function) depends on the `futures` create, which itself requires over a dozen other dependencies.

Move the function into its own feature, "stream". This allows enabling the "resolv" feature without the `futures` dependency tree, which substantially reduces unnecessary dependencies.


WhyNotHugo:
For reference, this reduces the amount of dependencies by **sixteen** for me:

```console
> cargo tree --prefix=none --features resolv | sort | uniq | wc -l
38

> cargo tree --prefix=none --features resolv,stream | sort | uniq | wc -l 
54
```

partim:
Thank you for the PR!

Switching to futures-util without default features results in 41 dependencies (basically the three futures crates). I feel like this might be acceptable in exchange for not having another feature?

WhyNotHugo:
Yeah, that sounds a lot better.

I tried removing features from futures, but didn’t realise I could just use that other one instead. I work on that and send a new PR.

PR #246, opened 2023-12-06
Replace `futures` dependency with `futures_util`

WhyNotHugo:
The latter is sufficient, and the former pulls over a dozen additional unused dependencies.

Closes: https://github.com/NLnetLabs/domain/pull/245


partim:
Thanks for the PR! Looking good!

(Minor nit for future PRs: We update the Changelog after merging because there’s often wonky merge conflicts – and we also have the PR number then.)

PR #247, opened 2023-12-22
Compile in `no_std` context

dzamlo:
Previously, the crates didn't compile in `no_std` context because some dependencies required std by default.


partim:
Thank you for the PR!

I will make a new release once I have also merged #248 which fixes a few more no-std issues.

PR #248, opened 2023-12-22
Fix `no_std`.

reitermarkus:
Compiling on `no_std` is currently broken since both `octseq` and `time` have default `std` features enabled.


reitermarkus:
Ah, I missed that https://github.com/NLnetLabs/domain/pull/247 is already open.

partim:
Thank you for the PR!

Minor issue: We need _time_ to be at least 0.3.1. Could you adjust that? And, I suppose we should also update _octseq_ to at least 0.3.2 to pull in its no-std fixes.

(I’m not sure why Github doesn’t want to run CI on your last commit.)

reitermarkus:
@partim, added the minimum versions.

partim:
Thanks! Looks like this also fixed the CI hickup.

I’ll merge the PR and make a new release.

PR #253, opened 2024-01-09
Fix displaying of chained domain names.

partim:
This PR fixes the `Display` implementation of `Chain<_,_>` so it considers all possible combinations of empty, non-empty, and root names. It now behaves similarly to the `Display` implementation of `Dname` in that it only adds a final dot to an absolute name if it is the root label only.  Similarly to `Dname`, it also adds `Chain::fmt_with_dot` which returns a temporary type that formats absolute names with a final dot.

Fixes #252.

### DIFF
--- a/src/base/name/chain.rs
+++ b/src/base/name/chain.rs
@@ -583,6 +583,11 @@ mod test {
     }
 
     /// Tests that displaying works as expected.
+    ///
+    /// The tricky bit is to produce to correct number of dots between the
+    /// left and the right part and at the end of the chain. This is made
+    /// difficult by empty relative names and absolute root names. So this
+    /// is what we are testing below in a number of combinations.
     #[test]
     fn display() {
         fn cmp<E: fmt::Debug, L, R>(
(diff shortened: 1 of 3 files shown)


phicoh (approved):

src/base/name/chain.rs
```diff
@@ -200,9 +215,27 @@ where
...
-impl<L: fmt::Display, R: fmt::Display> fmt::Display for Chain<L, R> {
+impl<L, R> fmt::Display for Chain<L, R>
+where
+    Self: ToLabelIter,
+{
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}.{}", self.left, self.right)
+        let mut empty = true;
+        for label in self.iter_labels() {
+            if label.is_root() {
```
ximon18, line 225: Only the last label should ever be a root, is that correct? Yet this is checked for every label, does that make sense?

partim: I can’t think of a way to only check on the last label. As in: how do I know a label is last? (This check is extremely cheap, basically checking if the label is empty.)

ximon18 (approved):

src/base/name/chain.rs
ximon18: Maybe using [`split_last()`](https://doc.rust-lang.org/std/primitive.slice.html#method.split_last) ?

partim: That and it also iterates over the whole thing to find the last label.

Philip-NLnetLabs (approved):

PR #254, opened 2024-01-12
Remove interop and ci-test features and set interop tests to ignore.

partim:
This PR removes the interop and ci-test features which were used to gate tests that rely on additional software being installed on the test system. This was brittle as it required new features to be added to the ci-test feature or they would not be tested in the CI workflow. Instead, we now simply add the `#[ignore]` attribute to these tests and can use `--all-features` without issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,10 +18,10 @@ jobs:
     - if: matrix.rust == 'stable'
       run: rustup component add clippy
     - if: matrix.rust == 'stable'
-      run: cargo clippy --all-features --tests --examples -- -D warnings
+      run: cargo clippy --all-features --all-targets -- -D warnings
     - if: matrix.rust == 'stable' && matrix.os == 'ubuntu-latest'
       run: cargo fmt --all -- --check
-    - run: cargo check --tests --examples --no-default-features
+    - run: cargo check --no-default-features --all-targets
     - run: cargo test --all-features
     - if: matrix.rust == 'nightly'
       run: |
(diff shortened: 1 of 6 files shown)


src/test/mod.rs
```diff
@@ -5,10 +5,9 @@
 //!
 //! The actual tests live with their respective modules in a sub-module called
 //! `interop`. Since they require additional software packages to be available
-//! and can be expensive, they are only run if the `"interop"` feature is
-//! given to `cargo test`.
+//! and can be expensive, they are tagged as ignored by default.
 //!
-#![cfg(all(test, feature = "interop"))]
+#![cfg(all(test, feature = "bytes", feature = "std"))]
```
ximon18, line 10: The line removed from `Cargo.toml` grouped the `bytes` and `ring` features into an `interop` feature yet here you replace `interop` with `bytes` and `std` - is that correct?

partim: I believe the code in this module only relies on those two features.

.github/workflows/ci.yml
```diff
@@ -18,14 +18,13 @@ jobs:
...
-    - run: cargo test --features=ci-test --verbose
+    - run: cargo check --tests --examples --no-default-features
+    - run: cargo test --all-features
     - if: matrix.rust == 'nightly'
       run: |
         cargo +nightly update -Z minimal-versions
-        cargo check --features=ci-test --verbose --all-targets
-        cargo test --features=ci-test
-      name: Check and test with minimal-versions
+        cargo check --all-features --all-targets
```
ximon18, line 29: Why check more targets here than above?

ximon18 (approved):

.github/workflows/ci.yml
partim: Fair point. I’ll change the other steps, too.

ximon18 (approved):

PR #255, opened 2024-01-23
Introduce simplified message data types.

partim:
This PR introduces a set of types for representing messages without the somewhat tedious generic octets sequences. Instead, all the types use `Bytes` as a compromise. The new `Message` keeps the elements of the four sections in vecs for easier manipulation.

I’ve chosen the name `plain` for this module since the types don’t have any type arguments. This is just a proposal and I am very open for alternative names.

As a side effect, this PR adds additional trait implementations and methods here and there.

This PR is a breaking change via _octseq_ needing a breaking change.


partim:
This PR is being replaced by a sequence of more atomic PRs.

PR #258, opened 2024-01-25
RFC 6891 extended RCODE conversion function fixes.

ximon18:
This PR fixes a couple of bugs in RFC 6891 extended RCODE conversion functions that prevent extended RCODE values > 15 being handled correctly.

This problem was observed while testing an implementation of [RFC 7873 Domain Name System (DNS) Cookies](https://datatracker.ietf.org/doc/html/rfc7873), when responding with extended RCODE `BADCOOKIE` (integer value 23).

23 is binary 0000_0001_0111 in 12-bit extended RCODE form, with 0000_0001 (1) being the extended part and 0111 (7) being the RFC 1035 DNS header RCODE part.

The wire format should have had the lowest bit of the OPT header extended RCODE field set but it was unset. This can be seen in the following Wireshark screenshot:

image
 
This PR fixes the problem by making the following two changes:

- `to_int()`: Like `from_int()`, work with the lowest 12-bits when converting to u16, not the lowest 4-bits.
- `to_parts()`: Like `from_parts()`, shift by 4-bits, not by 8-bits.

It also adds a new test called `optrcode_parts()` that fails without these changes and passes with them.

With the fix applied Wireshark shows the OPT header RCODE value being set correctly:

image

And `dig` was then able to correctly detect the BADCOOKIE RCODE and use the accompanying server cookie in subsequent requests, while without the fix the received server cookie was ignored.

### DIFF
--- a/src/base/iana/rcode.rs
+++ b/src/base/iana/rcode.rs
@@ -190,7 +190,7 @@ impl Rcode {
         }
     }
 
-    /// Returns the integer value for this rcode.
+    /// Returns the 4-bit integer value for this rcode.
     #[must_use]
     pub fn to_int(self) -> u8 {
         use self::Rcode::*;
@@ -527,7 +527,7 @@ impl OptRcode {
         }
     }
 
-    /// Returns the integer value for this rcode.
+    /// Returns the 12-bit integer value for this rcode.
     #[must_use]
     pub fn to_int(self) -> u16 {
         use self::OptRcode::*;
@@ -546,7 +546,7 @@ impl OptRcode {
             NotZone => 10,
             BadVers => 16,
             BadCookie => 23,
-            Int(value) => value & 0x0F,
+            Int(value) => value & 0x0FFF, // OPT RCODEs are 12 bits
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/base/iana/rcode.rs
```diff
@@ -190,7 +190,7 @@ impl Rcode {
         }
     }
 
-    /// Returns the integer value for this rcode.
+    /// Returns the 4-bit integer value for this rcode.
```
partim, line 193: Would it perhaps be better to keep the old text and add a line below saying that “Only the lower 4 bits of the returned octet are used by the rcode. The upper four bits are always zero.”?

src/base/iana/rcode.rs
```diff
@@ -854,3 +854,103 @@ impl From<OptRcode> for TsigRcode {
...
+
+//============ Tests =========================================================
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn optrcode_parts() {
+        // RFC 1035 defines a 4-bit RCODE in the main DNS header RCODE field.
```
partim, line 866: Perhaps this text should be in the doc string of OptRcode?

src/base/iana/rcode.rs
ximon18: Is that better?

partim: Yes to both ;)

src/base/iana/rcode.rs
ximon18: Is this better?

partim: Perfect!

partim (approved):

PR #264, opened 2024-02-05
Add SECURITY.md

AlexanderBand:


SECURITY.md
```diff
@@ -0,0 +1,31 @@
...
+previous major version is released, then support on these versions will only be
+provided for the remainder of the one-year-period.
+
+You can find detailed information on our software support policy here:
+
+https://www.nlnetlabs.nl/support/software-support-policy/
+
+## Reporting a Vulnerability
+
+We take security very seriously. If you have discovered a security vulnerability
```
maertsen, line 25:
```suggestion
If you have discovered a security vulnerability
```

SECURITY.md
```diff
@@ -0,0 +1,31 @@
...
+in one of our projects and you would like to report it to us, you can send an
```
maertsen, line 26:
```suggestion
in one of our projects and you would like to report it to us, please send an
```

SECURITY.md
```diff
@@ -0,0 +1,31 @@
...
+
+NLnet Labs adheres to the straightforward, semantic versioning scheme that is
+commonly used in the software industry.
+
+Support is provided in respect of the latest release, i.e. releases with the
+highest minor and patch version level. We do not backport security fixes to
+older (minor) versions. In the event a new major version is released (e.g.  from
+3.2.18 to 4.0.0), support will also be provided on the latest minor  version of
+the previous major version (3.2.18) for a period of one year from the release of
+the new major version (4.0.0).
```
maertsen, line 13:
```suggestion
the new major version (4.0.0), unless the previous major version number was 0 (the 'initial development phase').
```

maertsen (commented):
Two general observations:
  - the term 'support' is not defined here, nor on our website. The filename and first heading here implicitly conveys that 'support' is related to 'security', while the (same) text on the website is not in a security specific context and presumably is about where continued development happens.
  - the (general) statement about no backporting of security fixes may be read to conflict with the (specific) statement about support for the latest minor version of a previous major version. I am assuming that "specific beats general" and therefore this is an exception to that rule. But perhaps it is not. Regardless, I think it would be helpful to readers to clarify this.

partim:
I also stumbled over the lack of definition of “support.” I’m assuming it to mean security and bug fixes, but we should probably say that?

SECURITY.md
```diff
@@ -0,0 +1,31 @@
+# Security Policy
+
+## Supported Versions
+
+NLnet Labs adheres to the straightforward, semantic versioning scheme that is
+commonly used in the software industry.
```
tertsdiepraam, line 6: If we want to do this properly, we should probably set up `cargo-semver-checks` at some point.

PR #267, opened 2024-02-06
Reject empty TXT record data and never build it.

### LINKED ISSUE #260: Reject empty TXT record data.
`Txt::from_octets` and friends currently accept an empty octets sequence as record data. RFC 1035 says that TXT records contain  “[o]ne or more <character-string>s.” So we should really reject empty sequences and insist on a length of at least 1.

partim:
This PR changes the `rdata::Txt` type to reject empty record data as invalid.

RFC 1035 states that the TXT-DATA is “One or more \<character-string\>s”. This means that there will have to be at least the length octet of the first character string.

The PR also adjusts the `TxtBuilder` to never return a empty data. Since this requires being able to return an error in `TxtBuilder:.finish`, this is a breaking change. While at it, the PR also adds more control over the created character strings by making `close_char_str` public and adding a new `append_char_str` method.

The PR also improves the documentation for the TXT record data types.

Fixes #260.

### DIFF
--- a/src/rdata/rfc1035.rs
+++ b/src/rdata/rfc1035.rs
@@ -1615,7 +1615,7 @@ impl<Octs> Txt<Octs> {
     ///
     /// The `octets` sequence most contain correctly encoded TXT record
     /// data. That is, it must contain a sequence of at least one character
-    /// strings of at most 255 octets each preceded by a length octet. An
+    /// string of at most 255 octets each preceded by a length octet. An
     /// empty sequence is not allowed.
     ///
     /// Returns an error if `octets` does not contain correctly encoded TXT
@@ -1644,7 +1644,6 @@ impl Txt<[u8]> {
     ///
     /// The slice must contain correctly encoded TXT record data,
     /// that is a sequence of encoded character strings. See
-    ///
     pub fn from_slice(slice: &[u8]) -> Result<&Self, TxtError> {
         Txt::check_slice(slice)?;
         Ok(unsafe { Txt::from_slice_unchecked(slice) })
@@ -1654,9 +1653,8 @@ impl Txt<[u8]> {
     ///
     /// # Safety
     ///
-    /// The passed octets must contain correctly encoded TXT record data or
-    /// else an error is returned. See [`from_octets][Self::from_octets] for
-    /// the required content.
+    /// The passed octets must contain correctly encoded TXT record data.
+    /// See [`from_octets][Self::from_octets] for the required content.
     unsafe fn from_slice_unchecked(slice: &[u8]) -> &Self {
         unsafe { &*(slice as *const [u8] as *const Self) }
     }
@@ -2160,7 +2158,7 @@ impl<'a> Iterator for TxtIter<'a> {
 
 /// Iteratively build TXT record data.
 ///
-/// This type allows building TXT record data by starting with a empty data
+/// This type allows building TXT record data by starting with empty data
 /// and appending either complete character strings or slices of data.
 #[derive(Clone, Debug)]
 pub struct TxtBuilder<Builder> {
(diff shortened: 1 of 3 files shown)


src/rdata/rfc1035.rs
```diff
@@ -1572,27 +1572,54 @@ impl<N: fmt::Display> fmt::Display for Soa<N> {
...
+        builder.finish().map_err(Into::into)
     }
 }
 
 impl<Octs> Txt<Octs> {
     /// Creates new TXT record data from its encoded content.
+    ///
+    /// The `octets` sequence most contain correctly encoded TXT record
+    /// data. That is, it must contain a sequence of at least one character
+    /// strings of at most 255 octets each preceded by a length octet. An
```
ximon18, line 1618: Did you mean "one character string" here rather than "strings" ?

src/rdata/rfc1035.rs
```diff
@@ -1605,15 +1632,19 @@ impl<Octs> Txt<Octs> {
...
         Txt(octets)
     }
 }
 
 impl Txt<[u8]> {
     /// Creates new TXT record data on an octets slice.
+    ///
+    /// The slice must contain correctly encoded TXT record data,
+    /// that is a sequence of encoded character strings. See
+    ///
```
ximon18, line 1647: Should this trailing RustDoc blank line be removed?

ximon18: Actually the comment is incomplete as it ends with "See".

src/rdata/rfc1035.rs
```diff
@@ -1623,14 +1654,18 @@ impl Txt<[u8]> {
     ///
     /// # Safety
     ///
-    /// The passed octets must contain correctly encoded TXT record data,
-    /// that is a sequence of encoded character strings.
+    /// The passed octets must contain correctly encoded TXT record data or
+    /// else an error is returned. See [`from_octets][Self::from_octets] for
+    /// the required content.
```
ximon18, line 1659: How does this fn return an error as stated given that it returns `&Self`?

src/rdata/rfc1035.rs
```diff
@@ -2093,14 +2152,19 @@ impl<'a> Iterator for TxtIter<'a> {
...
-        self.0.next().map(CharStr::into_octets)
+        self.0.next().map(CharStr::as_slice)
     }
 }
 
 //------------ TxtBuilder ---------------------------------------------------
 
+/// Iteratively build TXT record data.
+///
+/// This type allows building TXT record data by starting with a empty data
```
ximon18, line 2163: I think this should say "starting with empty data" or "starting with an empty data".

src/rdata/rfc1035.rs
```diff
@@ -2121,17 +2186,46 @@ impl<Builder: OctetsBuilder + EmptyBuilder> TxtBuilder<Builder> {
...
         Self::new()
     }
 }
 
 impl<Builder: OctetsBuilder + AsRef<[u8]> + AsMut<[u8]>> TxtBuilder<Builder> {
-    fn builder_append_slice(&mut self, slice: &[u8]) -> Result<(), ShortBuf> {
-        self.builder.append_slice(slice).map_err(Into::into)
+    /// Tries appending a slice.
+    ///
+    /// Errors out of either appending the slice would result in exceeding the
```
ximon18, line 2198: I think this should say "Errors out _if_" rather than "of".

src/rdata/rfc1035.rs
```diff
@@ -2121,17 +2186,46 @@ impl<Builder: OctetsBuilder + EmptyBuilder> TxtBuilder<Builder> {
...
+        self.builder.append_slice(slice)?;
+        Ok(())
     }
 
-    pub fn append_slice(&mut self, mut slice: &[u8]) -> Result<(), ShortBuf> {
+    /// Appends a slice to the builder.
+    ///
+    /// The method breaks up the slice into individual octets strings if
+    /// necessary. If a previous call has started a new octets string, it
+    /// fills this one up first before creating a new one. Thus, be only
```
ximon18, line 2214: I'm not sure if this is what you mean but I find that "be only using this method" sounds strange to me.

ximon18 (approved):

PR #272, opened 2024-02-13
Fix TXT record serialization.

### LINKED ISSUE #268: serde_json round-trip of TXT record fails with spaces
Trying to round-trip a Txt record containing spaces with `serde_json` fails:

```rust
use domain::{base::Dname, rdata::ZoneRecordData};

type Data = ZoneRecordData<Vec<u8>, Dname<Vec<u8>>>;

fn main() {
    let a: Data = serde_json::from_str(r#"{"Txt":"test txt record"}"#).unwrap();
    println!("{:?}", a);
    let b = serde_json::to_string(&a).unwrap();
    println!("{}", b);
    let _: Data = serde_json::from_str(&b).unwrap();
}
```

Output:

```
ZoneRecordData::Txt(Txt(test\32txt\32record))
{"Txt":["test\\32txt\\32record"]}
thread 'main' panicked at src/main.rs:10:44:
called `Result::unwrap()` on an `Err` value: Error("invalid type: string \"test\\\\32txt\\\\32record\", expected a borrowed string", line: 1, column: 31)
```

My impression is that the TXT record wire format isn't supposed to do any sort of escaping, so it's strange to see this here.

partim:
This PR fixes TXT record data serialization for human-readable formats. TXT record data is now serialized as a sequence of character strings. It also deserializes from such a sequence. If supported by the format, it alternatively deserializes from a string that is broken up into 255 octet chunks if necessary.

It uses a newly added helper type `base::charstr::DeserializeCharStrSeed` that deserializes character strings by appending them to an octets builder. Since this may be useful for other cases, it has been made public.

This is a breaking change.

This fixes #268.

### DIFF
--- a/src/base/charstr.rs
+++ b/src/base/charstr.rs
@@ -95,8 +95,8 @@ use std::vec::Vec;
 /// # Serde support
 ///
 /// When the `serde` feature is enabled, the type supports serialization and
-/// deserialization. The format differs for human readable and non-human
-/// readable serialization formats.
+/// deserialization. The format differs for human readable and compact
+/// serialization formats.
 ///
 /// For human readable formats, character strings are serialized as a newtype
 /// `CharStr` wrapping a string with the content as an ASCII string.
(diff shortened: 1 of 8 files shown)


src/base/charstr.rs
```diff
@@ -295,6 +295,35 @@ impl CharStr<[u8]> {
             .map(|bytes| unsafe { Self::from_slice_unchecked(bytes) })
             .map_err(Into::into)
     }
+
+    /// Decodes the readable epresentation and appends it to a builder.
```
Philip-NLnetLabs, line 299:
```suggestion
    /// Decodes the readable presentation and appends it to a builder.
```

Philip-NLnetLabs (approved):
Looks good to me

PR #275, opened 2024-02-19
Client Cache

Philip-NLnetLabs:


Cargo.toml
```diff
@@ -27,6 +27,7 @@ bytes          = { version = "1.0", optional = true, default-features = false }
 chrono         = { version = "0.4.6", optional = true, default-features = false }
 futures-util   = { version = "0.3", optional = true }
 heapless       = { version = "0.8", optional = true }
+moka           = { version = "0.12.3", features = ["future"] }
```
ximon18, line 30: Shouldn't this new `moka` dependency be optional and included bythe `unstable-client-transport` feature?

src/net/client/mod.rs
```diff
@@ -1,8 +1,29 @@
...
 //!
 //! This module provides DNS transport protocols that allow sending a DNS
 //! request and receiving the corresponding reply.
 //!
+//! Currently the following transport protocols are supported:
+//! * [dgram] DNS over a datagram protocol, typically UDP.
+//! * [stream] DNS over an octet stream protocol, typically TCP or TLS.
+//!   Only a single connection is supported.
+//!   The transport works as long as the connection continues to exist.
+//! * [multi_stream] This is a layer on top [stream] where new connections
```
ximon18, line 15:
```suggestion
//! * [multi_stream] This is a layer on top of [stream] where new connections
```

src/net/client/mod.rs
```diff
@@ -1,8 +1,29 @@
...
+//!   are established as old connections are closed (or fail).
+//! * [dgram_stream] This is a combination of [dgram] and [multi_stream].
+//!   This is typically needed because a request over UDP can receive
+//!   a truncated response, which should be retried over TCP.
+//! * [redundant] This transport multiplexes requests over a collection of
+//!   transport connections. The [redundant] transport favors the connection
+//!   with the lowest response time. And of the other transports can be added
+//!   as upstream transports.
```
ximon18, line 23: I don't understand what "And of the other transports can be added as upstream transports" means, the English doesn't "parse" for me.

src/net/client/mod.rs
```diff
@@ -1,8 +1,29 @@
...
+//! * [cache] This is a simple message cache that behaves like a transport.
```
ximon18, line 24: I find this confusing, maybe instead say that it is a caching passthrough transport or a caching adapter that can be used to wrap an existing transport, or something?

src/net/client/mod.rs
```diff
@@ -127,6 +141,21 @@
...
+//! The current implementaton has the following limitations:
+//! * The [dgram] transport does not support DNS Cookies (RFC 7873,
+//!   Domain Name System (DNS) Cookies).
+//! * The [multi_stream] transport does not support timeouts or other limits on
+//!   the number of attempts to open a connection. The caller has to
+//!   implement a timeout mechanism.
+//! * The [cache] transport does not support:
+//!   * prefetching. In this context, prefetching means updating a cache entry
+//!     before it expires.
+//!   * RFC 8767 (Serving Stale Data to Improve DNS Resiliency)
```
ximon18, line 155: Are these RFC references in a particular order, and if not should they perhaps be ascending numerically?

ximon18: Also I think it would be good if RFC references are linked to the relevant IETF RFC pages.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+
+//------------ Config ---------------------------------------------------------
+
+/// Configuration of a cache.
+#[derive(Clone, Debug)]
+pub struct Config {
+    /// Maximum number of cache entries.
+    max_cache_entries: u64,
+
+    /// Maximum validity of a normal result.
```
ximon18, line 145: There are more comments on the constants, e.g. `MAX_VALIDITY`, than there are on these fields, but these fields comments appear in the public RustDoc while the comments on the constants do not. This is a bit strange and unhelpful.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+
+    /// Maximum validity of NODATA results.
+    max_nodata_validity: Duration,
+
+    /// Maximum validity of delegations.
+    max_delegation_validity: Duration,
+}
+
+impl Config {
+    /// Creates a new config with default values.
```
ximon18, line 165: What are the defaults? I think it would be good if they show up in the public RustDoc in some way.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+#[derive(Clone, Debug)]
+pub struct Config {
+    /// Maximum number of cache entries.
+    max_cache_entries: u64,
+
+    /// Maximum validity of a normal result.
+    max_validity: Duration,
+
+    /// Cache duration of transport failures.
+    transport_failure_duration: Duration,
```
ximon18, line 149: Is this a maximum, if so why are the others named `max_` but this is not?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+
+    /// Cache durations of misc. errors.
+    misc_error_duration: Duration,
```
ximon18, line 152: What is a "misc." error?
Is this a maximum, if so why are the others named `max_` but this is not?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+/// maximum of one day.
+const MAX_NXDOMAIN_VALIDITY: DefMinMax<Duration> = DefMinMax::new(
+    Duration::from_secs(3600),
+    Duration::from_secs(60),
+    Duration::from_secs(24 * 3600),
+);
+
+/// Limit on the amount of time to cache a NODATA response.
+///
+/// According to RFC 2308 the limit should be one to three hour with a
```
ximon18, line 77:
```suggestion
/// According to RFC 2308 the limit should be one to three hours with a
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+/// According to RFC 9520 at least 1 second and at most 5 minutes.
+const MISC_ERROR_DURATION: DefMinMax<Duration> = DefMinMax::new(
+    Duration::from_secs(30),
+    Duration::from_secs(1),
+    Duration::from_secs(5 * 60),
+);
+
+/// Limit on the amount of time to cache a NXDOMAIN error.
+///
+/// According to RFC 2308 the limit should be one to three hour with a
```
ximon18, line 67:
```suggestion
/// According to RFC 2308 the limit should be one to three hours with a
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+const TRANSPORT_FAILURE_DURATION: DefMinMax<Duration> = DefMinMax::new(
+    Duration::from_secs(30),
+    Duration::from_secs(1),
+    Duration::from_secs(5 * 60),
+);
+
+/// Limit on the amount of time to cache DNS result codes that are not
+/// NOERROR or NXDOMAIN.
+///
+/// According to RFC 9520 at least 1 second and at most 5 minutes.
```
ximon18, line 58: Linkify this RFC reference?

src/net/client/cache.rs
ximon18: Also, linkify this RFC reference?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+    Duration::from_secs(60),
+    Duration::from_secs(1_000_000_000),
+);
+
+// The following four flags are relevant to caching: AD, CD, DO, and RD.
+//
+// The AD flag needs to be part of the key when DO is clear. When replying,
+// if both AD and DO are not set in the original request then AD needs to be
+// cleared if it was set in the response (extra look up if no entry with
+// AD clear exists.
```
ximon18, line 97:
```suggestion
// AD clear exists).
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+//
+// The CD flag partitions the cache, responses to request with CD set must not
+// be visible to requests with CD and vice versa.
+//
+// A request with DO set can only be satisfied with a response to a request
+// with CD set. However, if CD in the request is clear then a reponse to a
+// request with CD set can be used if all unrequested DNSSEC records are
+// stripped.
```
ximon18, line 105: I don't understand this. You use request twice in the first sentence which confuses me. DO means DNSSEC OK, CD means Checking Disabled, so why would a request with DO set only be satisfiable by a "response to a request with CD set" - errm, what??

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+// The AD flag needs to be part of the key when DO is clear. When replying,
+// if both AD and DO are not set in the original request then AD needs to be
+// cleared if it was set in the response (extra look up if no entry with
+// AD clear exists.
+//
+// The CD flag partitions the cache, responses to request with CD set must not
+// be visible to requests with CD and vice versa.
+//
+// A request with DO set can only be satisfied with a response to a request
+// with CD set. However, if CD in the request is clear then a reponse to a
```
ximon18, line 103:
```suggestion
// with CD set. However, if CD in the request is clear then a response to a
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+// request with CD set can be used if all unrequested DNSSEC records are
+// stripped.
+//
+// A request with RD clear can be satisfied by a response to a request with
+// RD set. For simplicitly requests with RD set can only get a response to
+// a request with RD set. In theory some responses to requests with RD clear
```
ximon18, line 109: Again I find this use of request twice in the same sentence confusing, a response to a request to a response - eh?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+// could satisfy requests with RD set.
+
+// Negative caching is described in RFC 2308.
```
ximon18, line 112: Please link to RFC 2308.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+// NXDOMAIN and NODATA require special treatment. NXDOMAIN can be found
+// directly in the rcode field. NODATA is the condition where the answer
+// section does not contain any record that matches qtype and the message
+// is not a referral. NODATA is distinguished for a referral by the presence
```
ximon18, line 116:
```suggestion
// is not a referral. NODATA is distinguished from a referral by the presence
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+// of a SOA record in the authority section (a SOA record present implies
+// NODATA). A referral has one or more NS records in the authority section.
+// A NXDOMAIN reponse can only be cache if a SOA record is present in the
```
ximon18, line 119:
```suggestion
// An NXDOMAIN reponse can only be cached if a SOA record is present in the
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+// authority section. If the SOA record is absent then the NXDOMAIN response
+// should not be cached.
+// The TTL of the SOA record should reflect how long the response can be
+// cached. So no special treat is needed. Except that a different value should
```
ximon18, line 123:
```suggestion
// cached. So no special treatment is needed. Except that a different value should
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+// limit the maximum time a negative response can be cached.
+//
+// Caching unreachable upstream should be limited to 5 minutes.
+// Caching SERVFAIL should be limited to 5 minutes.
+
+// RFC 8020 suggests a separate <QNAME, QCLASS> cache for NXDOMAIN, but
```
ximon18, line 129: Please link to RFC 8020.

src/utils/config.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub const fn new(def: T, min: T, max: T) -> Self {
+        Self { def, min, max }
+    }
+
+    /// Returns the default value.
+    pub fn default(self) -> T {
+        self.def
+    }
+
+    /// Trims the given value to fit into the minimum/maximum range.
```
ximon18, line 29:
```suggestion
    /// Trims the given value to fit into the minimum/maximum range, inclusive.
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+        self.max_nxdomain_validity = MAX_NXDOMAIN_VALIDITY.limit(value)
+    }
+
+    /// Set the maximum time to cache NODATA results.
+    pub fn set_max_nodata_validity(&mut self, value: Duration) {
+        self.max_nodata_validity = MAX_NODATA_VALIDITY.limit(value)
+    }
+
+    /// Set the maximum time to cache delegations.
+    pub fn set_max_deletation_validity(&mut self, value: Duration) {
```
ximon18, line 202:
```suggestion
    pub fn set_max_delegation_validity(&mut self, value: Duration) {
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+
+impl Config {
+    /// Creates a new config with default values.
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    /// Set the maximum number of cache entries.
+    pub fn set_max_cache_entries(&mut self, value: u64) {
+        self.max_cache_entries = MAX_CACHE_ENTRIES.limit(value)
```
ximon18, line 172: Is there an eviction policy if the maximum is reached, e.g. is the oldest removed or is the entry that doesn't fit dropped or is a warning logged or something else? This should perhaps be documented on the `cache/mod.rs` page indicating that it is not just a cache but what kind of caching policy it implements.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+            max_nodata_validity: MAX_NODATA_VALIDITY.default(),
+            max_delegation_validity: MAX_DELEGATION_VALIDITY.default(),
+        }
+    }
+}
+
+//------------ Connection -----------------------------------------------------
+
+#[derive(Clone)]
+/// A connection that cache response from an upstream connection.
```
ximon18, line 224:
```suggestion
/// A connection that caches responses from an upstream connection.
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+            cache: Cache::new(config.max_cache_entries),
+            config,
+            clock,
+        }
+    }
+}
+
+impl<Upstream> Debug for Connection<Upstream> {
+    fn fmt(&self, _: &mut Formatter<'_>) -> Result<(), core::fmt::Error> {
+        todo!()
```
ximon18, line 282: I suggest that you either implement this or remove the trait impl.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+                        self.state =
+                            RequestState::GetResponseNoCache(request);
+                        continue;
+                    }
+                    let qname = question.qname();
+                    let qclass = question.qclass();
+                    let qtype = question.qtype();
+
+                    if !(opcode == Opcode::Query && qclass == Class::In) {
+                        // Anything other than a query on the Internet class
```
ximon18, line 398: Is this limitation worth mentioning in the `cache.rs` module level docs?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+                    let mut ad = header.ad();
+                    let cd = header.cd();
+                    let rd = header.rd();
+
+                    let dnssec_ok = if let Some(opt) = msg.opt() {
+                        opt.dnssec_ok()
+                    } else {
+                        false
+                    };
+                    if dnssec_ok && !ad {
```
ximon18, line 417:
```suggestion
                    let dnssec_ok = msg.opt().map_or(false, |opt| opt.dnssec_ok());
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+            cache,
+            config,
+            clock,
+        }
+    }
+
+    /// This is the implementation of the get_response method.
+    // This function is cancel safe.
+    async fn get_response_impl(&mut self) -> Result<Message<Bytes>, Error> {
+        loop {
```
ximon18, line 363: Why does this loop?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+                    return response;
+                }
+                RequestState::GetResponseNoCache(request) => {
+                    return request.get_response().await;
+                }
+            }
+        }
+    }
+
+    /// Try to find an cache entry for the key.
```
ximon18, line 457:
```suggestion
    /// Try to find a cache entry for the key.
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+    async fn cache_lookup(
+        &self,
+        key: &Key,
+    ) -> Result<Option<Arc<Value<C>>>, Error> {
+        // There are 4 flags that may affect the response to a query.
+        // In some cases the response to one value of a flag could be
+        // used for the ohter value.
```
ximon18, line 464:
```suggestion
        // used for the other value.
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+        // This function takes all 4 flags. First we fix the CD flag.
```
ximon18, line 465: I'm confused. This function doesn't "fix" anything and "fixing" the CD flag and using it "as is" sound mutually exclusive?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+        if key.dnssec_ok {
+            assert!(key.ad);
+        }
+
+        let opt_value = self.cache_lookup_ad(key).await?;
+        if opt_value.is_some() || key.dnssec_ok {
+            return Ok(opt_value);
+        }
+
+        if key.qclass == Class::In && is_dnssec(key.qtype) {
```
ximon18, line 536: In `get_response_impl()` you check for `Class::In` and bypass caching if not `In`, so why do you check the class here, and what does it mean if `key.qclass != Class::In` and this function continues on?

src/net/client/clock.rs
```diff
@@ -0,0 +1,127 @@
+//! A time interface that can be replaced by a fake time implementation
```
ximon18, line 1: This is nice, but not needed.

If `Instant` in `net/client/cache.rs` is replaced by `tokio::time::Instant` and in `do_client()` when `TIME_PASSES` one calls `tokio::time::advance()` instead of `clock.adjust_time()` then time is normal unless one annotates a test with `#[tokio::test(start_paused = true)]` instead of `#[tokio::test]`. Then only the call to `advance()` causes time to advance. If needed one can also start with time running and only "pause" it (so that it can be advanced) at certain points.

tests/net-client-cache.rs
```diff
@@ -0,0 +1,344 @@
...
+use domain::net::client::clock::{Clock, FakeClock};
+use domain::net::client::multi_stream;
+use domain::net::client::redundant;
+use domain::net::client::request::{
+    Error::NoTransportAvailable, RequestMessage, SendRequest,
+};
+use std::fs::File;
+use std::sync::Arc;
+
+const TEST_FILE_AD: &str = "test-data/client-cache/cache_ad.rpl";
```
ximon18, line 19: One can avoid the various `const FILE_NAME: &str = "file_path"` and N test fns that all invoke `async_test_cache()` by using a test parameterizer, e.g. like is done [here](https://github.com/NLnetLabs/domain/blob/5303cb215e3dabe6c4243de2b7dc672a43fdd260/tests/net-server.rs#L48).

src/base/opt/mod.rs
```diff
@@ -565,6 +565,14 @@ impl<Octs> OptRecord<Octs> {
         self.flags & 0x8000 != 0
```
ximon18, line 565: It might be advisable to extract this repeated use of 0x8000 to a constant to avoid an accidental change later to one occurence but not the others.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+                    let msg = self.request_msg.to_message()?;
+                    let header = msg.header();
+                    let opcode = header.opcode();
+
+                    // Extract Qname, Qclass, Qtype
+                    let mut question_section = msg.question();
+                    let question = match question_section.next() {
+                        None => {
+                            // No question. Just forward the request.
+                            let request = self
```
ximon18, line 375: If I were writing this I would factor out the common code in all the `GetResponseNoCache` then `continue` blocks, but, even if you agree, I don't insist on it being done right now.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+        if opt_value.is_some() || key.rd {
+            return Ok(opt_value);
+        }
+
+        // Look if there is something with RD=1. We can use the
+        // response unmodified.
+        let mut alt_key = key.clone();
+        alt_key.rd = true;
+        let opt_value = self.cache_lookup_do_ad(&alt_key).await?;
+        if let Some(value) = &opt_value {
```
ximon18, line 490: This `if` ... `new_from_value_and_response()` ... `self.cache_insert()` pattern is repeated multiple times below with only the `clear_rd()` part changing as I can see, so if I were writing this I would factor out these common parts, but if possible but not convenient right now that can be left till later.

src/net/client/mod.rs
Philip-NLnetLabs: I came up with "This is a simple message cache provided as a pass through transport." Is that more clear?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+                    // Just insert the error
+                    self.cache_insert(key.clone(), value.clone()).await;
+                }
+                Ok(msg) => {
+                    let msg = clear_rd(msg)?;
+                    let value =
+                        Arc::new(Value::<C>::new_from_value_and_response(
+                            value.clone(),
+                            Ok(msg),
+                            &self.config,
```
ximon18, line 502: I'm a bit concerned about the `clear_xxx()` functions as they don't actually clear anything in the passed value, they clone it, clear the flag in the clone, and return the clone, and this is not obvious from the function name. I would find it more like canonical Rust if the conversions were explicit and instead of adding one `clear_xxx()` fn per flag being cleared just use the existing flag setting functions, e.g.:

```
let mut msg: Message::<BytesMut> = msg.into();
msg.header_mut().set_rd(false);
let value =
    Arc::new(Value::<C>::new_from_value_and_response(
        value.clone(),
        Ok(msg.into()),
        &self.config,
    )?);
```

With these defined somewhere:

```
impl From<&Message<Bytes>> for Message<BytesMut> {
    fn from(msg: &Message<Bytes>) -> Self {
        // We started from a Message which was already checked so this shouldn't fail.
        Message::from_octets(msg.as_slice().into()).unwrap()
    }
}

impl From<Message<BytesMut>> for Message<Bytes> {
    fn from(msg: Message<BytesMut>) -> Self {
        // We started from a Message which was already checked so this shouldn't fail.
        Message::from_octets(msg.into_octets().freeze()).unwrap()
    }
}
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+        None => {
+            // Convert Rcode to OptRcode, this should be part of
+            // OptRcode
+            OptRcode::from_int(msg.header().rcode().to_int() as u16)
+        }
+    }
+}
+
+/// Return a message with the AA flag clear.
+fn clear_aa(msg: &Message<Bytes>) -> Result<Message<Bytes>, Error> {
```
ximon18, line 1103: See above, I would get rid of these `clear_xxx()` functions.

src/net/client/mod.rs
Philip-NLnetLabs: They are sort of sorted in order of importance. I'll add the links for the RFCs

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+            assert!(key.ad);
+        }
+
+        let opt_value = self.cache_lookup_ad(key).await?;
+        if opt_value.is_some() || key.dnssec_ok {
+            return Ok(opt_value);
+        }
+
+        if key.qclass == Class::In && is_dnssec(key.qtype) {
+            // An explicit request for one of the DNSSEC type but
```
ximon18, line 537:
```suggestion
            // An explicit request for one of the DNSSEC types but
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+        }
+        Ok(opt_value)
+    }
+
+    /// Insert new entry in the cache.
+    ///
+    /// Do not insert if the validity is zero.
+    /// Make sure to clear the AA flag.
+    async fn cache_insert(&self, key: Key, value: Arc<Value<C>>) {
+        if value.validity.is_zero() {
```
ximon18, line 617: Should this be even more aggressive and check for greater than some small threshold duration?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+            // Do not insert cache value that are valid for zero duration.
+            return;
+        }
+        let value = match prepare_for_insert(value.clone(), &self.config) {
+            Ok(value) => value,
+            Err(e) => {
+                // Create a new value based on this error
+                Arc::new(
+                    Value::<C>::new_from_value_and_response(
+                        value.clone(),
```
ximon18, line 627:
```suggestion
                        value,
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+    // optimize for the case where RD is clear.
+    let mut msg = Message::<Vec<u8>>::from_octets(msg.as_slice().into())?;
+    let hdr = msg.header_mut();
+    hdr.set_rd(false);
+    Ok(Message::<Bytes>::from_octets(msg.into_octets().into())?)
+}
+
+/// Prepare a value for inserting in the cache by clearing the AA flag if
+/// set.
+fn prepare_for_insert<C>(
```
ximon18, line 1134: This is very similar to repeated code blocks above, only this does `clear_aa()` instead of `clear_rd()` etc. I suggest to refactor these almost common code blocks out (and as said above to use `set_aa(false)` rather than `clear_aa()`).

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+}
+
+impl<CR, Upstream, C> Debug for Request<CR, Upstream, C>
+where
+    CR: Send + Sync,
+    Upstream: Send + Sync,
+    C: Clock + Send + Sync,
+{
+    fn fmt(&self, _: &mut Formatter<'_>) -> Result<(), core::fmt::Error> {
+        todo!()
```
ximon18, line 646: I suggest that you either implement this or remove the trait impl.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+        Ok(opt_value)
+    }
+
+    /// Insert new entry in the cache.
+    ///
+    /// Do not insert if the validity is zero.
+    /// Make sure to clear the AA flag.
+    async fn cache_insert(&self, key: Key, value: Arc<Value<C>>) {
+        if value.validity.is_zero() {
+            // Do not insert cache value that are valid for zero duration.
```
ximon18, line 618: This comment is redundant as you already say this in the fn RustDoc AND the code is quite self-evident.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+//------------ Value ----------------------------------------------------------
+
+#[derive(Debug)]
+/// The value to be cached.
+struct Value<C>
+where
+    C: Clock + Send + Sync,
+{
+    /// Creation time of the cache entry.
+    creation: C::Instant,
```
ximon18, line 750: I have a slight preference for `created_at` and `valid_for` instead of `creation` and `validity`.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+    response: Result<Message<Bytes>, Error>,
+}
+
+impl<C> Value<C>
+where
+    C: Clock + Send + Sync,
+{
+    /// Create a new value object.
+    fn new(
+        response: &Result<Message<Bytes>, Error>,
```
ximon18, line 765: I have a slight preference for passing in the cloned value rather than hiding the clone operation within this fn.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+
+/// Compute how long a response can be cached.
+fn validity(
+    response: &Result<Message<Bytes>, Error>,
+    config: &Config,
+) -> Result<Duration, Error> {
+    let msg = match response {
+        Err(_) => return Ok(config.transport_failure_duration),
+        Ok(msg) => msg,
+    };
```
ximon18, line 819:
```suggestion
    let Ok(msg) = response else {
        return Ok(config.transport_failure_duration);
    };
```

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+fn get_opt_rcode<Octs: Octets>(msg: &Message<Octs>) -> OptRcode {
+    let opt = msg.opt();
+    match opt {
+        Some(opt) => opt.rcode(msg.header()),
+        None => {
+            // Convert Rcode to OptRcode, this should be part of
+            // OptRcode
+            OptRcode::from_int(msg.header().rcode().to_int() as u16)
+        }
+    }
```
ximon18, line 1099: `Rcode` does support conversion to `OptRcode`.

You could alternately do one of the following:

```
    msg.opt().map_or_else(
        || msg.header().rcode().into(),
        |opt| opt.rcode(msg.header()),
    )
```

Or:

```
    msg.opt()
        .map(|opt| opt.rcode(msg.header()))
        .unwrap_or_else(|| msg.header().rcode().into())
```

src/net/client/cache.rs
Philip-NLnetLabs: The ones with max_* are also limited by the TTLs of the resource records. The transport_failure_duration is cached for exactly that amount of time (unless the entry is expired from the cache before that time).

src/net/client/protocol.rs
```diff
@@ -98,23 +98,21 @@ impl<Addr> TlsConnect<Addr> {
     }
 }
 
-impl<Addr> AsyncConnect for TlsConnect<Addr>
-where
-    Addr: ToSocketAddrs + Clone + Send + 'static,
-{
+impl AsyncConnect for TlsConnect {
```
ximon18, line 101: This change prevents TcpStream::connect() from attempting to connect in turn to each of multiple addresses until one is successful. Is that intentional?

src/net/client/request.rs
```diff
@@ -33,18 +33,21 @@ pub trait ComposeRequest: Debug + Send + Sync {
...
+    fn to_vec(&self) -> Result<Vec<u8>, Error>;
 
     /// Return a reference to a mutable Header to record changes to the header.
     fn header_mut(&mut self) -> &mut Header;
 
     /// Set the UDP payload size.
     fn set_udp_payload_size(&mut self, value: u16);
 
+    /// Set the DNSSEC OK flag.
+    fn set_dnssec_ok(&mut self, value: bool);
```
ximon18, line 49: Should this maybe be more generic, e.g. `fn set_edns_flag(FlagEnum)` ?

src/net/client/stream.rs
```diff
@@ -852,23 +862,25 @@ mod test {
         let mut queries = Queries::new();
 
         for i in 0..12 {
-            let (idx, item) = queries.insert(i).unwrap();
+            let (idx, item) = queries.insert(i).expect("test failed");
```
ximon18, line 865: Are these `expect()` calls really any more useful than the `unwrap()` calls? In both cases you will have to look at the code to find out where it failed as the expect messages are not distinct thus don't tell you where it failed, and the `unwrap()` calls make the code shorter.

src/utils/config.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// The minimum value,
+    min: T,
+
+    /// The maximum value,
+    max: T,
+}
+
+impl<T> DefMinMax<T> {
+    /// Creates a new value.
+    pub const fn new(def: T, min: T, max: T) -> Self {
```
ximon18, line 20: I have a slignt preference for min, def, max instead of def, min, max because I think of def being a value in between min and max.

test-data/client-cache/cache_aa.rpl
```diff
@@ -0,0 +1,198 @@
+; Test if caching clears the AA flag. We issue the same query twice.
```
ximon18, line 1: If these rpl files are taken as-is from Unbound it might be good to a comment referencing the origin.
If these files have been modified compared to the original that would also be useful to note.

test-data/basic.rpl
```diff
@@ -131,7 +131,7 @@ RANGE_BEGIN 0 100
 ENTRY_BEGIN
 MATCH opcode qtype qname
 ADJUST copy_id copy_query
-REPLY QR RD NOERROR
+REPLY QR RD RA NOERROR
```
ximon18, line 134: Was the original Unbound .rpl file wrong here?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+
+                    let mut ad = header.ad();
+                    let cd = header.cd();
+                    let rd = header.rd();
+
+                    let dnssec_ok = if let Some(opt) = msg.opt() {
+                        opt.dnssec_ok()
+                    } else {
+                        false
+                    };
```
ximon18, line 416:
```suggestion
                    let dnssec_ok = msg.opt().map_or(false, |opt| opt.dnssec_ok());
```

tests/net/deckard/client.rs
```diff
@@ -26,11 +32,19 @@ pub async fn do_client<R: SendRequest<RequestMessage<Vec<u8>>>>(
...
+                        "Reply message does not match at step {}",
+                        step_value.get()
+                    );
                     panic!("reply failed");
                 }
             }
-            StepType::TimePasses
-            | StepType::Traffic
+            StepType::TimePasses => {
+                clock.adjust_time(Duration::from_secs(
```
ximon18, line 43: See earlier comments on this PR about using `tokio::time::advance()` here instead.

tests/net/deckard/client.rs
```diff
@@ -67,8 +81,21 @@ fn entry2reqmsg(entry: &Entry) -> RequestMessage<Vec<u8>> {
     if reply.rd {
         msg.header_mut().set_rd(true);
     }
+    if reply.ad {
```
ximon18, line 84: Maybe this whole block could be simplified by replacing the pattern:

```
if reply.ad {
    msg.header_mut().set_ad(true);
}
```

With:

```
let mut header = msg.header_mut();
header.set_rd(reply.rd);
header.set_ad(reply.ad);
header.set_cd(reply.cd);
```

And so on...

tests/net/deckard/matches.rs
```diff
@@ -83,16 +85,47 @@ where
...
+        if verbose {
+            println!("match_msg: CD not in message",);
         }
         return false;
     }
     if matches.fl_do {
-        todo!();
+        if let Some(opt) = msg.opt() {
+            if !opt.dnssec_ok() {
+                if verbose {
```
ximon18, line 112: Maybe it's worth writing a tiny macro called `verbose_println!()` ?

tests/net/deckard/server.rs
```diff
@@ -75,8 +77,13 @@ fn do_adjust<Octs: Octets>(
         msg.push(rec).unwrap();
     }
     let mut msg = msg.additional();
-    for _a in &sections.additional {
-        todo!();
+    for a in &sections.additional {
+        let rec = if let ZonefileEntry::Record(record) = a {
```
ximon18, line 81: This can be simplified using a let-else pattern, e.g. `let ZonefileEntry::Record(rec) = a else { panic!() };`.

tests/net/deckard/server.rs
```diff
@@ -86,34 +93,37 @@ fn do_adjust<Octs: Octets>(
         msg.header_mut().set_aa(true);
     }
     if reply.ad {
-        todo!()
+        msg.header_mut().set_ad(true);
```
ximon18, line 96: Same suggestion as above, call `header_mut()` once and then skip the if checks and just pass the `reply.bool` value to the appropriate setter.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+);
+
+/// Limit on the amount of time a delegation is considered valid.
+const MAX_DELEGATION_VALIDITY: DefMinMax<Duration> = DefMinMax::new(
+    Duration::from_secs(1_000_000),
+    Duration::from_secs(60),
+    Duration::from_secs(1_000_000_000),
+);
+
+// The following four flags are relevant to caching: AD, CD, DO, and RD.
```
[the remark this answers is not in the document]
Philip-NLnetLabs: Can we assume knowledge of DNS RFCs or do we explain everything in every piece of code that we write?

src/net/client/cache.rs
Philip-NLnetLabs: I don't know how I typed that wrong. What I meant is:
// A request with DO set can only be satisfied with a response to a request
// with DO set. However, if DO in the request is clear then a response to a
// request with DO set can be used if all unrequested DNSSEC records are
// stripped.

src/net/client/cache.rs
Philip-NLnetLabs: I hope the following is more clear:
// A request with RD clear can be satisfied by a response to a request with
// RD set. For simplicitly requests with RD set will only get a cached
// response to another request with RD set. In theory some responses to
// requests with RD clear could be used to satisfy requests with RD set.
// However, this is not implemented.

src/utils/config.rs
```diff
@@ -0,0 +1,36 @@
...
+        Self { def, min, max }
+    }
+
+    /// Returns the default value.
+    pub fn default(self) -> T {
```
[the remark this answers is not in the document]
Philip-NLnetLabs: No, there is no way to construct a DefMinMax object with a default value. But a DefMinMax object does contain a default value for a configuation option.

src/net/client/cache.rs
Philip-NLnetLabs: From the Moka documentation:

Maintains near optimal hit ratio by using an entry replacement algorithms inspired by Caffeine:

    Admission to a cache is controlled by the Least Frequently Used (LFU) policy.
    Eviction from a cache is controlled by the Least Recently Used (LRU) policy.
    [More details and some benchmark results are available here](https://github.com/moka-rs/moka/wiki#admission-and-eviction-policies).

Supports expiration policies:

    Time to live.
    Time to idle.
    Per-entry variable expiration.

Do we care at the moment?

src/net/client/cache.rs
Philip-NLnetLabs: Otherwise the code gets too complex, in particular

self.state =
                            RequestState::GetResponseNoCache(request);
                        continue;

src/net/client/cache.rs
Philip-NLnetLabs: I rewrote the comment as:        
        // This function takes all 4 flags. First we take care of the CD flag.
        // This flag has to be used as is, so there is not much to do. Next
        // we pass the request to a function that looks at RD, DO, and AD.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+        let opt_value = self.cache_lookup_do_ad(key).await?;
+        if opt_value.is_some() || key.rd {
+            return Ok(opt_value);
+        }
+
+        // Look if there is something with RD=1. We can use the
+        // response unmodified.
+        let mut alt_key = key.clone();
+        alt_key.rd = true;
+        let opt_value = self.cache_lookup_do_ad(&alt_key).await?;
```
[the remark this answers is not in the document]
Philip-NLnetLabs: I changed the comment for cache_lookup_do_ad to:
   /// Try to find an cache entry for the key taking into account the
    /// DO and AD flags. The CD and RD flags are kept unchanged.

Is that more clear?

src/net/client/cache.rs
Philip-NLnetLabs: I'll remove the qclass check in this case. It doesn't bring anything even if the function would be called with a different class.

src/base/opt/mod.rs
Philip-NLnetLabs: Can you take this up with Martin? It is his choice to not have a constant for the DO bit.

src/net/client/cache.rs
Philip-NLnetLabs: I don't see how your version is better. Mine is a clear function, it take an input message and returns a modified one. If there is a better way to name those functions then of course the names can be changed.

Your version seems to move a lot of stuff to the caller and hides the clones in the '.into' calls making it less clear what is going on.

src/net/client/cache.rs
Philip-NLnetLabs: No. is_zero is used as a signal that something should not be cached. That is consistent with the use of TTL value 0 in DNS. There is nothing wrong with caching something for a small period of time. Unless later that turns out to be wrong. But that seems unlikely because TTL is in seconds, and 1 second is an eternity on modern computers.

src/net/client/request.rs
Philip-NLnetLabs: As far is I know, that doesn't exist in OptBuilder. As far as I know the only flag in existence is DO.

src/net/client/stream.rs
Philip-NLnetLabs: The main point is that it is easy to verify that there are no unwraps in production code. If there is another way to grep for unwrap that excludes unit tests, then let me know.

test-data/client-cache/cache_aa.rpl
Philip-NLnetLabs: The tests are new. Some of the boiler plate for a recursor has been stolen. I'm not sure if it makes sense to remove the boiler plate or not

test-data/basic.rpl
Philip-NLnetLabs: This was already a custom test and not meant to be an existing test

tests/net/deckard/server.rs
Philip-NLnetLabs: if let ZonefileEntry::Record(rec) = a {} else {
            panic!("include not expected")
        };

gives:
error[E0425]: cannot find value `rec` in this scope
  --> tests/net/deckard/server.rs:84:18
   |
84 |         msg.push(rec).unwrap();
   |                  ^^^ not found in this scope

src/net/client/cache.rs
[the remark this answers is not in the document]
ximon18: I world err on the side of providing more such as references to justify and explain what is being done, but indeed it is a judgement call.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1177 @@
...
+        // For DO=0, first try with DO=0 and then try with DO=1. If
+        // DO=1 has an answer, remove DNSSEC related resource records.
+        // If AD is clear then clear the AD bit.
+
+        // If DO is set then AD is irrelevant. Force AD to be set for
+        // consistency (if DO is set then with respect to the AD flag
+        // the behavior is as if AD is set).
+
+        if key.dnssec_ok {
+            assert!(key.ad);
```
partim, line 550: This feels a bit scary. Maybe have an enum that reflects the valid combinations of DO and AD bits? Or just force the AD bit to false here?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1177 @@
...
+                }
+                Ok(msg) => {
+                    if !msg.header().ad() {
+                        // AD is clear. Just insert this value.
+                        self.cache_insert(key.clone(), value.clone()).await;
+                        return Ok(Some(value.clone()));
+                    }
+
+                    let msg = clear_ad(msg)?;
+                    let value =
```
partim, line 620: You are doing this dance of creating a value and then inserting it quite a lot. Maybe it’s worth having a separate method for it?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1177 @@
...
+    GetResponseNoCache(Box<dyn GetResponse + Send + Sync>),
+}
+
+//------------ Key ------------------------------------------------------------
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+/// The key for cache entries.
+struct Key {
+    /// DNS name in the request.
+    qname: Dname<Vec<u8>>,
```
partim, line 713: Given that you are cloning the key quite a lot, it might be beneficial to use a `Dname<Bytes>` here since it is a lot cheaper to clone?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1177 @@
...
+        qtype: Rtype,
+        ad: bool,
+        cd: bool,
+        dnssec_ok: bool,
+        rd: bool,
+    ) -> Key
+    where
+        TDN: ToDname,
+    {
+        let mut qname = qname.to_dname().expect("to_dname should not fail");
```
partim, line 748: We should probably switch `ToDname` to have a pair of `try_to_dname` and `to_dname` methods to avoid these kinds of “unwraps.” Most other places already have this.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1177 @@
...
+            response,
+        })
+    }
+}
+
+//------------ Utility functions ----------------------------------------------
+
+/// Handle a cached value. Either return None if the value has expired or
+/// return a response message with decremented TTL values.
+fn handle_cache_value<TDN, C>(
```
partim, line 818: Could this be a method on `Value`?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1177 @@
...
+    let msg =
+        Message::<Bytes>::from_octets(result.finish().into_target().into())
+            .expect(
+                "Message should be able to parse output from MessageBuilder",
+            );
+    Ok(msg)
+}
+
+/// Return a new message without the DNSSEC type RRSIG, NSEC, and NSEC3.
+fn remove_dnssec(
```
partim, line 975: Could this be a convenience method on `Message`?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1160 @@
...
+            max_delegation_validity: MAX_DELEGATION_VALIDITY.default(),
+        }
+    }
+}
+
+//------------ Connection -----------------------------------------------------
+
+#[derive(Clone)]
+/// A connection that cache response from an upstream connection.
+pub struct Connection<Upstream, C: Clock + Send + Sync = SystemClock> {
```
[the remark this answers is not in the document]
Philip-NLnetLabs: The Send and Sync bounds come from Moka. The clock gets inserted in the cache and the cache is multi-treaded

src/net/client/mod.rs
```diff
@@ -127,6 +142,25 @@
 //! # }
 //! ```
 
+//! # Limitations
```
partim, line 145: Do we want to create issues for these and then link to them here?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1177 @@
...
+use std::fmt::{Debug, Formatter};
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::time::Duration;
+use std::vec::Vec;
+
+/// Configuration limit for the maximum number of entries in the cache.
+const MAX_CACHE_ENTRIES: DefMinMax<u64> =
+    DefMinMax::new(1_000, 1, 1_000_000_000);
```
ximon18, line 34: Given that the max of u64 is way higher than 1,000,000,000, where does this limit value come from?

src/net/client/cache.rs
Philip-NLnetLabs: I'll leave this in cache at least until it is clear what the validator needs. There are three problems: the name. It removes only some DNSSEC records. It can clear the AD flag, not clear if that is needed in general (or that it should do more, may something for plain), It removes records even if they are requested by the question.

src/net/client/cache.rs
ximon18: This still doesn't work for me. There's too much going on with the terms "request" and "response" in a single sentence. Is it trying to say "Responses to DNSSEC-aware clients (advertised by setting the DO bit in the request) are stored in a dedicated cache. Only responses from this cache (and not from any other cache) may be served to DNSSEC-aware clients" ?

src/utils/config.rs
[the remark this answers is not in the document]
ximon18: Ah good point.

src/net/client/cache.rs
ximon18: I was thinking that users want to understand what the code they are using is doing at a high level without having to try and read the code, so I would say "yes" we care at the moment, but that's just my opinion.

src/net/client/cache.rs
ximon18: I would find it very useful to have that text in the comment rather than "More than one question. Just forward the request." as it is now, as the current text leaves the reader having to guess why this decision has been made (even if they think they know, they can't know the mind of the developer who wrote the code/comment and so may be wrong).

src/net/client/cache.rs
Philip-NLnetLabs: Doesn't seem to be an action on this PR.

src/net/client/cache.rs
Philip-NLnetLabs: Bytes does not implement AsRef<[u8]> which is needed for Debug.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1177 @@
...
+    async fn get_response_impl(&mut self) -> Result<Message<Bytes>, Error> {
+        loop {
+            match &mut self.state {
+                RequestState::Init => {
+                    let msg = self.request_msg.to_message()?;
+                    let header = msg.header();
+                    let opcode = header.opcode();
+
+                    // Extract Qname, Qclass, Qtype
+                    let mut question_section = msg.question();
```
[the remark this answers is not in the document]
Philip-NLnetLabs: I didn't know about sole_question, but now that I do, how to I distinguish between qdcount != 1 and other parse errors?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1177 @@
...
+                            .send_request(self.request_msg.clone());
+                        self.state =
+                            RequestState::GetResponseNoCache(request);
+                        continue;
+                    }
+                    let qname = question.qname();
+                    let qclass = question.qclass();
+                    let qtype = question.qtype();
+
+                    if !(opcode == Opcode::Query && qclass == Class::In) {
```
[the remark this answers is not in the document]
Philip-NLnetLabs: I'm confused, which test?

src/net/client/cache.rs
partim: It certainly does: https://docs.rs/bytes/latest/bytes/struct.Bytes.html#impl-AsRef%3C%5Bu8%5D%3E-for-Bytes

src/net/client/cache.rs
partim: I created #281

src/net/client/cache.rs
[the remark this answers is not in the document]
Philip-NLnetLabs: I added references to the RFCs

src/net/client/clock.rs
Philip-NLnetLabs: I'll resolve this because it is now in a separate PR.

src/net/client/cache.rs
Philip-NLnetLabs: I must have done something wrong. Now I get a new error: 

760 |         qname.make_canonical();
    |               ^^^^^^^^^^^^^^ the trait `AsMut<[u8]>` is not implemented for `bytes::Bytes`

I can start out with a Vec<u8> but then how do I convert to Bytes?

src/net/client/cache.rs
Philip-NLnetLabs: Just a random number that seems sensible. It might be too high even. Just have sane defaults to avoid accidents.

src/net/client/mod.rs
Philip-NLnetLabs: Would that be good if we don't have time to address those issues this year?

src/net/client/cache.rs
ximon18: To better illustrate my thinking see https://github.com/NLnetLabs/domain/pull/282.

src/net/client/cache.rs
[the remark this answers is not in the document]
partim: You can’t. But maybe you don’t need to and just forward it in either case? Might even be good, whoever you forward this to might even be able or willing to deal with it.

src/net/client/cache.rs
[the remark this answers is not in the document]
partim: I’m now, too. You could quickly look at the opcode and forward if that is not QUERY even before looking at the question and only then get the question and look at that.

src/net/client/cache.rs
ximon18: Ah, that PR misses that your code might use a state machine not for logic reasons but purely to save state across an .await point. If that's the case a comment to that effect somewhere would be very useful, otherwise it is unclear why the code is structured that way.

partim: Minor idea: If you switch the state to an option, you can switch the loop to a `while let Some` kind of thing and automatically fuse the future.

src/net/client/cache.rs
partim: Just call `.into()` on it.

src/net/client/cache.rs
ximon18: Makes sense, but isn't clear from the code so if you could note that in a comment I would really appreciate that.

src/base/opt/mod.rs
ximon18: As it was already discussed let's just leave it for now. I'm not a fan of repeated use of identical values as in my experience they are a foot gun waiting to happen, but a named constant would be less readable than the hex manipulation as the latter is very common in working with bit flags.

src/net/client/cache.rs
ximon18: Leave it then, I find it harder to read your code, you find it harder to read mine ;-)

src/net/client/stream.rs
ximon18: I don't agree that support for grepping for `unwrap` should be a goal. As it is test code, leave it.

src/utils/config.rs
ximon18: Ah, my bad.

test-data/basic.rpl
ximon18: Ah, my bad.

tests/net/deckard/server.rs
ximon18: It should be:
```rust
        let ZonefileEntry::Record(rec) = a else {
            panic!("include not expected")
        };
```

ximon18 (approved):
Great work @Philip-NLnetLabs! :champagne:

src/net/client/cache.rs
Philip-NLnetLabs: 763 |     let qname: Dname<Bytes> = qname.into();
    |                                     ^^^^ the trait `From<name::dname::Dname<Vec<u8>>>` is not implemented for `name::dname::Dname<bytes::Bytes>`

src/net/client/mod.rs
partim: I don’t know. I mostly want to avoid that we forget about implementing this. But maybe that’ll happen all by itself anyway.

tests/net/deckard/matches.rs
partim: Or use logging for it throughout?

ximon18: PR #274 introduces logging via tokio tracing and indeed does some of that for Deckard/Stelline, so we could maybe revisit this after looking at that?

src/net/client/cache.rs
partim: Ah, sorry. The `OctetsInto` is the version of `Into` that converts the underlying octets, i.e., `qname.octets_into()` with `use octseq::octets::OctetsInto`.

Philip-NLnetLabs: Yes, that worked!

src/net/client/cache.rs
Philip-NLnetLabs: I added a comment to set_max_cache_entries, which should show up in to documentation.

src/net/client/cache.rs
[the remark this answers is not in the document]
Philip-NLnetLabs: I don't see a nice way of doing that. Just checking for opcode early would duplicate the send_request call. It would be nice to skip extracting qname/qclass/qtype if opcode is not Query, but it seems that it would make that code more complex. In addition, anything that is not using opcode == Query would probably not use the cache in the first place.

tests/net/deckard/matches.rs
Philip-NLnetLabs: The function is called in two different ways. One is to find a suitable entry to respond, that one should not log failures. The other is when a reply is expected by the client and failures need to be logged.

Cargo.toml
```diff
@@ -94,4 +95,4 @@ required-features = ["std", "rand"]
 
 [[example]]
 name = "client-transports"
-required-features = ["net"]
\ No newline at end of file
+required-features = ["net"]
```
ximon18, line 98: Shouldn't this include the "unstable-client-transport" feature in the set of `required-features` ?

tests/net-client-cache.rs
```diff
@@ -0,0 +1,344 @@
...
+    msg.push((Dname::vec_from_str("example.com").unwrap(), Aaaa))
+        .unwrap();
+    let req = RequestMessage::new(msg);
+
+    let mut request = cached.send_request(req.clone());
+    let reply = request.get_response().await;
+
+    println!("got {reply:?}");
+
+    if let Err(NoTransportAvailable) = reply {
```
ximon18, line 300: Maybe use `assert!()` here?

src/net/client/cache.rs
```diff
@@ -0,0 +1,1200 @@
...
+    /// The configuration of this connection.
+    config: Config,
+
+    /// The clock to use for expiring cache entries.
+    clock: C,
+}
+
+impl<Upstream> Connection<Upstream> {
+    /// Create a new connection with default configuration parameters.
+    pub fn new(upstream: Upstream) -> Self {
```
ximon18, line 283: In the generated RustDoc the lack of trait bounds on `Upstream` makes it very unclear what the caller should pass in as an argument to `nw()`. One way to make this clearer would be to add a `Upstram: SendRequest<CR>` bound on `new()`, or maybe even better at the `struct` level where it will appear at the top of hte generated docs.

src/net/client/cache.rs
```diff
@@ -0,0 +1,1200 @@
...
+        }
+    }
+}
+
+impl<Upstream, C> Connection<Upstream, C>
+where
+    C: Clock + Send + Sync + 'static,
+{
+    /// Create a new connection with default configuration parameters.
+    pub fn new_with_time(upstream: Upstream, clock: C) -> Self {
```
ximon18, line 303: If kept this should be documented as being intended for internal test purposes only. Adopting https://github.com/NLnetLabs/domain/pull/278 however would remove this issue by removing these `with_time` functions.

ximon18:
One thing not provided by this PR at present is any notion of cache metrics/statistics. I also don't see any obvious support on the underlying [moka](https://docs.rs/moka/latest/moka/) crate for this.so presumably one would have to wrap internal calls in `cache.rs` to track the miss/hit ratio, for example. I don't see how the user of the cache in this crate could do this themselves without forking the code as the `cache.get()` call is deep inside library code, so users of this library would have to wait for us to add such support later or PR it themselves.

src/net/client/cache.rs
partim: Generally, bounds are kept to the minimum necessary set of traits. So this should be in the struct’s docstring.

ximon18: But isn't it a minimum requirement anyway in this case to be able to actually use the struct as intended?

partim: Minimum requirement for the particular block. For the struct definition, you only need `Upstream` to be `Sized` (which is implied).

partim: Given that trait bounds trickle through when types are used, limiting them to what’s needed keeps where clauses short. Similarly for just all the impl blocks for the types themselves.

ximon18: I find the usability more important than the length of trait bounds. Either way this needs to be clear to a user, whether through a RustDoc comment, a helpful compilation error or a less helpful compilation error.

tests/net-client-cache.rs
Philip-NLnetLabs: I'm not sure how.

ximon18: Like so:
```
    assert!(matches!(reply, Err(NoTransportAvailable)), "Bad result {reply:?}");
```

ximon18:
A thought:

Would it be worth making the cache get and insert interface a trait taken by the xache client, and making the current in memory moka cache one impl of it?

Then the same client could be used with an alternate backing cache impl.

This occurred to me because of a similar discussion about zone storage.

WhyNotHugo:
As a downstream consumer of this crate (for [libdav](https://docs.rs/libdav/latest/libdav/) and [pimsync](https://pimsync.whynothugo.nl/)), the introduction of moka as a dependency is quite a burden: it substantially increases the amount of dependencies that I need to review, and total build time.

I'm mostly relying on `domain` for the stub resolver. If I understand correctly, `moka` is used in the context of DNSSEC validation, but the stub resolver doesn't currently implement DNSSEC validation.

My particular usage does a few DNS queries (often times 3-4 at startup for a process that then runs uninterrupted for days or weeks). Something simple and "slower" (e.g.: using `RwLock<HashMap>`) would probably not introduce any significant overhead for my use case. This is likely true for many other consumers.

> Would it be worth making the cache get and insert interface a trait taken by the cache client, and making the current in memory moka cache one impl of it?

I think this would be a pretty apt solution. Downstream consumers can easily create a newtype aronud `moka::cache::Cache` or `RwLock<HashMap>`, or whatever is the best fit.

Philip-NLnetLabs:
The client cache is mostly unrelated to DNSSEC validation. The cache does know about DNSSEC but is useful even without DNSSEC validation. There is no easy way around Moka. Moka has many nice features. Re-implementing them would essentially be a poor man's Moka.

That said, we could look into making the cache optional. The resolver currently does not use the cache.

WhyNotHugo:
I tried making the cache optional by using `C: Cache<Key, Arc<Value>>` for a cache (the exact key and value types varies on each usage). This would allow consumers to provide any implementation; either `moka` or a very minimal lightweight one.

However, the `Key` type used for the cache is a private type in a few cases, so this required making the type public, as well as a few other invasive changes. I'm not convinced that the approach I have taken here is feasible.

> Re-implementing them would essentially be a poor man's Moka.

I would find this very fitting for my use case.

> That said, we could look into making the cache optional. The resolver currently does not use the cache.

That sounds good to me, but I don't think I'm familiar enough to try and execute that myself.

Philip-NLnetLabs:
I created a PR that make the client cache optional. Could you please check if that works for you?

PR #276, opened 2024-02-22
Refactor IANA type variants into constants.

### LINKED ISSUE #273: Change IANA types from an enum into a struct wrapping an integer.
It occurred to me that the way IANA types are represented in `base::iana` is not great. Currently we have an enum with an `Int` variant for undefined values. But because the field of that variant is public, you can actually create an `Int` variant for a well-defined value which is weird at best and causes issues at worst.

Instead, we should probably have a type that (privately) wraps the underlying integer type and define associated constants for all the known variants.

Obviously, this is a massively breaking change but I think it is worth doing as it will make the code more robust. So, I’d like to do it with 0.10.0 already.

partim:
This PR changes the types for the IANA registered values into structs wrapping underlying integer types and provides the values as constants. This mostly just changes the spelling of those values. However, it also required some re-wiring of the type generation macros for record data and opt data since the type names and value names differ now.

This is a breaking change.

Fixes #273.

### DIFF
--- a/src/base/header.rs
+++ b/src/base/header.rs
@@ -286,7 +286,7 @@ impl Header {
     /// [`Rcode`]: ../../iana/rcode/enum.Rcode.html
     #[must_use]
     pub fn rcode(self) -> Rcode {
-        Rcode::from_int(self.inner[3] & 0x0F)
+        Rcode::saturating_from_int(self.inner[3])
     }
 
     /// Sets the value of the RCODE field.
@@ -1010,7 +1010,7 @@ mod test {
     fn header() {
         test_field!(id, set_id, 0, 0x1234);
         test_field!(qr, set_qr, false, true, false);
-        test_field!(opcode, set_opcode, Opcode::Query, Opcode::Notify);
+        test_field!(opcode, set_opcode, Opcode::QUERY, Opcode::NOTIFY);
         test_field!(
             flags,
             set_flags,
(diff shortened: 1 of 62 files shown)


src/base/header.rs
```diff
@@ -286,7 +286,7 @@ impl Header {
     /// [`Rcode`]: ../../iana/rcode/enum.Rcode.html
     #[must_use]
     pub fn rcode(self) -> Rcode {
-        Rcode::from_int(self.inner[3] & 0x0F)
+        Rcode::saturating_from_int(self.inner[3])
```
Philip-NLnetLabs, line 289: Why is this called saturating _from_int. It seems to just mask the lower 4 bits. What is the saturating part?

src/rdata/aaaa.rs
```diff
@@ -106,7 +111,7 @@ impl CanonicalOrd for Aaaa {
 
 impl RecordData for Aaaa {
     fn rtype(&self) -> Rtype {
-        Rtype::Aaaa
+        Aaaa::RTYPE
```
Philip-NLnetLabs, line 114: I find this confusing. It looks like a namespace that is the wrong way around.

src/base/header.rs
partim: The idea was to have a name reminiscent of something like `saturating_sub` which accepts the input and sanitizes it into a valid value which is what this does as well.

src/rdata/aaaa.rs
partim: Would it help if this were `Self::RTYPE`?

Philip-NLnetLabs: Yes, I think that is better

src/base/header.rs
Philip-NLnetLabs: Maybe call it masked_from_int?

partim: That sounds like a good idea. Will do!

Philip-NLnetLabs (approved):

PR #277, opened 2024-02-22
Partially seal error types.

### LINKED ISSUE #141: Seal error types.
Plenty of error types are currently public enums. This leads to the necessity of breaking changes any time their content needs to be adjusted. Instead, we should use private enums inside newtypes.

partim:
This PR looks at all the error types and splits them into meaningful error information, which will be available as enum variants, and internals, which will be hidden away behind opaque structs. For the former case, we will still use enums directly rather than have error kind enums. These will not be non-exhaustive since they should really cover all possible cases and exhaustive matching should work.

This is a breaking change.

Fixes #141


ximon18:
@partim: I get the point of this change, and understand how the hiding works when an enum variant has a public inner value but whose construction is private and thus only possible from within the crate code itself, and that for variants that can't be acted upon but merely need to be communicated handle this via a Display impl on the private type.

What I don't get is why things like `pub struct CharStrError;` have become `pub struct CharStrError(());`- what value does that inner unit type  add?

src/base/name/builder.rs
```diff
@@ -543,51 +538,54 @@ impl std::error::Error for PushNameError {}
...
-pub enum LabelFromStrError {
-    /// The string ended when there should have been more characters.
-    ///
-    /// This most likely happens inside escape sequences and quoting.
-    UnexpectedEnd,
+pub struct LabelFromStrError(LabelFromStrErrorEnum);
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(super) enum LabelFromStrErrorEnum {
+    SymbolChars(SymbolCharsError),
```
ximon18, line 545: Why does this inner type have an `Error` suffix but the `BadSymbol` one does not?

src/base/name/builder.rs
```diff
@@ -648,68 +625,79 @@ impl From<PushError> for FromStrError {
...
+
+//------------ PresentationError ---------------------------------------------
+
+/// An illegal presentation format was encountered.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct PresentationError(PresentationErrorEnum);
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+enum PresentationErrorEnum {
+    BadLabel(LabelFromStrError),
```
ximon18, line 662: This would be more consistent with examples earlier in the review if the inner type were called `BadLabelError` - was it intentionally named differently?

src/base/name/label.rs
```diff
@@ -479,12 +482,12 @@ impl OwnedLabel {
...
         while let Some(ch) = chars.next() {
             if res[0] as usize >= Label::MAX_LEN {
-                return Err(LabelFromStrError::LongLabel);
+                return Err(LabelFromStrErrorEnum::LongLabel.into());
             }
             let ch = match ch {
                 ' '..='-' | '/'..='[' | ']'..='~' => ch as u8,
                 '\\' => parse_escape(&mut chars, res[0] > 0)?,
-                _ => return Err(LabelFromStrError::IllegalCharacter(ch)),
+                _ => return Err(BadSymbol::non_ascii().into()),
```
ximon18, line 490: This new approach seems to lose context. Is it not useful to pass the offending character out?

src/base/scan.rs
```diff
@@ -1041,6 +1041,16 @@ pub(super) enum SymbolCharsEnum {
 }
 
 impl SymbolCharsError {
+    /// Creates a “bad escape” variant of the error.
+    pub(crate) fn bad_escape() -> Self {
```
ximon18, line 1045: Can/should these be `const`?

src/rdata/rfc1035/txt.rs
```diff
@@ -756,7 +756,7 @@ impl<Builder: OctetsBuilder + AsRef<[u8]> + AsMut<[u8]>> TxtBuilder<Builder> {
             Symbol::from_chars(&mut chars).map_err(E::custom)?
         {
             if len == 255 {
-                return Err(E::custom(CharStrError));
+                return Err(E::custom(CharStrError::default()));
```
ximon18, line 759: Isn't calling default on a unit struct a Clippy violation?

partim:
> What I don't get is why things like `pub struct CharStrError;` have become `pub struct CharStrError(());`- what value does that inner unit type add?

It blocks users from creating values other than through calling functions that return them. The reason for this is that it allows us later to change the type definition without it being a breaking change. (E.g., we might want to also use an inner enum.)

ximon18:
> > What I don't get is why things like `pub struct CharStrError;` have become `pub struct CharStrError(());`- what value does that inner unit type add?
> 
> It blocks users from creating values other than through calling functions that return them. The reason for this is that it allows us later to change the type definition without it being a breaking change. (E.g., we might want to also use an inner enum.)

Ah because that inner unit type is private, which is hard to see at a glance when using a unit struct. Thanks.

src/base/name/builder.rs
partim: I just figured it doesn’t need the suffix because the name kind of makes it obvious that it is an error type – and a shorter name is nicer?

src/base/name/builder.rs
partim: `LabelFromStrError` is a public type used by `Label`s `FromStr` impl, so I think the name makes sense. The variant name makes sense to me as it indicates that a label was rejected.

src/rdata/rfc1035/txt.rs
partim: I don’t think so – but keep in mind that `CharStrError` isn’t a unit type any more. You can’t do `CharStrError(())` here either because the inner unit is private.

In any case, this code has disappeared with an earlier PR.

ximon18 (approved):

PR #279, opened 2024-02-27
Rename Deckard to Stelline.

ximon18:
Maybe we also need a README noting the history of this testing approach?


partim:
Some introductory comments in `test/net/stelline/mod.rs` would probably be nice.

Would it make sense to lift this from `test` to `src`?

partim (approved):

ximon18:
> Would it make sense to lift this from `test` to `src`?

Maybe wait with that until we maybe move it all to src so that SystemTime can be mocked via `#[if(cfg)]` as needed by my server side PR #274.

ximon18:
> Some introductory comments in `test/net/stelline/mod.rs` would probably be nice.

As the creator and person with best knowledge of the history and context I hope that is someting @Philip-NLnetLabs can add.

tests/net/stelline/README.md
```diff
@@ -0,0 +1,21 @@
+# Stelline
+
+Stelline is a test runner based around DNS message request/response sequences defined in a human readable text file format.
+
+## Etymology and origins
+
+The Stelline replay file format is a variant of the original `.rpl` format used by the [Unbound](https://www.unbound.net/) test tool "testbound".
```
Philip-NLnetLabs, line 7: testbound is based on (or inspired by) on ldns-testns

ximon18: Is that better?

Philip-NLnetLabs: So it should be 'testbound was originally based on ldns-testns'

Philip-NLnetLabs (approved):
Looks good to me

tests/net/stelline/README.md
ximon18: I made some changes, let me know if it's good now.

PR #280, opened 2024-02-27
Fix nightly warnings.

partim:
This PR fixes warnings from lints added to the nightly compiler. It also changes the CI workflow to error out when there are warnings.


ximon18 (approved):
LGTM!

PR #282, opened 2024-02-27
Remove the cache state machine and loop.

ximon18:
PoC draft PR to better demonstrate my idea regarding review feedback on PR #275.

### DIFF
--- a/src/net/client/cache.rs
+++ b/src/net/client/cache.rs
@@ -4,13 +4,13 @@
 #![warn(clippy::missing_docs_in_private_items)]
 
 use crate::base::iana::{Class, Opcode, OptRcode, Rtype};
-use crate::base::name::ToDname;
-use crate::base::Dname;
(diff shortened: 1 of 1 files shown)


Philip-NLnetLabs (commented):
This is not great because it does not take prefetching into account. Part of the cache lookup is a potentially blocking request to a worker thread to do prefetching. There are other things that may need to be done. For example, to avoid multiple requests going out for the same name (unless that is done by another component).

The current code is easy to extend. Let's not lock it into something that is hard to extend.

ximon18:
Thanks @Philip-NLnetLabs, appreciated. I'm not sure what you mean by the prefetching comment but I assume that's just me missing something. I also overlooked the issue of preserving state across `.await` boundaries to avoid state loss on future cancellation. I will at some other moment see if I can find a different way to make the code, to me at least, easier to read.

PR #286, opened 2024-03-23
Add a queryable zone tree

ximon18:
This PR adds `zonefile/parsed.rs`, `zonetree/` and related `examples/`, which together enable, and show a library user how, to go from a zone file in presentation format to an in-memory tree of zones which can be queried to provide an answer, or walked (iterated over).

The tree also supports versioned write operations and a trait based zone implementation allowing for the in-memory tree for a zone to instead be some other (a)synchornous backing store (as demonstrated by the `mysql-zone.rs` example).

It also simplifies the return type of `Service::call()` slightly by moving part of the type signature to the `Transaction` type.


src/zonetree/answer.rs
```diff
@@ -0,0 +1,216 @@
...
+///
+/// [`Answer`] is the type returned by [`Zone::query()`].
+///
+/// Callers of [`Zone::query()`] will likely only ever need to use the
+/// [`Self::to_message()`] function. Alternatively, for complete control use
+/// the getter functions on [`Answer`] instead and construct a response
+/// message yourself using [`MessageBuilder`].
+///
+/// Implementers of alternate backing stores for [`Zone`]s will need to use
+/// one of the various `Answer` constructor functions when [`Zone::query()`]
```
partim, line 23: Apparently, the preferred way to refer to functions is without the parentheses.

src/zonetree/answer.rs
```diff
@@ -0,0 +1,216 @@
...
+    pub fn add_answer(&mut self, answer: SharedRrset) {
+        self.content = AnswerContent::Data(answer);
+    }
+
+    /// Sets the content of the authority section.
+    pub fn set_authority(&mut self, authority: AnswerAuthority) {
+        self.authority = Some(authority)
+    }
+
+    /// Prepares a response to the given request based on the properties of
```
partim, line 86: Random nit-pickery: The first line of a docstring should really just be one line (within the line length restrictions).

src/zonetree/in_memory/builder.rs
```diff
@@ -0,0 +1,172 @@
...
+        rrset: SharedRrset,
+    ) -> Result<(), OutOfZone> {
+        match self.get_node(self.apex.prepare_name(name)?) {
+            Ok(node) => node.rrsets().update(rrset, Version::default()),
+            Err(apex) => apex.rrsets().update(rrset, Version::default()),
+        }
+        Ok(())
+    }
+
+    /// Insert one or more resource records that represent a zone cut.
```
partim, line 101: What this actually does is convert a node (or owner name?) into a zone cut – it doesn’t actually insert RRsets into the RRsets of the node.

Maybe the name of the method is poorly chosen.

(This also applies to `insert_cname` below.)

partim (approved):

src/zonetree/answer.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// this answer.
+    ///
+    /// The response [Rcode], question, answer and authority sections of the
+    /// produced [`AdditionalBuilder`] will be populated based on the
+    /// properties of this [`Answer`] as determined by the constructor and
+    /// add/set functions called prior to calling this function.
+    ///
+    /// See also: [`MessageBuilder::start_answer()`]
+    pub fn to_message<RequestOctets: Octets, Target: Composer>(
```
Philip-NLnetLabs, line 96: It is not clear to me where the AA flag gets set

src/base/opt/cookie.rs
```diff
@@ -359,6 +359,14 @@ impl hash::Hash for ClientCookie {
     }
 }
 
+//--- PartialEq
+
+impl PartialEq for ClientCookie {
```
Philip-NLnetLabs, line 364: Why does PartialEq need to be implemented manually?

Philip-NLnetLabs (commented):
Some comments

src/zonetree/nodes.rs
```diff
@@ -0,0 +1,410 @@
...
+        }
+    }
+
+    /// Returns the string version of the apex name.
+    pub fn apex_name_display(&self) -> &str {
+        &self.apex_name_display
+    }
+
+    /// Returns the class name.
+    pub fn display_class(&self) -> Cow<str> {
```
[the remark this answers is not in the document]
ximon18: Done.

src/zonetree/answer.rs
ximon18: I will change it, though fns without parens look to me. For my education, what are you basing that on? Is it because of this in the rustdoc book? From: https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html

> You can also disambiguate for functions by adding () after the function name, or for macros by adding ! after the macro name. The macro ! can be followed by (), {}, or []. Example:

ximon18: FYI other existing code in domain already uses the`[xxx()]` rustdoc syntax, e.g. `OptRcode`.

ximon18: Done, in all places I found it both in code changed in this branch and elsewhere to make it consistent.

src/zonetree/answer.rs
ximon18: Changed.

src/zonetree/in_memory/builder.rs
ximon18: Does it even convert? It actually sort of does "insert" because if the "special" part of the node is empty it will push a value into it.

I don't think the caller really needs to know what it does internally, from the caller perspective the operations on a zone builder insert high level zone file entities into the zone. Perhaps one could argue that it doesn't insert "resource records" into its own structure, but from a caller perspective it takes resource records (in the form of `Rrset` and `SharedRrset` and `StoredRecord`) and "inserts" them into the zone builder, i.e. does something with them/processes them/whatever.

I don't have a better name or description at the moment and as you say if we change this maybe we should change the other fns too. It's a vague term but we could change them to `process_xxx` instead of `insert_xxx`?

src/zonetree/answer.rs
```diff
@@ -0,0 +1,217 @@
...
+/// Implementers of alternate backing stores for [`Zone`]s will need to use
+/// one of the various `Answer` constructor functions when
+/// [`ReadableZone::query()`] is invoked for your zone content in order to
+/// tailor the DNS message produced by [`Self::to_message()`] based on the
+/// outcome of the query.
+///
+/// [`Zone`]: crate::zonetree::Zone
+/// [`ReadableZone::query()`]: crate::zonetree::traits::ReadableZone::query()
+#[derive(Clone)]
+pub struct Answer {
```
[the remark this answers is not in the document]
ximon18: I think because it's supposed to represent a limited "answer" to a zone (tree) query only, not an entire DNS message. `to_message()` turns it into the start of a possible DNS message to send back to the client, but doesn't for example set things like AA (as it doesn't know if the application is truly authoritative for the zone contents I suppose - this relates to a comment you make below about the missing AA flag). Also there isn't currently any logic in `to_message()` to add anything to an additional sectio. @partim may be able to say more about this.

src/zonetree/answer.rs
ximon18: It doesn't get set. See https://github.com/NLnetLabs/domain/pull/286#discussion_r1553417085 above.

src/base/opt/cookie.rs
ximon18: No idea, but that wasn't added/changed in this PR.

src/zonetree/answer.rs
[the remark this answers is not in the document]
ximon18: It could also be because the way the builder works the caller wouldn't know what had already been added to the additional section and would have trouble therefore adding its own stuff to it, could that be part of it?

src/zonetree/answer.rs
```diff
@@ -0,0 +1,217 @@
...
+    Data(SharedRrset),
+    Cname(SharedRr),
+    NoData,
+}
+
+//------------ AnswerAuthority -----------------------------------------------
+
+/// The authority section of a query answer.
+#[derive(Clone)]
+pub struct AnswerAuthority {
```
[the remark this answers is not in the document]
ximon18: This seems to be a set of things that `Answer::to_message()` knows how to handle, rather than arbitrary authority section records, i.e. it isn't intended to do as you suggest. @partim?

src/zonetree/answer.rs
```diff
@@ -0,0 +1,217 @@
...
+            rcode,
+            content: AnswerContent::NoData,
+            authority: Default::default(),
+        }
+    }
+
+    /// Creates a new message with a populated authority section.
+    ///
+    /// The answer and additional sections will be empty.
+    pub fn with_authority(rcode: Rcode, authority: AnswerAuthority) -> Self {
```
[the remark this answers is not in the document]
ximon18: Also, it doesn't linkify `AnswerContent` or `AnswerAuthority` in the return value position while it does linkify `Rcode` in the same position.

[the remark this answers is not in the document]
ximon18: It's because those types are not exported from `zonetree/mod.rs`. I suspect that the intent is that only the tree can create `Answer` and thus any setters should be private, and then those types are only used internally and not for public view.

src/zonetree/answer.rs
partim: I think this is (or was at some point) in one of the style guides. Can’t find it right now. But this is the style that e.g. the std library uses.

src/zonetree/answer.rs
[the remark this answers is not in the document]
partim: The intention was to have the additional data as part of the `SharedRrset` in the `content`. That just hasn’t happened yet.

src/zonetree/answer.rs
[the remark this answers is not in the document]
ximon18: Actually, we need to expose those types and setter functions on `Answer` because `ReadableZone` allows a user defined backing store for `Zone` to be created and that needs to return the `Answer` type from calls to `query()`.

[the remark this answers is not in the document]
ximon18: Fixed.

src/base/opt/cookie.rs
Philip-NLnetLabs: It comes from commit b9e581910e3d088fecdd05a2d5b3dc3ca23e06d3 which is part of this PR.

ximon18: Oh. My apologies. Then it will be replaced with a Derive or removed entirely if not needed, I don't recall that change at all.

src/zonetree/in_memory/builder.rs
ximon18: Added to the todo list.

src/zonetree/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+//! use domain::base::name::Dname;
+//! use domain::zonefile::{inplace, parsed};
+//! use domain::zonetree::{Answer, Zone, ZoneBuilder, ZoneTree};
+//!
+//! // Prepare some zone file bytes to demonstrate with.
+//! let zone_file = include_bytes!("../../test-data/zonefiles/nsd-example.txt");
+//! let mut zone_bytes = std::io::BufReader::new(&zone_file[..]);
+//!
+//! // Read, parse and build a zone.
+//! let reader = inplace::Zonefile::load(&mut zone_bytes).unwrap();
```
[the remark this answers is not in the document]
ximon18: Added to the TODO list.

src/zonetree/answer.rs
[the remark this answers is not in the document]
ximon18: Added to the TODO list.

PR #287, opened 2024-03-26
Remove source_addr from ComposeRequest and AsyncConnect

Philip-NLnetLabs:


test-data/server/edns_downstream_cookies.rpl.not
ximon18, line 1: Great solution, renaming it to prevent it being run, but would you mind if we rename it to `.disabled` instead of `.not`?

ximon18 (approved):

test-data/server/edns_downstream_cookies.rpl.not
Philip-NLnetLabs: No problem

PR #288, opened 2024-03-29
add AsRef impl for Record to allow for passing records by references when signing

hunts:
One of the use case is to allow for both owned or borrowed records
in the mutable slice when composing the signed data.

Without this, in some cases, the caller would have to clone records
if they only hold a (im)mutable slice of record references or an immutable
slice of record values.


partim (approved):

hunts:
Thank you Martin, cheers!

PR #289, opened 2024-04-03
Make dname_from_addr public.

### LINKED ISSUE #218: function dname_from_addr
Source: https://github.com/NLnetLabs/domain/blob/v0.8.0/src/resolv/lookup/addr.rs#L101

Could this function be public in your crate ?
It's so neat I copied it into my project and linked the source

partim:
This PR makes the helper method `dname_from_addr` for creating the query name for a reverse pointer query public as an associated function on `Dname`. It also improves the IPv4 version by avoiding the `FromStr` impl and creating the domain name more properly.

Fixes #218.


src/base/name/dname.rs
```diff
@@ -181,6 +182,42 @@ impl<Octs> Dname<Octs> {
     {
         unsafe { Self::from_octets_unchecked(b"\0".as_ref().into()) }
     }
+
+    /// Creates a domain name for reverse IP address lookup.
+    ///
+    /// The returned name will use the standard suffixes of `in-addr.arpa.`
+    /// for IPv4 addresses and `ip6.arpa.` for IPv6.
+    pub fn from_addr(addr: IpAddr) -> Result<Self, PushError>
```
ximon18, line 190: I think it would be good to have "reverse" in the name somewhere as I misunderstood at first glance thinking oh is this going to resolve the address to a name for me?

src/base/name/builder.rs
```diff
@@ -266,6 +266,70 @@ where
         Ok(())
     }
 
+    /// Appends a label with the decimal representation of `u8`.
```
ximon18, line 269: I don't understand how this is "decimal" representation, it looks like a conversion to ASCII, what is "decimal" a reference to here? Also, a unit test might be good to have for this function.

partim: Decimal as in “representation of the value using decimal digits” – I think that’s sort of clear?

ximon18: Is not "Appends a label with the decimal interpretation of `u8`" rather than representation?

src/base/name/dname.rs
partim: How about `reverse_from_addr`?

src/base/name/builder.rs
partim: Strictly speaking, we are not interpreting the value (i.e., trying to determine its meaning) but only represent it as text.

ximon18: I read the arithmetic you are doing as interpreting the value. You are assuming its format and interpreting it according to the rules of that format. Anyway, if you feel that the name is correct please ignore this and move on.

Philip-NLnetLabs: I think decimal representation is the canonical expression in this case. https://en.wikipedia.org/wiki/Decimal_representation

ximon18 (approved):

PR #290, opened 2024-04-05
Rename Dname to Name.

### LINKED ISSUE #222: Difference between domain::base::name::Dname and domain::rdata::dname::Dname?
I've gotten these types mixed up a few times:

- https://docs.rs/domain/latest/domain/base/name/struct.Dname.html
- https://docs.rs/domain/latest/domain/rdata/dname/struct.Dname.html

How exactly do they differ? In which cases in the second one used? It is possible to maybe fold them into a single type? The same-name does lead to a bit of confusion.

partim:
This PR renames all types and functions that refer to domain name from using the term “dname” to just “name.” For instance, `Dname` becomes `Name`, `ToDname` becomes `ToName`, and `ToDname::to_dname` becomes `ToName::to_name`.

This annoyingly wide-reaching change is motivated by a clash with the DNAME record type that would always be confusing. So we rather do it now.

This is a really breaking change. Apologies.

Fixes #222.


Philip-NLnetLabs (approved):
Looks fine to me.

PR #295, opened 2024-04-12
Fix brittle test

ximon18:
An example failure can be seen here: https://github.com/NLnetLabs/domain/actions/runs/8659258392/job/23744751321#step:8:526

This PR explicitly terminates mock client connections during the test once they have received all of their responses, rather than leave them connected till the end of the test when the server shuts down. This hopefully helps with the `net::server::tests::service_test` which works fine locally but is brittle on GH Actions.

Result: 10 successful runs can be seen here: https://github.com/NLnetLabs/domain/actions/runs/8660164375

### DIFF
--- a/src/net/server/tests.rs
+++ b/src/net/server/tests.rs
@@ -41,6 +41,8 @@ struct MockStream {
 
     /// The rate at which messages should be made available to the server.
     new_message_every: Duration,
+
+    pending_responses: usize,
 }
 
 impl MockStream {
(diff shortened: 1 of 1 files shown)


src/net/server/tests.rs
```diff
@@ -80,11 +84,13 @@ impl AsyncRead for MockStream {
                         last_ready.replace(Instant::now());
                         return Poll::Ready(Ok(()));
                     } else {
-                        // End of stream
-                        /*return Poll::Ready(Err(io::Error::new(
```
partim, line 84: I s’pose you could just delete the commented out code?

partim (approved):

PR #296, opened 2024-04-12
Redefine stub resolver transport types.

### LINKED ISSUE #91: domain::resolv::stub::StubResolver doesn't fail over to TCP if there's no TCP entry in the ResolvConf
This might not be a bug, but it's perhaps surprising that the stub resolver doesn't fail over to TCP on truncation unless there's TCP explicitly added in the ResolvConf::servers. I would assume that adding a server with protocol UDP means you prefer to talk to that server over UDP, but failover on truncation would work. Similarly, if you add a server with protocol TCP it means you prefer to talk to server over TCP and there failover on truncation doesn't apply. I am happy to add a PR but thought I would open a ticket first to get your input on what would be the best way to solve this. I would probably make the failover to TCP on TC automatic with no changes to the API, but it would also work if you added something like `Protocol::UdpOrTcp` or convey this through some other way.

partim:
This PR redefines the stub resolver’s `Transport::Udp` type to be `Transport::UdpTcp`, i.e., use UDP but retry with TCP if the answer is truncated and slightly reworks the server selection code in the stub resolver to use all servers if `use_vc` is set.

This is a breaking change.

Fixes #91.

### DIFF
--- a/src/resolv/stub/mod.rs
+++ b/src/resolv/stub/mod.rs
@@ -146,35 +146,29 @@ impl StubResolver {
 
         // We have 3 modes of operation: use_vc: only use TCP, ign_tc: only
         // UDP no fallback to TCP, and normal with is UDP falling back to TCP.
-        if self.options.use_vc {
-            for s in &self.servers {
-                if let Transport::Tcp = s.transport {
-                    let (conn, tran) = multi_stream::Connection::new(
-                        TcpConnect::new(s.addr),
-                    );
-                    // Start the run function on a separate task.
-                    let run_fut = tran.run();
-                    fut_list_tcp.push(async move {
-                        run_fut.await;
-                    });
-                    redun.add(Box::new(conn)).await?;
-                }
-            }
-        } else {
-            for s in &self.servers {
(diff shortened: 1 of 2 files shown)


src/resolv/stub/mod.rs
```diff
@@ -146,35 +146,29 @@ impl StubResolver {
...
+        for s in &self.servers {
+            // This assumes that Transport only has UdpTcp and Tcp. Sadly, a
+            // match doesn’t work here because of the use_cv flag.
+            if self.options.use_vc || matches!(s.transport, Transport::Tcp) {
+                let (conn, tran) =
+                    multi_stream::Connection::new(TcpConnect::new(s.addr));
+                // Start the run function on a separate task.
+                let run_fut = tran.run();
+                fut_list_tcp.push(async move {
+                    run_fut.await;
```
gthess, line 159: Is `let run_fut` still needed here?
```suggestion
                fut_list_tcp.push(async move {
                    tran.run().await;
```

Philip-NLnetLabs (approved):

PR #297, opened 2024-04-15
Fix compilation with feature zonefile by moving zonetree related functionality out of zonefile.

ximon18:


partim:
I wonder if it would be better to make `zonefile::parsed::Zonefile` the actual zone builder (i.e., replace `zonetree::in_memory::ZoneBuilder` with it? The two are closely connected, anyway. Merging them may make it less likely to build a broken tree?

ximon18:
The current division feels somewhat artificial so in that regard I concur. However, that is a much bigger change than a compilation fix so perhaps add it to the TODO list rather than make the change in this PR.

partim:
Fair. How about just moving `zonefile::parsed` to `zonetree::parsed`? Having types like `StoredDname` in `zonefile` feels wrong to me.

PR #310, opened 2024-05-02
Minor fix for 0.10.0

xofyarg:
Hi, I included two trivial changes when working on an upgrade from domain-0.7 to 0.10. Please let me know if I missed anything. Thanks.

### DIFF
--- a/src/rdata/dnssec.rs
+++ b/src/rdata/dnssec.rs
@@ -2030,10 +2030,8 @@ impl<Octs> RtypeBitmap<Octs> {
     pub fn scan<S: Scanner<Octets = Octs>>(
         scanner: &mut S,
     ) -> Result<Self, S::Error> {
-        let first = Rtype::scan(scanner)?;
         let mut builder =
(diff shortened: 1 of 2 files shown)


partim:
Thank you for the PR!

I am wondering if the type bitmap can ever be empty? Given that both RRSIG and NSEC(3) seem to be included in it, there should at least be the NSEC(3) record set?

Philip-NLnetLabs:
The NSEC3 bitmap has all of the records present at the owner name, which is empty for an empty non-terminal.

partim:
Ah, right. I suppose the PR can be merge, then.

partim (approved):

xofyarg:
I had the same question, and my colleague pointed me at this one:

> ... the NSEC3 RR exists because it corresponds to an empty non-terminal, in which case the NSEC3 RR will have an empty Type Bit Maps field

https://datatracker.ietf.org/doc/html/rfc5155#section-8.5

PR #312, opened 2024-05-05
Make FamilyName::new() public

achow101:
A `FamilyName` is needed in order to use `SortedRecords::sign()`, however there does not appears to be a way for one to be created. Make `FamilyName::new()` public to facilitate that.


partim (approved):

partim:
Thank you for the PR!

PR #313, opened 2024-05-05
Remove incorrect compose_len from ProtoRrsig

achow101:
`ProtoRrsig` is used to produce the data that is signed. According to https://datatracker.ietf.org/doc/html/rfc4034#autoid-23, this serialization begins with the first rdata field which is `type_covered`, however `ProtoRrsig` is inserting a length at the beginning. As such, signatures produced by `SortedRecords::sign()` are invalid. Fix this by removing that length from the serialization.

### DIFF
--- a/src/rdata/dnssec.rs
+++ b/src/rdata/dnssec.rs
@@ -505,22 +505,10 @@ impl<Name: ToName> ProtoRrsig<Name> {
         self.signer_name.compose_canonical(target)
     }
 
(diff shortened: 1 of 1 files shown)


partim:
Thank you for the PR! Looks like we lost a fix at some point as I’m sure this was all working.

Before we can merge the PR, could you perchance merge main into your branch just so that the CI is happy?

achow101:
Rebased

PR #314, opened 2024-05-05
Don't skip children of apex in RecordsIter if apex is not there

achow101:
If none of the records in the `RecordsIter` have the apex as their name, but they are in the apex's zone, don't skip those children. Otherwise rrsigs will not be created for any queries for just subdomains.

### DIFF
--- a/src/sign/records.rs
+++ b/src/sign/records.rs
@@ -489,7 +489,10 @@ impl<'a, N, D> RecordsIter<'a, N, D> {
         N: ToName,
     {
         while let Some(first) = self.slice.first() {
-            if apex == first {
+            if first.class() != apex.class() {
+                continue;
+            }
+            if apex == first || first.owner().ends_with(apex.owner()) {
                 break;
             }
             self.slice = &self.slice[1..]


achow101:
> I think, however, it should be a separate check before the look rather than testing in every iteration.

I chose to put it in every iteration because theoretically there could be records that would be before the apex, and then records that are children of the apex, but not the apex itself. However I'm not sure if that's actually something reasonable to consider.

partim:
Ah yes, you are right – I forgot about that case. But then, shouldn’t just checking for `ends_with` be enough? And, because this is actually kind of expensive, perhaps add an outer loop that skips over records with the wrong class without checking?

achow101:
> But then, shouldn’t just checking for `ends_with` be enough?

Yes. But I don't fully understand what the point of `class` is, and whether anyone actually uses something other than `IN` for it. I was basing this off of `Family::is_in_zone` which checks for class, so figured this should probably do that too.

> And, because this is actually kind of expensive, perhaps add an outer loop that skips over records with the wrong class without checking?

I moved the class check to be done first at the top of the loop so it would just skip any records that don't have a class matching the apex.

partim:
Apologies for the delay! This looks good now – I will merge it.

PR #316, opened 2024-05-10
Make sure `Found{Hosts, Addrs}` have `is_empty` and `len` methods

tertsdiepraam:
I noticed that `FoundAddrs` did not have an `is_empty` method while implementing the `lookup` command for `dnsi`. I also figured that `len` methods for `FoundHosts` and `FoundAddrs` would be a nice addition.

I just kind of guessed the implementation of these methods, so I don't really know whether they are correct :)


partim:
The answer sections can contain other records (even legitimately, any possible CNAME records go there), so using the ANCOUNT can be wrong. Which also means that the `is_empty` impl in `FoundHosts` is potentially wrong and should be removed.

We could impl methods based on using `self.iter()`, but given that isn’t exactly free, maybe we shouldn’t hide the O(n)-ness.

PR #318, opened 2024-05-22
add some missing methods/trait bounds

xofyarg:
These were existed in previous version, but are otherwise missing.


partim (approved):

partim:
Thank you for the PR. There’s probably some more missing `?Sized` bounds lurking in the code …

hunts:
Yeah, this is another one: #325

PR #320, opened 2024-05-29
Bump to rust 1.78

tertsdiepraam:
Because `triomphe` requires 1.76 and we might as well bump to whatever Alpine has.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        rust: [1.70.0, stable, beta, nightly]
+        rust: [1.78.0, stable, beta, nightly]
     env:
         RUSTFLAGS: "-D warnings"
     steps:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "domain"
 version = "0.10.1-dev"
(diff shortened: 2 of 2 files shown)


partim (approved):

hunts:
It is not always ideal for a library to chase the latest compiler as MSRV. If 1.76 works, I think we should just set to 1.76 rather then 1.78. The end binary depends on domain crate could have situations that no easy to upgrade to latest rustc in timely manner.

partim:
We are using the rustc version that Alpine releases (ie., not edge but currently 3.20) have. Normally they are a few versions back but for some reason they currently have 1.78, so we decided to pick that in the hope to not having the bump again soon.

If you provide some guidance as to what your requirements are, we happily change this policy.

tertsdiepraam:
Not using 1.78 is fine with me, but I was looking into using `{split_,}{first,last}_chunk{,_mut}` which were introduced in 1.77. Another possibly useful change in 1.77 is that the `net` module was moved to `core`.

hunts:
I don't think it is a deal breaker when domain bump the Rust MSRV to latest if there is a need. We always have the option to use a forked version. But just let you know the potential issue with bumping MSRV.

In our use case, the DNS server integrates with a WebAssembly runtime as the sandbox for hosting AoT compiled WSAM programs. Upgrade rustc is sometimes difficult if they don't play nicely together. It takes quite some time for us to fix and re-release those things.

PR #321, opened 2024-05-29
Domain name compression is broken in 0.10.0

hunts:
There is a bug in `SliceLabelsIter`'s `next()` implement that is causing it not able to return the full/right labels in the octets slice.

`StaticCompressor` leverages `SliceLabelsIter` to search known domain names in the underlying octets, is not able to find existing names because of this. Thus, leading to broken domain name compression in 0.10.0.


partim:
Thank you for the PR!

This actually used to use `compose_len` – I probably broke it when changing that function to return a `u16`.

partim (approved):

PR #330, opened 2024-06-07
Proposal for a MSRV policy.

partim:
This PR proposes a revised policy for dealing with minimum supported Rust versions.

It suggests to keep the `rust-version` field in `Cargo.toml` at the oldest Rust version that is able to builds with minimal dependency versions and all features enabled. In addition, the Rust version necessary to build with latest dependencies is provided in the README and as a comment in `Cargo.toml`. Both these versions are checked in the CI workflow.

The reasoning behind the proposal is that `rust-version` sets a hard limit on the version, so it should be chosen as low as possible. While technically it could be lower with a subset of the features selected, this feels like a slippery slope and setting the version with all features selected feels like an acceptable compromise.

Only documenting the version that is necessary to build with latest dependencies isn’t ideal but at least provides some guidance as to what to choose for people who go look for some.

This is a draft PR since it is a proposal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,24 @@ If you have ideas, requests, or proposals for future features, please
 don’t hesitate to open Github issues.
 
 
+## Minimal Supported Rust Versions
+
+We are trying to maintain a decent balance between being able to use older
+compiler versions and using new features that improve safety and usability
+of the crate.
+
+### Minimum version: 1.68.2
+
+The `rust-version` given in `Cargo.toml` is the oldest version that can
+be used to compile the crate with the minimum versions of all dependencies
+selected.
+
+### Current version: 1.76.0
+
+This is the minimum Rust version required to build with latest version of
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -24,6 +24,24 @@ If you have ideas, requests, or proposals for future features, please
...
+We are trying to maintain a decent balance between being able to use older
+compiler versions and using new features that improve safety and usability
+of the crate.
+
+### Minimum version: 1.68.2
+
+The `rust-version` given in `Cargo.toml` is the oldest version that can
+be used to compile the crate with the minimum versions of all dependencies
+selected.
+
```
Philip-NLnetLabs, line 38: One thing I would add to the documentation is that the command 'cargo +nightly update -Z minimal-versions' gives the minimum version of the dependencies. That might be helpful for people trying to compile using an older version.

PR #331, opened 2024-06-07
Removed unnecessary clippy allows and rustfmt suggestions

tertsdiepraam:


src/stelline/channel.rs
```diff
@@ -178,15 +177,15 @@ impl Clone for ClientServerChannel {
 }
 
 impl ClientServerChannel {
-    #[allow(dead_code)]
+    // #[allow(dead_code)]
```
hunts, line 180: Can be removed as well?

tertsdiepraam: Indeed, I missed those. Thank you!

partim (approved):

PR #333, opened 2024-06-11
Fix TSIG position calculation if there are other additional records.

partim:
This PR fixes a mistake in the `tsig` module while calculating the start of the TSIG record when there were other records in the additional section, causing the TSIG code to fail if OPT records were in use.

### DIFF
--- a/src/tsig/mod.rs
+++ b/src/tsig/mod.rs
@@ -1316,11 +1316,12 @@ impl<'a, Octs: Octets + ?Sized> MessageTsig<'a, Octs> {
         let mut start = section.pos();
         let mut record = section.next()?;
         loop {
+            let record_start = section.pos();
             record = match section.next() {
                 Some(record) => record,
                 None => break,
             };
-            start = section.pos();
+            start = record_start;
         }
         record
             .ok()?
--- a/src/tsig/interop.rs
+++ b/src/tsig/interop.rs
@@ -7,6 +7,7 @@ use crate::base::message_builder::{
     AdditionalBuilder, AnswerBuilder, MessageBuilder, StreamTarget,
 };
 use crate::base::name::Name;
+use crate::base::opt::TcpKeepalive;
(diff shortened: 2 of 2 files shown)


src/tsig/mod.rs
```diff
@@ -1316,11 +1316,12 @@ impl<'a, Octs: Octets + ?Sized> MessageTsig<'a, Octs> {
         let mut start = section.pos();
         let mut record = section.next()?;
         loop {
+            let record_start = section.pos();
             record = match section.next() {
```
tertsdiepraam, line 1320: As far as I can tell, this doesn't seem to do the thing it says in the comment which is "checks that there is exactly one TSIG record in the additional section". Should it be doing that or should that be removed from the comment (or am I missing something)?

partim: True, it only checks that the last record is a TSIG record. I want to say that’s fine, but I’m honestly not sure.

tertsdiepraam: RFC 8945 states:
> If an incoming message contains a TSIG record, it MUST be the last record in the additional section. Multiple TSIG records are not allowed. If multiple TSIG records are detected or a TSIG record is present in any other position, the DNS message is dropped and a response with RCODE 1 (FORMERR) MUST be returned

Src: https://datatracker.ietf.org/doc/html/rfc8945#name-server-processing-of-reques

So it seems like something that's probably a good thing to do. But it also sounds like it should return `Err` in this case instead of `None`.

partim: Thanks for digging that up! So I guess that should indeed be fixed. Perhaps it’s better in a separate PR? (What with it changing the return type and all that.)

tertsdiepraam: Yes that's probably a good idea

ximon18 (approved):
This fixes the issue I was seeing in manual testing with both NSD and BIND.

PR #334, opened 2024-06-11
Ensure TSIG is the last record

tertsdiepraam:


src/tsig/mod.rs
```diff
@@ -1311,23 +1333,29 @@ impl<'a, Octs: Octets + ?Sized> MessageTsig<'a, Octs> {
     /// Checks that there is exactly one TSIG record in the additional
     /// section, that it is the last record in this section. If that is true,
     /// returns the parsed TSIG records.
-    fn from_message(msg: &'a Message<Octs>) -> Option<Self> {
-        let mut section = msg.additional().ok()?;
+    fn from_message(msg: &'a Message<Octs>) -> Result<Self, TsigError> {
+        let mut section = msg.additional().map_err(|_| TsigError::Missing)?;
```
partim, line 1337: Maybe this should actually return something that gets turned into a form error? Processing is going to fail later, anyway, so might just return the error early?

src/tsig/mod.rs
```diff
@@ -1311,23 +1333,29 @@ impl<'a, Octs: Octets + ?Sized> MessageTsig<'a, Octs> {
...
+
+        // Find the first TSIG record, which we will assert to be the last
+        // one to verify that it is the only one.
```
partim, line 1340: Sounds very poetic ;)

src/tsig/mod.rs
```diff
@@ -1311,23 +1333,29 @@ impl<'a, Octs: Octets + ?Sized> MessageTsig<'a, Octs> {
...
+                .map_err(|_| TsigError::Invalid)?;
+
+            if let Some(record) = record {
+                // We got a valid TSIG, now assert that it's the last record:
+                if section.next().is_some() {
+                    return Err(TsigError::Position);
+                }
+                return Ok(MessageTsig { record, start });
+            }
+            start = section.pos();
```
partim, line 1355: We don’t really need to mutable and updated `start` any more. You can just grab `section.pos()` at the start of the iteration now.

src/tsig/mod.rs
```diff
@@ -1870,3 +1898,16 @@ impl fmt::Display for ValidationError {
...
 impl std::error::Error for ValidationError {}
+
+//------------ TsigError -----------------------------------------------------
+
+/// Error while retrieving a TSIG from a message
+#[derive(Debug)]
+enum TsigError {
+    /// The TSIG record was present but malformed
+    Invalid,
+    /// A TSIG record was found that was **not** the last record in the message
```
partim, line 1909: Can I have empty line between the variants when there are docstrings, please?

partim:
We probably should a bunch of test cases for this.

ximon18 (approved):

PR #335, opened 2024-06-11
RFC 1035 zone maintenance support.

ximon18:
This PR introduces the `ZoneMaintainer` component.

For primary zones it provides "on-zone-change" triggered sending of NOTIFY messages zones to "secondaries".

For "secondary" zones it provides NOTIFY-in and SOA timer based SOA query and (A/I)XFR-out.

Zone persistence on change (either edit to a local primary zone or sync of a local secondary zone with changes obtained from a remote primary) is also currently included in this PR (see `examples/serve-zone.rs` via the `ArchiveZone` impl of the `ZoneStore` trait and by using the same `Zone` wrapping "hack" that `ZoneMaintainer` uses to monitor zones for changes).

This PR needs cleaning up before review.


examples/serve-zone.rs
```diff
@@ -2,10 +2,9 @@
 //!
 //! Try queries such as:
 //!
-//!   dig @127.0.0.1 -p 8053 NS example.com
-//!   dig @127.0.0.1 -p 8053 A example.com
-//!   dig @127.0.0.1 -p 8053 AAAA example.com
-//!   dig @127.0.0.1 -p 8053 CNAME example.com
+//!   dig @127.0.0.1 -p 8053 NS example.com dig @127.0.0.1 -p 8053 A
+//!   example.com dig @127.0.0.1 -p 8053 AAAA example.com dig @127.0.0.1 -p
+//!   8053 CNAME example.com
```
tertsdiepraam, line 7: This reformatting seems to have gone wrong somehow. Maybe we should put this in code fences instead of indenting?

src/net/client/auth.rs
```diff
@@ -0,0 +1,392 @@
...
+}
+
+impl<CR, Upstream, K> Debug for Request<CR, Upstream, K>
+where
+    CR: ComposeRequest,
+    Upstream: SendRequest<AuthenticatedRequestMessage<CR, K>>,
+{
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), core::fmt::Error> {
+        f.debug_struct("Request")
+            .field("fut", &format_args!("_"))
```
tertsdiepraam, line 223:
```suggestion
            .field("fut", &"_")
```

src/net/client/stream.rs
```diff
@@ -78,21 +93,73 @@ impl Config {
 
     /// Sets the response timeout.
     ///
+    /// Forrequests where ComposeRequest::is_streaming() returns true see
```
tertsdiepraam, line 96:
```suggestion
    /// For requests where `ComposeRequest::is_streaming()` returns true see
```

src/net/server/connection.rs
```diff
@@ -923,35 +978,35 @@ where
...
-                ControlFlow::Continue(())
-            }
-            _ => {
-                // Everything else is either unrecoverable or unknown to us at
-                // the time of writing and so we can't guess how to handle it,
-                // so abort.
-                error!("I/O error: {}", err);
-                ControlFlow::Break(ConnectionEvent::DisconnectWithoutFlush)
-            }
+/// Handle I/O errors by deciding whether to log them, and whethr to continue
```
tertsdiepraam, line 990:
```suggestion
/// Handle I/O errors by deciding whether to log them, and whether to continue
```

tertsdiepraam (commented):
This is all still massively going over my head, but I found some typos :)

ximon18:
Thx @tertsdiepraam !

The code is not finished functionally yet, thus certainly not  in terms of docs, comments, examples etc either.

So I'd suggest to limit yourself to high level review at this point.

I'd be happy to walk you through it if you would appreciate that, either in person next week or online before then.

tertsdiepraam:
Oh yeah don't mind me. I was just interested to look through it and figured I'd comment on some things while I was going through it. I'd love a walkthrough at some point though.

PR #336, opened 2024-06-13
Re-enable and fix the Stelline cookies test

ximon18:
PR #274 included a DNS cookies middleware processor and a corresponding Stelline `.rpl` test.

However, due to the mechanism to mock server time (needed to by the cookies rpl test) not being ready at that point the necessary support for the test was lacking so the test was merged to main in a disabled state (by renaming it to `.not` so that it was not included in the .rpl tests run by `net::server::tests::server_tests()`.

Since then PRs #315 and #328 added the majority of the pieces that were missing making it possible for this PR to re-enable the Stelline cookies test and add further fixes & missing support to make all of its steps runnable and pass.

This PR also:
- Uses a newer version of `mock_instant` which has support for thread local mock system time, ensuring that tests that mock time (which run in their own single thread) are able to run in parallel without impacting each other.
- Simplifies the server-side DNS cookie code by removing  `#[cfg(feature = "siphasher")]` guards and instead makes the `unstable-server-transports` feature depend on the `siphasher` feature.


Cargo.toml
```diff
@@ -60,9 +60,9 @@ validate    = ["bytes", "std", "ring"]
 zonefile    = ["bytes", "serde", "std"]
 
 # Unstable features
-unstable-client-transport = [ "moka", "net", "tracing" ]
-unstable-server-transport = ["arc-swap", "chrono/clock", "libc", "net", "tracing"]
-unstable-stelline = ["tokio/test-util", "tracing", "tracing-subscriber", "unstable-server-transport", "zonefile"]
+unstable-client-transport = ["moka", "net", "tracing"]
+unstable-server-transport = ["arc-swap", "chrono/clock", "libc", "net", "siphasher", "tracing"]
+unstable-stelline = ["tokio/test-util", "tracing", "tracing-subscriber", "unstable-server-transport", "zonefile", "mock_instant"]
```
partim, line 65: How does this even work when `mock_instant` is only a dev dependency?

Philip-NLnetLabs (approved):

PR #340, opened 2024-06-18
Add support for netblocks to `CookiesMiddlewareProcessor`.

ximon18:
This PR adds support (via new dependency `ipnetwork`) for net blocks in the deny list of the cookie middleware processor, ala Unbound, otherwise the deny list is difficult to use beyond a few IP addresses.


partim:
You might want to consider using [inetnum](https://github.com/NLnetLabs/inetnum) rather than `ipnetwork` – and add things that are potentially missing.

ximon18:
> You might want to consider using [inetnum](https://github.com/NLnetLabs/inetnum) rather than `ipnetwork` – and add things that are potentially missing.

Nice idea, but inetnum seems to lack all of the needed functionality.

ximon18:
> > You might want to consider using [inetnum](https://github.com/NLnetLabs/inetnum) rather than `ipnetwork` – and add things that are potentially missing.
> 
> Nice idea, but inetnum seems to lack _all_ of the needed functionality.

My apoloiges @partim, of course it has what is needed, I just didn't see it as it is called `Prefix` rather than network or netblock etc.

I have updated this PR to use `inetnum`. I did note one regression compared to using `ipnetwork` which is that `FromStr` for `Prefix` is less accepting than that of `ipnetwork` as it will reject 127.0.0.1/24 as it has non-zero host bits. For end users however while stricter it still provides the functionality that's actually needed.

ximon18:
> > > You might want to consider using [inetnum](https://github.com/NLnetLabs/inetnum) rather than `ipnetwork` – and add things that are potentially missing.
> > 
> > 
> > Nice idea, but inetnum seems to lack _all_ of the needed functionality.
> 
> My apoloiges @partim, of course it has what is needed, I just didn't see it as it is called `Prefix` rather than network or netblock etc.
> 
> I have updated this PR to use `inetnum`. I did note one regression compared to using `ipnetwork` which is that `FromStr` for `Prefix` is less accepting than that of `ipnetwork` as it will reject 127.0.0.1/24 as it has non-zero host bits. For end users however while stricter it still provides the functionality that's actually needed.

I've updated the PR again to use its own `FromStr` impl that handles the bare IP address case which `FromStr` in `inetnum` rejects as lacking a prefix len. @partim and I briefly discussed extending the support in `inetnum` but that snowballed into a larger discussion which is parked for the moment, instead this is the pragmatic step forward for now.

src/net/server/middleware/processors/cookies.rs
```diff
@@ -203,6 +267,14 @@ impl CookiesMiddlewareProcessor {
         //   Cookie, the response SHALL have the RCODE NOERROR."
         self.response_with_cookie(request, Rcode::NOERROR.into())
     }
+
+    /// Is the given IP address required to authenticate itself?
+    ///
+    /// If the given IP address is on our deny list it is required to
+    /// authenticate itself.
+    fn must_authenticate(&self, ip: IpAddr) -> bool {
+        self.deny_list.iter().any(|netblock| netblock.contains(ip))
```
ximon18, line 276: This might not be very efficient for larger numbers of blocks. Maybe use (optionally, i.e. behind a feature gate) https://docs.rs/rotonda-store/latest/rotonda_store/struct.MultiThreadedStore.html here?

Also, looking up configuration based on incoming IP address is also needed for handling incoming XFR requests, so maybe this functionality should be moved to a common module under `src/net/`?

@partim? @density215?

ximon18: Thinking about this a bit more, the lookup result when using a simple vec of netblocks is dependent on the order they are added, so it's not just about efficiency, a proper prefix store will also return the best match rather than just the first match.

partim: There’s probably also going to be rate limiting code that needs to work on prefixes, so having something more general may be useful. Would it make sense to do something generic here? Or are we happy to settle on rotonda-store at least under the hood?

ximon18: Maybe this needs turning around like was done for XFR. For XFR the code used to apply access control but there was quite a lot of configuration and exactly how you want to restrict access is perhaps not something we should decide here. So for XFR I moved the checking out to an `XfrDataProvider` trait that the middleware uses a user supplied impl of. The only sample impls of `XfrDataProvider` actually don't do any access control at all, that's left to the end user. Should we maybe do the same here?

partim: I think that would be a good idea.

ximon18:
As noted [here](https://github.com/NLnetLabs/domain/pull/340#discussion_r1786020095) if we were to do this it would be better to invert it such that we check against a user supplied ACL impl as that allows the user to decide which kind of prefix store they want to use rather than we imposing choices on them. I will close this PR and create an issue to follow this up later.

PR #341, opened 2024-06-19
Option to keep stream connections open if even no Keepalive option is received.

Philip-NLnetLabs:


src/net/client/stream.rs
```diff
@@ -39,13 +40,31 @@ use tokio::time::sleep;
...
+/// TCP connection of an idle session, unless an idle timeout has been
+/// established using some other signalling mechanism, for example,
+/// [edns-tcp-keepalive]."
+/// However, RFC 7858, Section 3.4 says: "In order to amortize TCP and TLS
+/// connection setup costs, clients and servers SHOULD NOT immediately close
+/// a connection after each response.  Instead, clients and servers SHOULD
+/// reuse existing connections for subsequent queries as long as they have
+/// sufficient resources.".
+/// We set the default to 10 seconds, which is that same as what stubby
+/// uses. Minium zero to allow idle timeout to be disabled. Assume that
```
ximon18, line 61:
```suggestion
/// uses. Minimum zero to allow idle timeout to be disabled. Assume that
```

ximon18 (approved):
LGTM. V similar but cleaner than equivalent changes I made in PR #335.

PR #342, opened 2024-06-21
Move `mock_instant` to `dev-dependencies`

tertsdiepraam:


ximon18 (approved):
LGTM. Also tested on my `fix-stelline-cookies-test` branch which actually makes use of `mock_instant` and it works there too.

PR #343, opened 2024-06-21
Add `serde::Serialize` for `AllRecordData`.

tertsdiepraam:
This helps with the JSON output in `dnsi`.

`Deserialize` is left for later, becasue we don't need it yet.

### DIFF
--- a/src/rdata/macros.rs
+++ b/src/rdata/macros.rs
@@ -506,6 +506,17 @@ macro_rules! rdata_types {
         /// This enum collects the record data types for all currently
         /// implemented record types.
         #[derive(Clone)]
+        #[cfg_attr(
+            feature = "serde",
+            derive(serde::Serialize),
+            serde(bound(
+                serialize = "
+                    O: AsRef<[u8]> + octseq::serde::SerializeOctets,
+                    N: serde::Serialize,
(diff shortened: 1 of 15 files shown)


src/rdata/macros.rs
```diff
@@ -506,6 +506,17 @@ macro_rules! rdata_types {
         /// This enum collects the record data types for all currently
         /// implemented record types.
         #[derive(Clone)]
+        #[cfg_attr(
+            feature = "serde",
+            derive(serde::Serialize),
+            serde(bound(
+                serialize = "
+                    O: AsRef<[u8]> + octseq::serde::SerializeOctets,
```
tertsdiepraam, line 514: @partim I copied this from elsewhere, but isn't `AsRef<[u8]>` always enough because we can do `octs.as_ref().serialize_octets(serializer)`?

partim: You are right. It’s possible that `SerializeOctets` was doing more things in an earlier iteration. Or I was just not thinking straight.

partim: Ah, but then you have to do your own `Serialize` impl, no? This way we can just derive it?

PR #349, opened 2024-07-09
Add `PushError::LimitExceeded`.

ximon18:
As discussed in the comments on #348 this PR introduces a new `PushError` variant for the limit exceeded case.

This was not done in the original PR as it is a breaking change.

### DIFF
--- a/src/base/message_builder.rs
+++ b/src/base/message_builder.rs
@@ -421,7 +421,7 @@ impl<Target> MessageBuilder<Target> {
     ///
     /// This message is atop the octets slices derived from the builder, so
     /// it can be created cheaply.
-    pub fn as_message(&self) -> Message<&[u8]>
+    pub fn as_message(&self) -> Message<&[u8]>git
     where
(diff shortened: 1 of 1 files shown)


src/base/message_builder.rs
```diff
@@ -2332,6 +2332,7 @@ impl<Target: Composer> Truncate for TreeCompressor<Target> {
 #[derive(Clone, Copy, Debug)]
 pub enum PushError {
```
tertsdiepraam, line 2333: It might be nice to document these variants to explain what they mean in rustdoc.

tertsdiepraam (commented):
Looks like a good change to me, just one comment about some stuff that was already there.

Philip-NLnetLabs (approved):

src/base/message_builder.rs
ximon18: Pushed. Let me know what you think @partim: Do the provided doc strings match the original intention of the PushError variants that existed prior to this PR?

PR #350, opened 2024-07-09
TSIG: Add `From<ServerTransaction<K>>` to `ServerSequence<K>`.

ximon18:
Useful when needing [`ServerSequence`] in order to sign multiple responses but only having a [`ServerTransaction`].


ximon18:
Closing as this is included in PR #380.

PR #352, opened 2024-07-11
Fix zone walking to include non-leaf CNAMEs

ximon18:
Querying a zone for a CNAME works, but when walking the zone (e.g. for diagnostic dumping or to do AXFR) a CNAME that is not at a leaf node (i.e. a CNAME at a.example.com while there also exists a deeper b.a.example.com leaf node) is incorrectly skipped by the current zone walking code.

This PR adds the missing code to emit the non-leaf CNAME to the zone walker callback operation.

This PR does NOT include unit tests to cover this fix because (a) there are no existing unit tests for this code and writing an entire set would time that I don't have right now, (b) strict unit tests are not really possible because they would also have to exercise the code in `nodes.rs` to construct the tree before wakling it and so some sort of "integration" test is needed, and (c) the new XFR work being done for PR #335 will include Stelline tests that walk zones (for AXFR) and should visit all parts of the tree.

While the Stelline tests will be at the outermost testing layer, exercising domain as a black box, so one could perhaps argue that testing somewhere nearer to the tree walking code would also be good to have, they will at least address the immediate need to exercise the entire zone walking code.

### DIFF
--- a/src/zonetree/in_memory/read.rs
+++ b/src/zonetree/in_memory/read.rs
@@ -101,7 +101,22 @@ impl ReadZone {
                 answer
             }
             Some(Special::NxDomain) => NodeAnswer::nx_domain(),
-            Some(Special::Cname(_)) | None => self.query_children(
+            Some(Special::Cname(cname)) => {
+                if walk.enabled() {
+                    let mut rrset = Rrset::new(Rtype::CNAME, cname.ttl());
+                    rrset.push_data(cname.data().clone());
+                    walk.op(&SharedRrset::new(rrset));
+                }
+
+                self.query_children(
+                    node.children(),
+                    label,
+                    qname,
+                    qtype,
+                    walk,
+                )
+            }
+            None => self.query_children(
                 node.children(),
                 label,
                 qname,


ximon18:
No test is included in this PR, partly because the state of the Stelline server testing in this PR isn't able to test AXFR which is what needs to be queried to exercise zone walking which is what this PR fixes.

A manual test however with NSD, domain without the fix and domain with the fix shows that this PR resolves an issue.

Given the zone file:
```
com.               900     IN  SOA     a.gtld-servers.net. nstld.verisign-grs.com. 1720688795 1800 900 604800 86400
alt.com.           3600    IN  A       1.2.3.4
a.alt.com.           3600  IN  CNAME   example.com.
b.a.alt.com.         3600  IN  A       1.2.3.4
www.terminal.com.   3600   IN  A       1.2.3.4
alt.terminal.com.   3600   IN  CNAME   www.example.com.
```

A sorted AXFR dig query output when run against NSD and domain with the fix in this PR (applied to the `xfr` branch which has full support for AXFR) looks the same:

_(with dig command: `dig +noall +answer @127.0.0.1 -p <PORT> -t AXFR com | sort` where `<PORT>` is the local port on which either NSD or domain are serving the zone)_

```
a.alt.com.              3600    IN      CNAME   example.com.
alt.com.                3600    IN      A       1.2.3.4
alt.terminal.com.       3600    IN      CNAME   www.example.com.
b.a.alt.com.            3600    IN      A       1.2.3.4
com.                    900     IN      SOA     a.gtld-servers.net. nstld.verisign-grs.com. 1720688795 1800 900 604800 86400
com.                    900     IN      SOA     a.gtld-servers.net. nstld.verisign-grs.com. 1720688795 1800 900 604800 86400
www.terminal.com.       3600    IN      A       1.2.3.4
```

Without the fix in this PR domain misses some records in the AXFR output:
```
alt.com.                3600    IN      A       1.2.3.4
b.a.alt.com.            3600    IN      A       1.2.3.4
com.                    900     IN      SOA     a.gtld-servers.net. nstld.verisign-grs.com. 1720688795 1800 900 604800 86400
com.                    900     IN      SOA     a.gtld-servers.net. nstld.verisign-grs.com. 1720688795 1800 900 604800 86400
www.terminal.com.       3600    IN      A       1.2.3.4
```

partim (approved):

PR #356, opened 2024-07-19
Fix TSIG server sequence signing and switch test to dig.

partim:
This PR fixes an issue with signature generation in the TSIG server sequence, i.e., when the answer to a request signed with a TSIG key has multiple messages.

This issue wasn’t caught by the interop test since drill happily consumed to wrong signatures. Consequently, this PR switches the test to use dig which won’t error out but at least complain about them in the output.

### DIFF
--- a/src/tsig/interop.rs
+++ b/src/tsig/interop.rs
@@ -19,6 +19,7 @@ use std::io::{Read, Write};
 use std::net::{IpAddr, SocketAddr, TcpListener, TcpStream, UdpSocket};
 use std::process::Command;
 use std::str::FromStr;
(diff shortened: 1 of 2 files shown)


partim:
This PR should probably replace #345.

ximon18:
The tests pass for me locally but in order to actually verify that DIG isn't complaining I had to modify the code to print `output.stdout`. I think it would be useful to always print this output, and at the moment if the `output.status.success()` test fails you don't see the output then either which is very confusing when trying to work out what went wrong.

ximon18 (approved):
LGTM.

PR #358, opened 2024-07-22
Add `OptRcode::is_ext()` to test if an Rcode is extended per RFC 6891.

ximon18:


partim (approved):

src/base/iana/rcode.rs
```diff
@@ -558,6 +558,20 @@ impl OptRcode {
...
+    pub fn is_ext(&self) -> bool {
+        // https://datatracker.ietf.org/doc/html/rfc6891#section-6.1.3
+        // 6.1.3. OPT Record TTL Field Use
+        //   ...
+        //   "EXTENDED-RCODE
+        //       Forms the upper 8 bits of extended 12-bit RCODE (together
+        //       with the 4 bits defined in [RFC1035].  Note that
+        //       EXTENDED-RCODE value 0 indicates that an unextended RCODE is
+        //       in use (values 0 through 15)."
+        self.0 >> 4 != 0
```
ximon18, line 572: Maybe this should be done with a mask instead of a bit shift.

PR #363, opened 2024-07-24
Allow glue (e.g. A/AAAA) at NS record.

ximon18:


src/base/iana/rtype.rs
```diff
@@ -425,3 +425,13 @@ int_enum! {
 }
 
 int_enum_str_with_prefix!(Rtype, "TYPE", b"TYPE", u16, "unknown record type");
+
+pub trait IsGlue {
```
partim, line 429: What’s the reason to have this as  a trait rather than just a normal method?

src/zonetree/parsed.rs
```diff
@@ -123,36 +123,64 @@ impl Zonefile {
...
+                            .insert(record.owner().clone(), record.into());
+                        Ok(())
                     }
-                    self.cnames.insert(record.owner().clone(), record.into());
-                    Ok(())
                 }
                 _ => {
-                    if self.zone_cuts.contains(record.owner())
-                        || self.cnames.contains(record.owner())
+                    // Only gue records can only be added at the same owner as
```
partim, line 165: The comment probably broke during an edit …

src/base/iana/rtype.rs
ximon18: Errm, none? I saw a macro and avoided it ;-)

ximon18: Done.

partim (approved):

PR #365, opened 2024-07-24
RFC 9619 QDCOUNT is (usually) one

ximon18:
Since it was easy to add.


partim (approved):

partim:
I’m assuming a query count of 0 is checked elsewhere?

PR #369, opened 2024-08-06
Service enhancement prerequisites for XFR support

ximon18:
Taken from the `xfr` branch in order to split PR #335 into several smaller PRs.

This PR makes changes, some of which are breaking (i.e. not backward compatible with earlier unstable server functionality without minor adjustments to existing consumer code, e.g. see the changes to `examples/serve-zone.rs`), including:

- Adds support for indicating to downstream Service impls that a certain number of bytes should be reserved (to make space for adding a TSIG RR or EDNS options during middleware post-processing).

- Adds support for strongly typed passing of arbitrary metadata between middleware that produces a type and middleware that consumes the type (e.g. passing the used TSIG key name from TSIG middleware to XFR middleware).

- Replaces the blanket `impl Service for Arc` with a more general `impl Service for Deref`.

- Removes the blanket `impl Service for fn` as I have multiple times been blocked from defining some other Service impl because the fn one matches, because its presence can cause very confusing compiler error messages if your `impl Service for T` is not quite in sync with the signature of your other `impl T` blocks, and because Hyper doesn't need it so why do we?

- Adds an "enabled" flag to the `CookiesMiddlewareSvc` (and later the `TsigMiddlewareSvc` will have it as well) making it easier to create a service stack with optionally enabled layers based on runtime configuration.

There was some internal discussion today with @Philip-NLnetLabs about passing data between middleware layers instead as a map of string key to some value pairs, or adding an enum to Request with things that a middleware might need (e.g. TSIG key name, or other arbitrary T), but that isn't reflected (yet?) in this PR.

The upside of the approach in this PR is it allows a producing middleware to express the metadata type it produces and a consuming middleware to express compatibility with a certain metadata type (e.g. via a trait impl, see https://github.com/NLnetLabs/domain/commit/88a8f021db0051b82b1692760f492fa0fb12d3e7) and to then have compile time checking that the middleware are compatible with each other, without having to express up front every possible data that might need to be passed around with a request or else as some arbitrary Other(T). But I'm undecided about the best way to do this ...

The changes in this PR currently lack RustDocs, and testing (as the consuming and testing is done at the point of use in the `xfr` branch).

### DIFF
--- a/src/base/message_builder.rs
+++ b/src/base/message_builder.rs
@@ -21,9 +21,7 @@
 //! something that looks like a [`Record`]. Apart from actual values
 //! of these types, tuples of the components also work, such as a pair of a
 //! domain name and a record type for a question or a triple of the owner
-//! name, TTL, and record data for a record. If you already have a question
-//! or record, you can use the `push_ref` method to add
-//!
+//! name, TTL, and record data for a record.
 //!
 //! The `push` method of the record
 //! section builders is also available via the [`RecordSectionBuilder`]
@@ -868,16 +866,6 @@ impl<Target: Composer> AnswerBuilder<Target> {
     pub fn push(
         &mut self,
         record: impl ComposeRecord,
-    ) -> Result<(), PushError> {
-        self.builder.push(
-            |target| record.compose_record(target).map_err(Into::into),
-            |counts| counts.inc_ancount(),
-        )
-    }
-
-    pub fn push_ref(
-        &mut self,
-        record: &impl ComposeRecord,
     ) -> Result<(), PushError> {
         self.builder.push(
             |target| record.compose_record(target).map_err(Into::into),
--- a/src/base/message_builder.rs
+++ b/src/base/message_builder.rs
@@ -1946,14 +1946,6 @@ where
     }
 }
 
-impl<Target> FreezeBuilder for StreamTarget<Target> {
-    type Octets = Target;
-
-    fn freeze(self) -> Self::Octets {
-        self.into_target()
-    }
-}
-
 //------------ StaticCompressor ----------------------------------------------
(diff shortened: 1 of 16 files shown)


src/base/message_builder.rs
```diff
@@ -262,6 +260,31 @@ impl<Target: Composer> MessageBuilder<Target> {
...
+            let header = self.header_mut();
+            header.set_id(msg.header().id());
+            header.set_qr(true);
+            header.set_opcode(msg.header().opcode());
+            header.set_rd(msg.header().rd());
+        }
+        let mut builder = self.question();
+        for item in msg.question().flatten() {
+            if builder.push(item).is_err() {
+                break;
```
Philip-NLnetLabs, line 281: I think this is wrong. Under normal circumstances, a question will always fit. So if it doesn't then the rcode needs to be changed to reflect what is wrong. For example, if the first question doesn't fit then it should be SERVFAIL. If there is a second question, then FORMERR is more appropriate.

Philip-NLnetLabs (commented):
Looks good to me except for start_error.

src/base/message_builder.rs
ximon18: I think I was thinking that if push() fails there are serious memory issues and at that point attempting to make any further changes may also fail, better to just stop and try to work with what we have. I have changed it to reset the question section to empty and set the rcode to SERVFAIL. Is that better?

ximon18: So don't reset the question section, just set rcode to SERVFAIL?

ximon18: Done.

Philip-NLnetLabs (approved):

PR #373, opened 2024-08-08
Add a TSIG client.

ximon18:

Also extends the `client-transports.rs` example to demonstrate TSIG (requires a suitably configured server).


src/net/client/tsig.rs
```diff
@@ -0,0 +1,704 @@
...
+/// TODO
+pub struct Connection<Upstream, K> {
+    /// Upstream transport to use for requests.
+    ///
+    /// This should be the final transport, there should be no further
+    /// modification to the request before it is sent to the recipient.
+    upstream: Arc<Upstream>,
+
+    /// TODO
+    key: Option<K>,
```
Philip-NLnetLabs, line 83: It seems that if key is equal to None, this code silently disables TSIG processing. For a security function that is an accident waiting to happen.

src/net/client/tsig.rs
```diff
@@ -0,0 +1,704 @@
...
+                                        RequestStateMulti::Init => {
+                                            unreachable!()
+                                        }
+
+                                        RequestStateMulti::GetResponse(
+                                            ref mut _request,
+                                            _tsig_client,
+                                        ) => {
+                                            self.state =
+                                                RequestStateMulti::Complete;
```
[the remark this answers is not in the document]
ximon18: Ah it was called in `stream_complete()` which has since been removed in the refactored approach to streaming support in the client code.

src/net/client/tsig.rs
ximon18: It was, if I recall correctly, envisaged as a way to use the same client instance for both signed and unsigned requests, in particular because signing is being added primarily for XFR requests and responses, but a client requiring a key would not then be usable for other types of request, which perhaps on a TCP stream it would be of interest to reuse the stream rather than have to use a separate client and a separate TCP stream.

However, I see your point, if one creates a TSIG client that is because you want to sign requests, and it is a footgun to be able to NOT sign requests with a signing client. I have removed the Option, with the assumption that users can create a separate non-signing client if needed. If the efficiency angle I just referred to about TCP streams is a point of interest we can revisit this, but for now it resolves your concern.

ximon18: You can see the effect this has by looking at https://github.com/NLnetLabs/domain/pull/335/commits/dbf80ced2dbfa376a2a90304091e01e28f197c8c. This issue occurs because the test code doesn't know in advance if it has to create a signing client or not, so has to if else branch and box the created client to erase the different types in order to use them without caring if the client is signing or not. For now I'm hoping that this is an artefact of test code and that real application code wouldn't be as affected.

ximon18:
We should decide what to do with the TSIG client extension that was added to `client-transports.rs`. The example isn't very useful as it won't work without having a server that allows AXFR with TSIG using your key bytes and key name, i.e. only a server you spin up yourself. For everyone else who runs the example it will just fail at the TSIG client step.

The only public DNS server offering open XFR access with TSIG signing is from [Swtich](https://portal.switch.ch/pub/open-data/#tab-fccd70a3-b98e-11ed-9a74-5254009dc73c-3) but we also shouldn't configure the example code to do AXFR to that as (a) it's a bit impolite and (b) we'd have to ask users to accept the Switch terms before running the example.

We could make the example take optional TSIG key name and key file base64 bytes and alg name as input ala dig, and also make the queried host configurable too, and then only run the TSIG client code if TSIG key details were provided.

What do you think @Philip-NLnetLabs and @partim?

src/net/client/tsig.rs
ximon18: **Update:** I see now that it was even documented why the key is optional:

> Supplying the key is optional. The transport only affects the request and response if a key is supplied. This allows for optional signing without having to construct a different client stack.

Philip-NLnetLabs:
I think it is fine if it fails. Maybe print something to that effect. I see the examples mostly as code snippets.

Philip-NLnetLabs (approved):

PR #374, opened 2024-08-12
Fix TXT record equality

dklbreitling:
A TXT record with strings ["foo", "bar"] is not equal to a TXT record with strings ["foob", "ar"].
This is important for resolvers and DNS servers deduplicating RRs. This fixes `PartialEq` accordingly.

### DIFF
--- a/src/rdata/rfc1035/txt.rs
+++ b/src/rdata/rfc1035/txt.rs
@@ -331,9 +331,7 @@ where
     Other: AsRef<[u8]>,
 {
     fn eq(&self, other: &Txt<Other>) -> bool {
-        self.iter()
-            .flat_map(|s| s.iter().copied())
-            .eq(other.iter().flat_map(|s| s.iter().copied()))
+        self.0.as_ref().eq(other.0.as_ref())
     }
 }
 
@@ -347,9 +345,7 @@ where
     Other: AsRef<[u8]>,
 {
     fn partial_cmp(&self, other: &Txt<Other>) -> Option<Ordering> {
-        self.iter()
-            .flat_map(|s| s.iter().copied())
-            .partial_cmp(other.iter().flat_map(|s| s.iter().copied()))
+        self.0.as_ref().partial_cmp(other.0.as_ref())
     }
 }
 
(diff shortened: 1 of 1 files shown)


janik-cloudflare (approved):

partim:
Thank you for the PR!

I’m not sure this is a good idea, though, as this implementation means that TXT record data parsed as a `Txt<_>` and as `UnknownRecordData<_>` can compare differently. This will become relevant once we add a new record data enum that only treats record data types with possibly compressed domain names specially and keeps everything else as plain data – that new type and `AllRecordData<_>` would then potentially compare differently.

One option is to implement a wrapper type, say `TxtContent<_>`, that treats the content as a single string and implements the comparison traits accordingly. Would that work for your use cases?

(As an aside, if we change the `PartialEq` and `Eq` impls, we need to also change `PartialOrd`, `Ord`, and `Hash` accordingly.)

partim:
Erm, I think I’ve gotten this all backwards and you are actually implementing what I am arguing for. Sorry about that.

But wouldn’t it then be easier to just compare the underlying octets?

janik-cloudflare:
Cool cool :) I think comparing the octets is a great idea in general. Looking at `src/rdata/rfc1035/a.rs` that just seems to `#[derive(..., Eq, Hash, Ord, PartialEq, PartialOrd)]`, so perhaps we could do the same here? We'll take a stab at it. Thanks for the feedback @partim!

partim:
Deriving will force trait bounds that will then break the macros that create the rdata enums. I think `src/rdata/rfc1035/null.rs` is a better example of what you need (ie., you should only require `Octs: AsRef<[u8]>` for all those traits and that, sadly, means you have to create manual impls).

dklbreitling:
@partim thanks! We stole the `impl`s from `null.rs`, as per your suggestion.

Also changed `CanonicalOrd`. Happy to change if this is not what you intended. Any feedback is welcome.

src/rdata/rfc1035/txt.rs
```diff
@@ -1035,6 +1016,28 @@ mod test {
...
+                        .append_charstr(
+                            CharStr::from_slice(string.as_bytes()).unwrap(),
+                        )
+                        .unwrap();
+                }
+                builder.finish().unwrap()
+            })
+            .collect::<Vec<_>>();
+
+        assert_ne!(records[0], records[1]);
```
hunts, line 1038: It's better to also include a test case the check two records are equal, i.e. `assert_eq!(...)`.

dklbreitling: Thank you, I included such a test now.

partim (approved):

partim:
Thank you again for the PR and the subsequent changes! This looks good to merge now.

janik-cloudflare:
Great news, super exciting, thanks for the quick merge @partim!

dklbreitling:
Thanks for the quick changes!

PR #375, opened 2024-08-21
XFR response interpreter.

ximon18:
Based on code in the `xfr` branch which should be replaced by this code, in order to split PR https://github.com/NLnetLabs/domain/pull/335 into several smaller PRs.

This PR adds an interpreter for XFR responses which handles the low level XFR protocol details and emits higher level zone updates based on the received data for consumers to deal with without regard for the low level details.

Subsequent PRs will build on this work to:
- Update zones based on the output of `XfrResponseInterpreter` (See https://github.com/NLnetLabs/domain/pull/376)
- Use a streaming capable client with the `XfrResponseInterpreter` in this PR to do XFR client work for the `ZoneMaintainer` in the `xfr` branch.


src/net/client/xfr.rs
```diff
@@ -0,0 +1,1123 @@
...
+
+    /// Process a single AXFR/IXFR response message.
+    ///
+    /// During processing events will be emitted to the registered
+    /// [`XfrEventHandler`] for handling.
+    ///
+    /// Returns Ok(true) if the XFR response was the last in the seqence,
+    /// Ok(false) if more XFR response messages are needed to complete the
+    /// sequence, or Err on error.
+    pub async fn answer(
```
partim, line 57: Perhaps the method should be named `process_answer` or similar? The current name sounds like this method answers something which is very confusing for the client …

src/net/client/xfr.rs
```diff
@@ -0,0 +1,1123 @@
...
+        // line above using `map_err()` and `?` as the Rust compiler complains
+        // about attempting to return `resp` while a reference to it still
+        // exists.
+        let (xfr_type, answer) = match res {
+            Ok(values) => values,
+            Err(err) => return Err(Error::from_check_error(resp, err)),
+        };
+
+        // Walk over all records that are of the type that can exist in a zone
+        // file. TODO: Should this actually walk over all possible record
```
partim, line 75: In response to the TODO question: It probably should indeed process everything – it’s more important that the transferred zone is the same as the original than that it is correct. One option is to use `AllRecordData`, another option (probably better) is to make the type generic over a `ParseAnyRecordData<..>` type.

src/net/client/xfr.rs
```diff
@@ -0,0 +1,1123 @@
...
+            }
+
+            Mode::TransferComplete => {
+                // We already finished processing an XFR response sequence. We
+                // don't expect there to be any more messages to process!.
+                return Err(Error::Malformed);
+            }
+
+            Mode::TransferFailed => {
+                // We had ot terminate processing of the XFR response sequence
```
partim, line 119: Typo: ”We had _to_ terminate …”

src/net/client/xfr.rs
```diff
@@ -0,0 +1,1123 @@
...
+                    }
+                }
+            }
+        }
+    }
+}
+
+//------------ RecordResult ---------------------------------------------------
+
+/// An event emitted by [`XfrResponseProcessor`] during transfer processing.`
```
partim, line 446: There seems to be a stray backtick (?) at the end of the line.

src/net/client/xfr.rs
```diff
@@ -0,0 +1,1123 @@
...
+    /// Handle the given [`XfrEvent`].
+    ///
+    /// Returning an Err will cause transfer processsing to be aborted and the
+    /// error to be returned to the client of [`XfrResponseProcessor`], except in
+    /// the case of [`XfrEvent::ProcessingFailed`] for which the return value of
+    /// this handler will be ignored by [`XfrResponseProcessor`].
+    fn handle_event(
+        &self,
+        evt: XfrEvent<XfrRecord>,
+    ) -> impl std::future::Future<Output = Result<(), Error>> + Send;
```
partim, line 520: I’m not sure we should use impl trait in trait method return types just yet. I think at this time I would prefer if we use an associated type for the future type and provide a blanket impl for `Fn(...)` that uses boxed trait objects.

src/net/client/xfr.rs
```diff
@@ -0,0 +1,1123 @@
...
+            Rtype::AXFR => Ok(XfrType::Axfr),
+            Rtype::IXFR => Ok(XfrType::Ixfr),
+            _ => Err(()),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use core::net::Ipv4Addr;
```
partim, line 618: You should probably use `crate::base::net::IpV4Addr` here for now.

src/net/client/xfr.rs
ximon18: I agree, but for consistency it was named after the existing `ClientSequence::answer()`. I'm not sure about the word "process" in the name or in the name of the type either, need to think about the exact API and naming to offer.

src/net/client/xfr.rs
ximon18: I briefly tried `AllRecordData` but couldn't work out how to use it in this context. How would the `ParseAnyRecordData<..>` idea look, if you can give a bit more explanation of what you have in mind that would be appreciated.

src/net/client/xfr.rs
ximon18: The Rust compiler told me to do it this way ;-) I _think_ I understand your suggestion, will see if I actually do.

ximon18: By the way, if I understand correctly, the blanket Fn impl approach if the same as the one we had previously for `service()` kept causing problems with (a) not being able to implement other impls due to overlapping impls and (b) caused very confusing compiler errors when the compiler thought I was trying in client code to impl the Fn trait but actually I wasn't. So, if that's what you mean, I'm not so sure it's a good idea.

partim: Maybe we can avoid the explicit use of futures by flipping this on its head? The processor could return some sort of iterator over events for each answer given to it.

src/net/client/xfr.rs
partim: I think you only need to swap out the use of `RecordSection::limit_to` with `RecordSection::into_records` and everything else should stay the same. Might be missing something, though.

ximon18:
@partim: One problem witb the inversion to an iterator based alproach is that the iterator must be polled while the events were previously pushed.

In the next PR a Zone updates itself, or is updated, by something receiving the events.

I think I'll still need something for that use case that polls the iterator and sends events because, if the Zone or Zone updater did that, it would need to detect end-of-iterator and receipt of the next message and push the message into the XfrResponseProcessor to update the iterator which means it would need to be near in some way to the XFR client code.

src/base/message.rs
```diff
@@ -439,6 +438,20 @@ impl<Octs: Octets + ?Sized> Message<Octs> {
...
     }
 
+    /// Returns whether this is the answer to some other message.
+    ///
+    /// The method checks whether the ID fields of the headers are the same,
+    /// whether the QR flag is set in this message. It does NOT check whether
+    /// the questions are the same which is useful in case of responses such
+    /// as subsequent RFC 5936 AXFR responses which are not required to copy
+    /// the question from the request to the response.
+    pub fn is_answer_header<Other: Octets + ?Sized>(
```
Philip-NLnetLabs, line 448: I think this is wrong. Low level transports check the ID, no need to check it at other places. In addition this seems only need for zone transfers. We should not put this in Message. It clutters the interface.

src/net/client/xfr.rs
```diff
@@ -0,0 +1,1147 @@
+//! Parsing of AXFR/IXFR response messages for higher level processing.
```
Philip-NLnetLabs, line 1: This doesn't seems to be a client transport so it should go somewhere else. Maybe net::xfr could be a place for XFR functionality.

src/net/client/xfr.rs
```diff
@@ -0,0 +1,1147 @@
...
+///
+/// If the `TransferComplete` event has not been seen it means that the
+/// sequence is incomplete and the next response message in the sequence
+/// should be passed to [`process_next_answer()`] along with the exhausted
+/// iterator. This will populate thr [`XfrEventIterator`] with more records
+/// to parse thereby causing iteration to resume.
+///
+/// The process of producing and consuming iterators continues until the end
+/// of the transfer is detected or a parsing error occurs.
+pub struct XfrResponseProcessor;
```
Philip-NLnetLabs, line 47: This doesn't seems to be object. It doesn't have any state. Maybe just put some functions in a module?

src/net/client/xfr.rs
```diff
@@ -0,0 +1,1147 @@
...
+        req: &Message<Bytes>,
+        resp: &Message<Bytes>,
+    ) -> Result<XfrType, CheckError> {
+        // Check the request.
+        let req_header = req.header();
+        let req_counts = req.header_counts();
+
+        if req.is_error()
+            || req_header.qr()
+            || req_counts.qdcount() != 1
```
Philip-NLnetLabs, line 158: This seems wrong for the 2nd and later messages. The query may be absent.

src/net/client/xfr.rs
```diff
@@ -0,0 +1,1147 @@
...
+
+        let Some(qtype) = req.qtype() else {
+            return Err(CheckError::NotValidXfrResponse);
+        };
+
+        let xfr_type = match qtype {
+            Rtype::AXFR => XfrType::Axfr,
+            Rtype::IXFR => XfrType::Ixfr,
+            _ => return Err(CheckError::NotValidXfrResponse),
+        };
```
Philip-NLnetLabs, line 173: This doesn't belong in this function. It should check the reply not the request. Maybe another function should check the request.

src/net/client/xfr.rs
```diff
@@ -0,0 +1,1147 @@
...
+
+        // https://datatracker.ietf.org/doc/html/rfc1995#section-3
+        // 3. Query Format
+        //   "The IXFR query packet format is the same as that of a normal DNS
+        //    query, but with the query type being IXFR and the authority
+        //    section containing the SOA record of client's version of the
+        //    zone."
+        if matches!(xfr_type, XfrType::Ixfr) && req_counts.nscount() != 1 {
+            return Err(CheckError::NotValidXfrResponse);
+        }
```
Philip-NLnetLabs, line 183: Again checking the request.

src/net/client/xfr.rs
```diff
@@ -0,0 +1,1147 @@
...
+
+        if resp.is_error()
+            || !resp.is_answer_header(req)
+            || resp_header.opcode() != Opcode::QUERY
+            || resp_header.tc()
+            || resp_counts.ancount() == 0
+            || resp_counts.nscount() != 0
+        {
+            return Err(CheckError::NotValidXfrResponse);
+        }
```
Philip-NLnetLabs, line 197: Duplicate checks, is_answer_header also checks that the answer section is not empty.

src/net/client/xfr.rs
```diff
@@ -0,0 +1,1147 @@
...
+        let req_counts = req.header_counts();
+
+        if req.is_error()
+            || req_header.qr()
+            || req_counts.qdcount() != 1
+            || req_counts.ancount() != 0
+            || req_header.opcode() != Opcode::QUERY
+        {
+            return Err(CheckError::NotValidXfrQuery);
+        }
```
Philip-NLnetLabs, line 163: This doesn't belong in this function. It should check the reply not the request. Maybe another function should check the request.

src/net/client/xfr.rs
```diff
@@ -0,0 +1,1147 @@
...
+/// should be passed to [`process_next_answer()`] along with the exhausted
+/// iterator. This will populate thr [`XfrEventIterator`] with more records
+/// to parse thereby causing iteration to resume.
+///
+/// The process of producing and consuming iterators continues until the end
+/// of the transfer is detected or a parsing error occurs.
+pub struct XfrResponseProcessor;
+
+impl XfrResponseProcessor {
+    /// Process a single AXFR/IXFR response message.
```
Philip-NLnetLabs, line 50: Ah, maybe this gets called only for the first message?

src/net/client/xfr.rs
```diff
@@ -0,0 +1,1147 @@
...
+        };
+
+        Ok(iter)
+    }
+
+    /// Process a subsequent XFR response message.
+    ///
+    /// Revives the given iterator with new data based on the given response.
+    pub async fn process_next_answer<'a>(
+        it: &mut XfrEventIterator<'a>,
```
Philip-NLnetLabs, line 104: Very confusing, an object with no state requires the caller to pass an iterator around that has the actual state.

src/net/client/xfr.rs
```diff
@@ -0,0 +1,1147 @@
...
+
+//------------ XfrRecord ------------------------------------------------------
+
+/// The type of record processed by [`XfrResponseProcessor`].
+pub type XfrRecord =
+    Record<ParsedName<Bytes>, AllRecordData<Bytes, ParsedName<Bytes>>>;
+
+//------------ XfrResponseProcessor -------------------------------------------
+
+/// An AXFR/IXFR response processor.
```
Philip-NLnetLabs, line 28: I wonder why this doesn't use a callback. It seems that the caller has to feed the individual response messages into this, then for each message loop over an iterator to do something, moving to the next  response when the iterator is empty.

partim:
My idea for the iterator-based approach was slightly different: First you create a processor by giving it the request. Once you have the first answer message, you call a method that gives you an iterator over the events in that answer. You now poll that iterator until it runs out. Then you ask the processor if it wants another answer and if so, go and receive this message and pass it to the method again to get an iterator. You repeat this until the processor tells you it has seen all the answers.

Thus, the processor holds the permanent state (the request and perhaps some status) and the iterator keeps a mut ref to the processor.

ximon18:
> Thus, the processor holds the permanent state (the request and perhaps some status) and the iterator keeps a mut ref to the processor.

@partim: That would still have the same problem with respect to updating a zone as I mentioned above. I'm finding the original handler design more appealing at this point, @Philip-NLnetLabs also asks why it js not so (it was but I changed it).. Or am I missing something?

Philip-NLnetLabs:
I think an iterator would be fine, if it was an iterator over the entire transfer. A model where the caller has to feed responses and keep polling an iterator feels awkward to me. It's not a blocker for me. If there are good reason for the this iterator approach then just keep it.

ximon18:
Indeed the phased iteration is undesirable. Iterators cannot be async so there's no way to non-blocking wait in the iterator till the next data becomes available. One could wait until all messages have been received then iterate but that will use more memory and delay processing so I am not in favour of that. The sending of events to a  callback didn't have this issue, but needed the callback to be async so that the recipient can do async things, but that influenced the trait interface and @partim wanted some changes, then we both thought of the iterator idea, but now I'm beginning to think the callback approach has advantages.

partim:
The main reason for proposing the iterator approach is avoiding futures in public traits. If that makes your work down the line more complicated, then having them may be a good enough reason.

ximon18:
An async stream might make a lot pf sense as that is in essence what we are dealing with. I can try on Monday seeing what that would look like.

ximon18:
> An async stream might make a lot of sense as that is in essence what we are dealing with. I can try on Monday seeing what that would look like.

I spent this morning trying this but had trouble getting pins and lifetimes and mutability to compile. As I'm not convinced it would actually be a better interface, and as @Philip-NLnetLabs also already asked why not use a callback approach, and in the interest of progressing this work at this point I have now switched back to the callback/event handler approach from earlier in this PR.

src/net/client/xfr.rs
ximon18: This is still implemented as a Future but now as an associated type to work around minimal versions CI error  `"impl Trait only allowed in function and inherent method return types, not in trait method return"`.

src/base/message.rs
ximon18: The reason for this was that the `XfrResponseProcessor` has no direct link to client transports, anything that can provide a `Message` can use it.

src/net/client/xfr.rs
ximon18: Moved to `net::xfr`.

src/net/client/xfr.rs
ximon18: This was an artefact of moving from callback to iterator approach. As the code has now been reverted to the callback approach this issue no longer exists.

src/net/client/xfr.rs
ximon18: The iterator code has been reverted so this comment is outdated. Resolving.

src/net/client/xfr.rs
ximon18: It was callback but was changed to be an iterator. As that change has now been reverted this comment is now outdated. Resolving.

src/net/client/xfr.rs
ximon18: I see your point. The idea was to closely replicate the complete set of checks defined in the XFR RFCs in the XFR related code and not have some checks here and some checks there. I can reduce the checks to just those that are not already done elsewhere.

src/base/message.rs
Philip-NLnetLabs: If somebody wants to apply XFR processing to a random collection of responses why check the ID? It needlessly complicates the interface. In any case, even if you want to check the ID in XfrResponseProcessor, just check it there. Do not add very specialized functions to Message.

ximon18: If someone accidentally processes unrelated messages it would help catch that problem.

I tend to prefer not making assumptions in library code about inputs but I appreciate there is a balance that can be found between defence-in-depth and defence-by-design, assuming that the API is constrained enough that the expected usage is known/guided in advance.

I didn't consider splitting an existing function to be adding very specialized behaviour.

I'll reconsider the code in question  in light of your comments.

src/net/client/xfr.rs
ximon18: Following internal discussions the design has now been changed to a sync iterator-per-response approach. I will resolve this outdated comment.

src/net/client/xfr.rs
ximon18: I'm re-opening this conversation because I have for now changed it back to ``ZoneRecordData` because that's all that the zone tree supports. I didn't want to expand the scope to include zone tree changes at this point, if we decide that `AllRecordData` is what we really want. I have not yet looked at how `ZoneRecordData` and `AllRecordData` actually differ so am not able to make a decision on this yet.

src/base/message.rs
ximon18: I removed the extra method as you requested.

ximon18:
> My idea for the iterator-based approach was slightly different: First you create a processor by giving it the request. Once you have the first answer message, you call a method that gives you an iterator over the events in that answer. You now poll that iterator until it runs out. Then you ask the processor if it wants another answer and if so, go and receive this message and pass it to the method again to get an iterator. You repeat this until the processor tells you it has seen all the answers.
> 
> Thus, the processor holds the permanent state (the request and perhaps some status) and the iterator keeps a mut ref to the processor.

The code has been updated to work almost like this. Rarther than holding a mut ref to the entire processor it holds a mut ref just to the part of the shared state that it needs to update.

src/net/xfr/processing/iterator.rs
```diff
@@ -0,0 +1,81 @@
...
+
+        Ok(Self { state, iter })
+    }
+}
+
+impl<'a, 'b> Iterator for XfrEventIterator<'a, 'b> {
+    type Item = Result<XfrEvent<XfrRecord>, IterationError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.iter.next() {
```
partim, line 66: You can `match self.iter.next()?` here (note the question mark) and save yourself from that slightly silly last match arm.

partim (commented):
One thing I would add is a method like `RecordProcessor::is_complete` that basically just checks if the last event was “end of transfer.” This allows you do make a simple outer loop without having to keep track of that last even yourself:

```
    while !processor.is_complete() {
        let answer = /* read answer from client */;
        for event in processor.process_answer(answer)? {
            /* handle event here */
        }
    }
```

examples/client-transports.rs
```diff
@@ -337,6 +337,8 @@ where
         + Sync
         + 'static,
 {
+    use domain::net::xfr::processing::XfrResponseProcessor;
```
Philip-NLnetLabs, line 340: Maybe this should go into a separate example and limit client-transports to stuff in net::client?

ximon18: Okay. Maybe that has value, not sure as it would still need to use a client as it is about processing a client response. In this case I thought it helped to make the tsig axfr example output more useful.

I could instead make the client-transport tsig example dump out zonefile text format rrs instead of what it currently prints, and leave the use of xfr response processing to the zone maintainer which actually processes the result of transfers and applies them.

Philip-NLnetLabs: Yes, that sounds good

ximon18: I'll just simplify this to be consistent with the rest of this example so that it just debug prints the replies and nothing more.

src/net/client/xfr.rs
ximon18: I'm going to resolve this conversation as in the time available I'm not going to adjust the underlying zone tree to handle more than `ZoneRecordData` and this PR is intended to be used to help feed data into that zone tree, i.e. has to be compatible with it.

src/net/client/xfr.rs
ximon18: The current code still has this block but not for the purpose of checking the request but for determining if it is AXFR or IXFR.

src/net/client/xfr.rs
ximon18: I'm resolving this because (a) `is_answer_header()` no longer exists and (b) a client cannot use `is_answer()` to verify the response because `is_answer()` disallows a valid XFR response because it requires QDCOUNT in the response to match QDCOUNT in the request, but RFC 5936 does not require subsequent AXFR responses to contain the question, so the checking makes sense to do in this code as it can't have been checked the standard way by the caller.

src/net/client/xfr.rs
ximon18: Resolving because the code no longer checks the request.

src/net/client/xfr.rs
ximon18: Resolving as I think this is outdated, I don't think any action was required for this feedback and I no longer know given the updated code if it still has any meaning.

Philip-NLnetLabs (approved):

PR #376, opened 2024-08-21
XFR zone updating.

ximon18:

Note: None of the changes in this PR are breaking as they only affect the unstable zonetree and XFR related parts of `domain`.

The major changes in this PR are the addition of `ZoneUpdater`, `ZoneDiffBuilder` and `ZoneDiff` types.

`ZoneUpdater` takes the `ZoneUpdate`s produced by `XfrZoneUpdateIterator` (added in PR #375, enhanced slightly in this PR) and applies them to a `Zone`, optionally creating a `ZoneDiff` in the process (to be used by a later PR which will serve IXFR out based on created zone diffs).

This PR also makes a number of smaller related changes: (in no particular order)
- Added, improved and fixed various RustDoc texts.

- `ZoneUpdate` related changes:
  - Renamed the inner record type of `ZoneUpdate` from `XfrRecord` to `ParsedRecord` as it is not XFR specific.
  - Added variant `ZoneUpdate::DeleteAllRecords` which is now emitted at the start of an AXFR update by `XfrZoneUpdateIterator`. `ZoneUpdater` uses this to "remove" all records in the zone before adding the new records. ("remove" in quotes because the remove is a marker operation that is non-destructive so that existing readers continue to see the existing zone version as it should be while the new zone version is being created).
  - Removed the superfluous serial number data members associated with some `ZoneUpdate` variants.

- `zonetree::in_memory` changes:
  - Factor out `ZoneApex::prepare_name()` as `zonetree::util::rel_name_rev_iter()` as it is also now used by `ZoneUpdater`.
  - Improve some naming, e.g. there was mix of terms `remove` and `clean`, all references to `clean` have now been renamed to `remove` based names instead.
  - Altered the logic in `Versioned::remove_all()` (formerly `Versioned::clean()`) as it made destructive changes to the zone that would have impacted readers of the current zone version while the new zone version was being created.
  - Adds an `arc_into_inner()` function for Rust <1.70.0 which doesn't have `Arc::into_inner()`. For our purposes I think it's good enough, it was based on the actual Rust 1.70.0 code, and once our MSRV moves up from 1.68.2 to 1.70.0 and beyond we will be able to get rid of this function (and also the newly added dependency on the `rustversion` crate that came with this change).

- Additional trait bounds where needed.

- Removed some unused code.

Note: This PR contains some unit tests but Stelline tests are left to a later PR.

The diff construction and manner in which diffs are referenced feels suboptimal and ugly and simplistic in places and overly complex in others, and none of these changes are intended to be memory or processor efficient at this point. However, this is currently more about getting all of the puzzle pieces for initial XFR support together in an initial form than any of this being perfect.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,11 +61,11 @@ zonefile    = ["bytes", "serde", "std"]
 
 # Unstable features
 unstable-client-transport = ["moka", "net", "tracing"]
-unstable-server-transport = ["arc-swap", "chrono/clock", "libc", "net", "rustversion", "siphasher", "tracing"]
+unstable-server-transport = ["arc-swap", "chrono/clock", "libc", "net", "siphasher", "tracing"]
 unstable-stelline = ["tokio/test-util", "tracing", "tracing-subscriber", "unstable-server-transport", "zonefile"]
 unstable-validator = ["validate", "zonefile", "unstable-client-transport"]
 unstable-xfr = []
-unstable-zonetree = ["futures-util", "parking_lot", "serde", "tokio", "tracing"]
+unstable-zonetree = ["futures-util", "parking_lot", "rustversion", "serde", "tokio", "tracing"]
 
 [dev-dependencies]
 lazy_static        = { version = "1.4.0" }
--- a/src/base/serial.rs
+++ b/src/base/serial.rs
@@ -4,21 +4,19 @@
 //! viewed as the 32 bit modulus of a larger number space. Because of that,
 //! special rules apply when processing these values. This module provides
 //! the type [`Serial`] that implements these rules.
-use core::cmp::Ordering;
-use core::{cmp, fmt, str};
-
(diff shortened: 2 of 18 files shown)


Philip-NLnetLabs (approved):

ximon18:
I think I've have introduced a bug while extracting the code from the xfr branch to this PR as a couple of  tests in the xfr branch are now failing after locally merging. I will investigate, and ideally add some tests here too (though at least it was caught by a Stelline test).

ximon18:
> I think I've have introduced a bug while extracting the code from the xfr branch to this PR as a couple of tests in the xfr branch are now failing after locally merging. I will investigate, and ideally add some tests here too (though at least it was caught by a Stelline test).

There were two bugs, both caught by tests in the `xfr` branch and both now fixed in this branch. There are however still no tests in this branch and ideally there should be unit tests in this PR that ensure the behaviour is as expected.

PR #377, opened 2024-08-26
Service request metadata and byte reservation net client

Philip-NLnetLabs:
Add explicit support to client transports for requests that may result in multiple responses.


ximon18 (approved):

src/net/client/request.rs
```diff
@@ -114,15 +193,26 @@ pub struct RequestMessage<Octs: AsRef<[u8]>> {
 }
 
 impl<Octs: AsRef<[u8]> + Debug + Octets> RequestMessage<Octs> {
-    /// Create a new BMB object.
-    pub fn new(msg: impl Into<Message<Octs>>) -> Self {
+    /// Create a new RequestMessage object.
+    pub fn new(msg: impl Into<Message<Octs>>) -> Result<Self, Error> {
         let msg = msg.into();
+
+        // On UDP, IXFR results in a single responses, so we need to accept it.
```
ximon18, line 200:
```suggestion
        // On UDP, IXFR results in a single response, so we need to accept it.
```

src/net/client/stream.rs
```diff
@@ -748,12 +1099,223 @@ where
...
+                }
+
+                // Any other record, just continue.
+            }
+            XFRState::Done => {
+                // We got a record after we are done. Switch to error state.
+                xfr_state = XFRState::Error;
+                return (false, xfr_state, false);
+            }
+            XFRState::Error => panic!("should not be here"),
```
ximon18, line 1263: Use `unreachable!()` instead?

PR #379, opened 2024-08-30
Add `ZonefileFmt` trait for printing records as zonefiles

tertsdiepraam:


src/base/show.rs
```diff
@@ -0,0 +1,284 @@
...
+impl From<fmt::Error> for Error {
+    fn from(_: fmt::Error) -> Self {
+        Self
+    }
+}
+
+pub type Result = std::result::Result<(), Error>;
+
+/// Show a value as zonefile format
+pub trait Show {
```
Philip-NLnetLabs, line 15: Why not pick a name that has something to do with zonefile format. How would anyone guess that 'Show' means zonefile format? It's not like the name has to be short. For example, ToZonefileFormat would be a lot more desciptive.

partim: The trait creates data in presentation format and I think ‘show’ is a pretty good verb for that. It is also consistent with the related traits in the crate. We have “parse“ and “compose” for dealing with wire format and “scan“ and now “show” for presentation format.

partim: I disagree with your notion that names must be (or, indeed, can be) fully self describing. That sort of thing leads to long, complicated names and people having assumptions about how things work.

These concepts are very fundamental to what this crate is providing, so I think giving them short designators and provide a thorough definition of what they do is a good approach. You look this up in the documentation and then remember that “show” means printing in presentation format.

Philip-NLnetLabs: Just a lot of friction for people who are not an expert in using the domain crate.  Inventing terminology that is only used in one piece of software tends to make people want to avoid that software.

The whole parse, compose, and scan are bad enough. Time to move to something more sensible.

Obviously, if a name is used a lot, then it needs to be short, even if there is some initial confusion. But show would be rare enough for most people that it will just feel awkward.

Maybe we only want to write software for people who are experts in using the domain crate. In that case using random names is fine.

Philip-NLnetLabs: It has taken until now to discover that we don't actually have a way to write zonefile format and that using Display for that has issues.

So it seems that this feature may be fundamental, but is not commonly used. Features that are rarely used benefit from longer names that are more descriptive. The whole point of this feature is to generate zonefile format. So the name can include that without being overly specific.

This feature is introduced because the intent is to not show the output (to a user). We already have Display for showing output. This feature is a specific text-based encoding that is meant to be used by other software. Obviously, expert users can look at text-based encodings. But that is not the main purpose. Some fields that could have been mnemonic have to be output as numbers because that is what zonefile format requires. Again a good hint that the purpose is not to show that output to a user.

src/base/iana/macros.rs
```diff
@@ -201,7 +209,14 @@ macro_rules! int_enum_str_decimal {
 
         impl core::fmt::Display for $ianatype {
             fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
-                write!(f, "{}", self.to_int())
+                write!(f, "{}", self.to_int())?;
+
+                if let Some(m) = self.to_mnemonic_str() {
+                    write!(f, "(")?;
+                    f.write_str(m)?;
+                    write!(f, ")")?;
```
bal-e, line 217: Use `write!(f, "({m})")?` instead.

src/base/name/traits.rs
```diff
@@ -343,6 +343,35 @@ pub trait ToName: ToLabelIter {
...
+impl<'a, T> fmt::Display for DisplayWithDot<'a, T>
+where
+    T: ToLabelIter + ?Sized,
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut labels = self.0.iter_labels();
+        let first = match labels.next() {
+            Some(first) => first,
+            None => unreachable!("at least 1 label must be present"),
+        };
```
bal-e, line 363: Use `.expect()`?

src/base/record.rs
```diff
@@ -1479,6 +1496,51 @@ impl Ttl {
...
+                    (minutes as u16, "minutes"),
+                    (seconds as u16, "seconds"),
+                ] {
+                    if n == 0 {
+                        continue;
+                    }
+                    if first {
+                        write!(f, " ")?;
+                    }
+                    write!(f, "{n}{unit}")?;
```
bal-e, line 1528: To be clear, the format here is `3weeks 2hours 1minutes`?  That doesn't seem especially nice, unless it's for use in a particular format.  Either way, some documentation should be provided to show what the result will look like.

src/base/zonefile_fmt.rs
```diff
@@ -0,0 +1,346 @@
...
+        self.writer.write_str(x)
+    }
+}
+
+/// A more structured wrapper around a [`PresentationWriter`]
+///
+/// Writing comments is not allowed with this type because comments can only
+/// appear when a token is surrounded by parentheses.
+pub struct Presenter<'a> {
+    writer: &'a mut (dyn PresentationWriter + 'a),
```
bal-e, line 205: This shouldn't be a trait object.  `std::fmt` uses this paradigm internally and they've suffered from a lot of performance issues because of it.  Based on the uses I've seen in this crate, I think `Presenter` can be generic over the underlying writer type, or the functions it provides can be moved to `PresentationWriter` instead with default impls.  Changing this interface later will be a lot harder, and I think we should get ahead of this while we can.

src/rdata/dnssec.rs
```diff
@@ -431,6 +432,34 @@ impl<Octs: AsRef<[u8]>> fmt::Debug for Dnskey<Octs> {
...
+
+        p.block(|p| {
+            p.write_token(self.flags)?;
+            p.write_comment(
+                format_args!(
+                    "flags:{}{}{}{}",
+                    if revoked { " revoked" } else {""},
+                    if sep { " sep" } else {""},
+                    if zone_key { " zone_key" } else {""},
+                    if self.flags == 0 { " <none>" } else { "" },
```
bal-e, line 451: Nit: inconsistent formatting

bal-e (changes requested):

src/base/name/traits.rs
tertsdiepraam: You'd be right if this was a `panic!`, but `unreachable!` has a slightly different meaning.

src/base/opt/mod.rs
```diff
@@ -358,6 +359,15 @@ impl<Octs: AsRef<[u8]> + ?Sized> fmt::Debug for Opt<Octs> {
     }
 }
 
+//--- Show
+
+impl<Octs: AsRef<[u8]> + ?Sized> Show for Opt<Octs> {
+    fn show(&self, p: &mut Presenter) -> show::Result {
+        // XXX TODO Print this properly.
+        p.write_token("OPT ...")
```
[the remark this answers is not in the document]
tertsdiepraam: Yep, was already incomplete in the `Display` impl and seems like a can of worms to big to open here.

src/base/iana/macros.rs
```diff
@@ -52,6 +52,14 @@ macro_rules! int_enum {
             /// is hidden in a `Int` variant.
             #[must_use]
             pub const fn to_mnemonic(self) -> Option<&'static [u8]> {
+                match self.to_mnemonic_str() {
+                    Some(m) => Some(m.as_bytes()),
+                    None => None,
```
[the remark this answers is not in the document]
tertsdiepraam: That's not `const` unfortunately.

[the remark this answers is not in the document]
tertsdiepraam: (keyword generics when?)

src/base/zonefile_fmt.rs
tertsdiepraam: Fiiiiiiiiiiine, but only because Philip cared about performance too :smile:

src/base/iana/macros.rs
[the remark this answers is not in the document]
bal-e: Ooh, that makes sense, alright!

src/base/opt/mod.rs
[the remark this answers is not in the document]
bal-e: Alright, no worries.

src/base/name/traits.rs
bal-e: Hm, that's fair, I suppose.  I wish we had a more concise syntax for this.

src/base/record.rs
bal-e: I'd argue that we should have a separate method if we want compatibility with something specific.  You could name this `dig_display()` or something if that's exactly what you want to use it for.

tertsdiepraam: Remember that this is for a comment in the zonefile format (hence also `pub(crate)`)

src/base/zonefile_fmt.rs
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    fn end_block(&mut self) -> Result {
+        self.block_indent = None;
+        self.fmt_token(format_args!(")"))
+    }
+
+    fn fmt_comment(&mut self, args: fmt::Arguments<'_>) -> Result {
+        if self.block_indent.is_some() {
+            self.writer.write_fmt(format_args!("\t; {}", args))?;
```
bal-e, line 166: You can just call `write!` here, right?

bal-e (commented):
Looks really good, @tertsdiepraam!

src/rdata/svcb/rdata.rs
```diff
@@ -446,7 +445,7 @@ where
     Name: fmt::Display,
 {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{} {} {}", self.priority, self.target, self.params)
+        write!(f, "{} {}. {}", self.priority, self.target, self.params)
```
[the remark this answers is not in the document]
tertsdiepraam: I used that in the `ZonefileFmt` impl, but not here because it requires `Name: ToName` which would make `ToName` required *everywhere*.

src/base/zonefile_fmt.rs
tertsdiepraam: Yep :)

src/rdata/svcb/rdata.rs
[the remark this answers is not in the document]
bal-e: Ah, right.  Makes sense.

src/base/zonefile_fmt.rs
```diff
@@ -0,0 +1,337 @@
...
+}
+
+/// The simplest possible zonefile writer
+///
+/// This writer does not do any alignment, comments and squeezes each record
+/// onto a single line.
+struct SimpleWriter<'a> {
+    first: bool,
+    writer: &'a mut (dyn fmt::Write + 'a),
+}
```
bal-e, line 84: This could be defined as
```rust
struct SimpleWriter<W> {
    first: bool,
    writer: W,
}
```
The primary problem is that `fmt::Formatter` already contains a trait object, and using it in this struct will result in two layers of vtables.

bal-e (approved):

PR #380, opened 2024-09-03
Add TSIG middleware.

ximon18:

Notable changes:
- Add TSIG response signing middleware.
- Add TSIG Stelline test recipe.
- Update Stelline server integration tests to use the new TSIG middleware.
- Remove unnecessary Unpin bounds in `net::server::message::Request`.
- Tsig module changes:
  - Derive `Clone` for `Key`.
  - Added `Key::compose_len()` for determining how many response bytes to reserve.
  - Added `From<ServerTransaction<K>> for ServerSequence<K>`.
  - Various minor RustDoc improvements.
  - Made `tsig::ServerError::unsigned()` public.
-  Make `Time48` obey mock (predictable and controllable) time so that TSIG signing uses mock time during Stelline tests.
-  Don't set the AA flag on test service responses as (a) actual zone serving doesn't do this yet and this still needs fixing, and (b) it violates the expectations of the TSIG Stelline test that verifies at a byte level the TSIG response signature.

This PR will also benefit from merging PR #334 and PR #390.


src/net/server/middleware/tsig.rs
```diff
@@ -0,0 +1,460 @@
...
+//------------ TsigMiddlewareSvc ----------------------------------------------
+
+/// TSIG message authentication middlware.
+///
+/// This middleware service validates TSIG signatures on incoming requests, if
+/// any, and adds TSIG signatures to responses to signed requests.
+///
+/// | RFC    | Status  |
+/// |--------|---------|
+/// | [8945] | TBD     |
```
Philip-NLnetLabs, line 40: Why does it say TBD here?

Philip-NLnetLabs: Maybe we just comment it out? It doesn't seem to add anything to the documentation at the moment. Or just change TBD to something like Implemented. What else do we want to do with TSIG?

ximon18: If a quick review of RFC 8945 shows that we've implemented it then indeed we can say it is implemented. This table was really just a reminder to self to check what is implemented and note any omissions. It doesn't have to be in this format. If we implement all of 8945 from a server perspective then we can just say in the descriptive text that it is RFC 8945 compliant and get rid of the table.

Philip-NLnetLabs: I think it would be good to do that before merging. Then we don't have to come back to this. I don't think there is much to TSIG processing.

ximon18: Agreed.

src/net/server/middleware/tsig.rs
```diff
@@ -0,0 +1,496 @@
...
+            }
+        }
+    }
+
+    fn get_relevant_question(
+        msg: &Message<RequestOctets>,
+    ) -> Option<Question<ParsedName<RequestOctets::Range<'_>>>> {
+        if Opcode::QUERY == msg.header().opcode() && !msg.header().qr() {
+            if let Some(q) = msg.first_question() {
+                if matches!(q.qtype(), Rtype::SOA | Rtype::AXFR | Rtype::IXFR)
```
ximon18, line 320: While not a requirement of RFC 8945, one could argue that this violates the intent of the RFC because the RFC refers to DNS UPDATE which uses the UPDATE opcode, but this requires the QUERY opcode. The thinking here was that we are adding TSIG support to authenticate zone transfers, which is the other use case described by RFC 8945. Is this too strict? If this test fails the middleware will allow the request to proceed, so this is defining the set of request types which will be TSIG validated. Perhaps we should allow all requests? (but not checking opcode could be problematic, perhaps some opcodes change the message format in ways incompatible with TSIG?). @partim, @Philip-NLnetLabs: What do you think?

src/net/server/middleware/tsig.rs
```diff
@@ -0,0 +1,496 @@
...
+                    let num_bytes_to_reserve = tsig.key().compose_len();
+                    new_req.reserve_bytes(num_bytes_to_reserve);
+
+                    return ControlFlow::Continue(Some((
+                        new_req,
+                        TsigSigner::Transaction(tsig),
+                    )));
+                }
+
+                Err(err) => {
```
ximon18, line 149: This is wrong. `build_message()` should be called on the received error.

src/net/server/middleware/tsig.rs
Philip-NLnetLabs: I think that for a middleware TSIG layer it is best to reject what we cannot support at the moment. So if get_relevant_question fails then don't proceed.

Philip-NLnetLabs: Maybe a FORMERROR is most clear if we get a TSIG record with something the TSIG middleware doesn't support.

ximon18:
For RFC 8945 compliance we also need https://github.com/NLnetLabs/domain/pull/334.

src/net/server/middleware/tsig.rs
ximon18: I pushed a fix for this. @tertsdiepraam, as you worked in this area recently on PR #334 you might also want to take a look.

Philip-NLnetLabs:
That needs to listed as a limitation. Certainly for UPDATE that is something we need.

ximon18:
> That needs to listed as a limitation. Certainly for UPDATE that is something we need.

Done.

src/net/server/middleware/tsig.rs
ximon18: I think it depends what you mean by "cannot support".

Currently if a TSIG signed request is received by the TSIG middleware but is an opcode or qtype that is not considered "relevant" then the TSIG middleware will ignore it and pass it on unchanged to the next layer, either another middleware service or the application service. The TSIG RR would be left intact in the additional section. 

Middleware or application service layers that really need to know whether or not the request was considered validly signed could check the `Option<Key>` passed out by the TSIG middleware service to check (a) is it validly signed (yes if `Some`, unsigned if `None`) and (b) if signed, was it signed with the required key?

"Irrelevant" queries have to be passed on unchanged otherwise requests of other types (e.g. a query for an AAAA record) would not propagate through the service layers even not signed.

My question was actually about something else: should we check for and validate TSIG signatures on more/all types of request rather than just the limited set it checks for now?

ximon18 (commented):
> I think that for a middleware TSIG layer it is best to reject what we cannot support at the moment. So if get_relevant_question fails then don't proceed. 

After some internal discussion it seems to me that there is no reason to restrict the opcode or query type on which TSIG verification is performed. If the request has a TSIG RR and the server stack includes TSIG middleware then that middleware should verify the TSIG signature and also sign any corresponding response. So I'm going to remove this restriction entirely.

src/net/server/middleware/tsig.rs
Philip-NLnetLabs: We can't do update at the moment. So I'd say supporting query and notify should be enough.

ximon18: My comment got posted below instead of here:

> After some internal discussion it seems to me that there is no reason to restrict the opcode or query type on which TSIG verification is performed. If the request has a TSIG RR and the server stack includes TSIG middleware then that middleware should verify the TSIG signature and also sign any corresponding response. So I'm going to remove this restriction entirely.

Regarding your response, we don't do UPDATE but we can verify TSIG on UPDATE requests as it is no different than verifying TSIG on any other request. As stated above I intend to remove the check entirely, let library users decide if they want / need more restrictions, and let it be used wherever it adds value, which is also how NSD works.

src/net/server/middleware/tsig.rs
ximon18: Hmm. I couldn't find the fix I thought I pushed for this so I just pushed it now.

Philip-NLnetLabs (approved):

src/net/server/middleware/tsig.rs
```diff
@@ -0,0 +1,506 @@
...
+                call_res.feedback(),
+                Some(ServiceFeedback::BeginTransaction)
+            ) {
+                // Does it need converting from the variant that supports
+                // single messages only (ServerTransaction) to the variant
+                // that supports signing multiple messages (ServerSequence)?
+                // Note: confusingly BeginTransaction and ServerTransaction
+                // use the term "transaction" to mean completely the opppsite
+                // of each other. With BeginTransaction we mean that the
+                // caller should instead a sequence of response messages
```
tertsdiepraam, line 352: There's a typo here (not sure what it should be)

src/net/server/middleware/tsig.rs
```diff
@@ -0,0 +1,506 @@
+//! RFC 8495 TSIG message authentication middleware.
+//!
+//! This module provides a TSIG request validation and response signing
+//! middleware service. The underlying TSIG RR processing is implemented using
+//! the [`rdata::tsig`][crate::rdata::tsig] module.
+//!
+//! Signed requests thta fail signature verification will be rejected.
```
tertsdiepraam, line 7:
```suggestion
//! Signed requests that fail signature verification will be rejected.
```

src/net/server/middleware/tsig.rs
```diff
@@ -0,0 +1,506 @@
...
+//! will be signed using the same key as the request.
+//!
+//! # Determining the key that a request was signed with
+//!
+//! The key that signed a request is output by this middleware via the request
+//! metadata in the form [`Option<KS::Key>`], where `KS` denotes the type of
+//! [`KeyStore`] that was used to construct this middleware. Upstream services
+//! can choose to ignore the metadata by being generic over any kind of
+//! metadata, or may offer a [`Service`] impl that specifically accepts the
+//! [`Option<KS::Key>`] metadata type, enabling the upstream service to use
```
tertsdiepraam, line 22:
```suggestion
//! `Option<KS::Key>` metadata type, enabling the upstream service to use
```
It's not a valid link because of `KS`

src/net/server/middleware/tsig.rs
```diff
@@ -0,0 +1,506 @@
...
+
+//------------ TsigMiddlewareSvc ----------------------------------------------
+
+/// RFC 8495 TSIG message authentication middleware.
+///
+/// This middleware service validates TSIG signatures on incoming requests, if
+/// any, and adds TSIG signatures to responses to signed requests.
+///
+/// Upstream services can detect whether a request is signed and with which
+/// key by consuming the [`Option<KS::Key>`] metadata output by this service.
```
tertsdiepraam, line 67:
```suggestion
/// key by consuming the `Option<KS::Key>` metadata output by this service.
```

src/net/server/middleware/tsig.rs
```diff
@@ -0,0 +1,506 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    fn map_stream_item(
+        request: Request<RequestOctets, ()>,
+        stream_item: ServiceResult<NextSvc::Target>,
+        pp_config: &mut PostprocessingConfig<KS::Key>,
```
tertsdiepraam, line 339: Not sure how much of this is based on previous convention, but a mutable config is a bit weird to me. It also has some strange implications for other bits where theres now `&mut ()` and `&mut bool`. I wonder if there's some other way to implement that kind of caching behaviour that you're doing here.

src/net/server/middleware/tsig.rs
ximon18: s/opppsite/opposite/ - I'll fix it, thanks!

src/net/server/middleware/tsig.rs
ximon18: It was just the fall out of realizing that `Arc<Mutex<..>>` wasn't need to ensure mutability and that mutability is only there to allow an `Option` member in a struct to be `take()`n. It's not about caching (or at least not what I think of as caching, not sure what you had in mind when using that term). This has changed several times already, and there's some talk of maybe revisiting the streaming side of the `Service` interface in the future to split out the common single request -> response case from the single request -> multiple/streaming response case, and this post-processing map stuff is all about the latter case. So long answer short, I don't have an immediate idea for how to do this differently and this stuff is all behind an "unstable" feature gate as it's a design in-progress so we may well come back and change this.

tertsdiepraam: Maybe caching was too ambiguous, I meant the "store result for future invocations" part, which is also why this is mutable right?

ximon18: It was called config as it was originally configuration to drive the post-processing phase, but it became mutable because that config for TSIG includes a signer which may need replacing with a different one if the response is a stream not a single response. There was some discussion about maybe reworking the underlying TSIG module code to present a single signing interface for single and multi-response but that is out of scope for now. If that already existed the mutation here wouldn't happen. For now I can rename config  to state as that's what it effectively became instead of the original intent.

tertsdiepraam: If it's gonna change later I'm not too worried about it at all. Thanks for the response!

tertsdiepraam (approved):

src/net/server/middleware/tsig.rs
ximon18: Actually, I'm getting mixed up. That is a thing (replacing the signer) but in this case it's just because the signer needs to be consumed so taken out of the struct, but I can't consume the entire struct.

ximon18: Maybe it should really have two parts: immutable config, and mutable state. But I don't really want to get into refining this further right now, I think it's better to let this evolve as we add more middleware that needs to post-process, or we revisit this entirely as we go as mentioned.

src/net/server/middleware/tsig.rs
ximon18: @tertsdiepraam: Can I resolve this conversation?

PR #382, opened 2024-09-04
Add RFC 1996 NOTIFY support.

ximon18:

This PR adds RFC 1996 "slave" side minimalistic NOTIFY request support, delegating the actual work to a caller supplied helper that implements a new `Notifiable` trait.


src/net/server/middleware/mod.rs
```diff
@@ -20,11 +20,11 @@
 //! Currently the following middleware are available:
 //!
 //! [`Service`]: crate::net::server::service::Service
-
 #[cfg(feature = "siphasher")]
 pub mod cookies;
 pub mod edns;
 pub mod mandatory;
+pub mod notify;
```
Philip-NLnetLabs, line 27: Implementing this as a middleware feels wrong. But I don't have a better solution at the moment.

ximon18: Can you say a bit more about why?

To my thinking, receiving a certain kind of request and handling it so that an application service does not have to is what middleware is for. This new middleware understands the NOTIFY request format, extracts the useful data from it and delegates further handling to a supplied helper, and knows how to compose the standards compliant response that should be sent. A later PR then supplies one possible helper implementation, namely the `ZoneMaintaner` that does RFC 1034 zone maintenance including using a received NOTIFY as a trigger to fetch updates for a zone that it maintains.

I would not want the entire ZoneMaintainer as middleware, it does not just listen to and respond to requests but also has background timers and logic and initiates its own outbound requests. And I prefer to avoid adding the (not so great but still additional) complexity of handling NOTIFY to the application service in order to then signal the ZoneMaintainer to act.

So, I'm not opposed to your objection, not at all, but to me at least this middleware fits nicely into the bigger picture and I'm unsure what to do with your feedback at this point. Would you perhaps prefer an entirely different approach than the middleware layered service etc that these PRs are building up?

Philip-NLnetLabs: You are right about middleware in general. However the middleware in src/net/server tends to operate on aspects of a message but ultimately pass the message to a service. Here the notify code passes the message on if it is not a notify message, operating a bit like a filter in pipeline.

ximon18: Ah I see what you mean, thx for clearing that up, I think it's a good point but it isn't something that, to me at least, is a problem. Or am I missing a reason why this is actually a problem?

In my vision middleware is not limited in what it can do. It can inspect, alter, or entirely handle and respond to incoming queries and/or upstream replies. It's just layers in front of the application service. NOTIFY for me is like the XFR middleware: from the application service perspective the NOTIFY and AXFR/IXFR messages are entirely handled below the application service.

ximon18: Okay. Why would each zone have its own middleware? I wouldn't expect that. The NOTIFY middleware is supposed to delegate the handling to something else that is actually of the zones under management. in the `xfr` branch you can see this is done by the `ZoneMaintainer`, which sounds like what you are describing but without the "router". There's nothing stopping someone adding some kind of request routing middleware, but what problem does it solve here? Perhaps we should talk about this IRL instead of here?

Philip-NLnetLabs: So I think you should do the same thing as 'ZoneMaintainer' with opcode. It is just not a good demultiplexing strategy to have a list of objects that all try to see if the request if for that object and pass on the request if not. Indeed, let's see if we can talk IRL

ximon18: @Philip-NLnetLabs and I discussed this and @Philip-NLnetLabs is proposing an optimization by OPCODE via a "routing" middleware that forks incoming requests to separate middleware "chains" based on OPCODE, so that not every QUERY has to be checked to see if it is a NOTIFY (for example). We agreed that I would add a note for now in the NOTIFY code that we might want such a router later, though nothing stops a user from adding it themselves in theory. I think I will also add a quick note to the net::server::middleware RustDoc pointing out that while the demonstrated chains and provided middleware services are all intended to be used in linear fashion, one could create such a forking router (or even choose to use certain middleware only on certain ports via separate server instances with their own middleware chains).

Philip-NLnetLabs (approved):

PR #383, opened 2024-09-04
Add RR batching functionality.

ximon18:

This PR implements a "batcher" of resource records which can fill up an in-progress response until it is full then invoke a callback to send it before starting to fill the next response. This is useful for large multi-response answers to queries such as AXFR and IXFR.

Will be followed by an XFR specific batcher implementation that is used by the XFR middleware in the `xfr` branch.


src/net/server/batcher.rs
```diff
@@ -0,0 +1,473 @@
...
+        _state: &T,
+        _answer: &AnswerBuilder<StreamTarget<Target>>,
+    ) -> bool {
+        false
+    }
+
+    /// Do something with the completed message.
+    #[allow(clippy::result_unit_err)]
+    fn batch_ready(
+        _state: &T,
```
Philip-NLnetLabs, line 145: Maybe these parameters do not need an underscore?

ximon18: Fixed.

Philip-NLnetLabs (approved):

PR #384, opened 2024-09-05
Add XFR middleware.

ximon18:
Extraction of `XfrMiddlewareSvc` from the `xfr` branch and conversion such that access control checks are no longer hard-coded into the middleware nor dependent on policy data being stored in a specific form, and renames `ZoneDiff` to `InMemoryZoneDiff` and adds a `ZoneDiff` trait to remove the fixed use of in-memory diffs by `XfrMiddlewareSvc`.

#### Notes about the changes to `src/zonetree/in_memory/read.rs`:

Changes in `read.rs` are just cleaning up:
- In `query_node_here_and_below()` it now only invokes the walker if actually enabled, though nothing was actually broken by not checking for this (as when not enabled it does nothing), this change just makes the code more consistent with other code in `read.rs`.
- In `query_node_here_but_not_below()` calls to the walker have been removed, but only because they never get invoked anyway, as the only caller of this fn only calls if it not walking.
- Some fixes/tweaks to code comments.


src/net/server/middleware/xfr/responder.rs
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        // SAFETY: msg.sole_question() was already checked in
+        // get_relevant_question().
+        let qclass = self.msg.sole_question().unwrap().qclass();
+
+        // Note: NSD apparently uses name compresson on AXFR responses
+        // because AXFR responses they typically contain lots of
+        // alphabetically ordered duplicate names which compress well. NSD
+        // limits AXFR responses to 16,383 RRs because DNS name
```
Philip-NLnetLabs, line 77: Did you mean octets here?

ximon18: Yes. Thanks for spotting that. Replaced with bytes (as that is consistent with what is said further down in the same comment).

Philip-NLnetLabs (approved):

PR #386, opened 2024-09-11
Link to definition of DNS

ximon18:
Looking at the RustDocs for this crate it struck me that it refers immediately to DNS without defining or linking to any definition of what DNS is. For people looking for a DNS crate this will be self evident, for others who stumble across the crate maybe less so. Perhaps it is overkill, but I wondered if it might be good to actually link to a definition of what DNS is, and chose https://www.rfc-editor.org/rfc/rfc9499.html for that.


maertsen:
I like https://www.sciencedirect.com/science/article/pii/S1574013722000132 (note, find better url) for a DNS tutorial. I’m guessing those who don’t know DNS might be better served by a tutorial than by an RFC?

ximon18:
> I like https://www.sciencedirect.com/science/article/pii/S1574013722000132 (note, find better url) for a DNS tutorial. I’m guessing those who don’t know DNS might be better served by a tutorial than by an RFC?

Nice idea, though I think it's something separate than my intent that maybe warrants a sentence of its own, e.g. "If you want to learn more about DNS you may find the following article helpful?".

maertsen:
> > I like https://www.sciencedirect.com/science/article/pii/S1574013722000132 (note, find better url) for a DNS tutorial. I’m guessing those who don’t know DNS might be better served by a tutorial than by an RFC?
> 
> Nice idea, though I think it's something separate than my intent that maybe warrants a sentence of its own, e.g. "If you want to learn more about DNS you may find the following article helpful?".

👍

> **Update:** After looking at that article, it indeed starts with a definition of what DNS is, so maybe it would be a good substitute. 

Nah, the approach above makes sense. Also, it was just a thought.

> However, one reason I like the RFC link is that that is what the DNS community officially thinks DNS is, not what some article on the internet says.

heh, it’s not “some article on the internet”. It’s a peer reviewed academic paper in a good venue.

ximon18:
> heh, it’s not “some article on the internet”. It’s a peer reviewed academic paper in a good venue.

Sorry, I didn't examine what it was, it wasn't my intent to be dismissive or disrespectful. Even so, that RFC exists to be the authoritative definition of what DNS is so to my mind is the primary source to reference. Contrary to that however I suppose it's fair to say that if you don't know what DNS stands for, an RFC is not going to be the easiest starting point. One other concern that may be invalid in the academic context, but I lack experience there, is whether or not the article will remain available at that link. I would expect that to be true for an approved RFC.

partim (approved):

PR #390, opened 2024-09-13
Network servers should not ignore `ServiceError`.

ximon18:
Description to come...

### DIFF
--- a/src/net/server/connection.rs
+++ b/src/net/server/connection.rs
@@ -8,7 +8,6 @@ use std::net::SocketAddr;
 use std::sync::Arc;
 
 use arc_swap::ArcSwap;
-use futures::StreamExt;
 use octseq::Octets;
 use tokio::io::{
     AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt, ReadHalf, WriteHalf,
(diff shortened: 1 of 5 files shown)


src/net/server/dgram.rs
```diff
@@ -502,74 +509,28 @@ where
...
-                                            warn!(%addr, "Failed to send response: {err}");
-                                        }
-
-                                        metrics.dec_num_pending_writes();
-                                        metrics.inc_num_sent_responses();
-                                    }
-                                }
-                            }
+                    match Message::from_octets(buf) {
+                        Err(err) => {
```
Philip-NLnetLabs, line 513: It seems that in this case no error gets returned to the client

src/net/server/connection.rs
```diff
@@ -682,112 +690,24 @@ where
...
-                                                    error!("Unable to queue message for sending: queue is full.");
-                                                    break;
-                                                }
-                                            }
-                                        }
-                                    }
-                                }
-                            }
-                            trace!("Finished processing service call results for request id {request_id}");
+                            dispatcher.dispatch(request, service, ()).await
```
Philip-NLnetLabs, line 710: I get the impression that if Message::from_octets fails the client doesn't get a reply

src/net/server/invoker.rs
```diff
@@ -0,0 +1,188 @@
+/// Common service invoking logic for network servers.
```
Philip-NLnetLabs, line 1: Should the check for QR be moved into this?

src/net/server/dgram.rs
ximon18: We don't have a DNS request from which to extract a query ID to copy into the error response. So we could generate a FORMERR response and send it, but with what ID?

ximon18: Note: When merged with `main` there is an additional `Ok(msg) if msg.header().qr() => ` match stanza which also doesn't send a reply, but that DOES know the ID of the request and so should respond.

Philip-NLnetLabs (approved):

PR #392, opened 2024-09-16
Add NXNAME RR type and invalid query type EDE

mworsley-cloudflare:
Adding NXNAME resource record type and the EDE code to reject invalid queries for it
[EDE 30](https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#:~:text=30-,Invalid%20Query%20Type,-%5Bdraft%2Dietf) and [NXNAME](https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#:~:text=NXDOMAIN%20indicator%20for%20Compact%20Denial%20of%20Existence) are both IANA reserved

### DIFF
--- a/src/base/iana/exterr.rs
+++ b/src/base/iana/exterr.rs
@@ -149,6 +149,9 @@ int_enum! {
     /// is otherwise configured to support. Examples of this include
     /// its most recent zone being too old or having expired.
     (INVALID_DATA => 24, b"Invalid Data")
+
+    /// The requested resource record type should not appear in a query.
+    (INVALID_QUERY_TYPE => 30, b"Invalid Query Type")
 }
 
 /// Start of the private range for EDE codes.
--- a/src/base/iana/rtype.rs
+++ b/src/base/iana/rtype.rs
@@ -364,6 +364,11 @@ int_enum! {
     /// See RFC 7043.
     (EUI64 => 109, b"EUI64")
 
+    /// NXNAME.
+    ///
+    /// IANA-Reserved.
+    (NXNAME => 128, b"NXNAME")
+
     /// Transaction key.
     ///
     /// See RFC 2930.


Philip-NLnetLabs:
Did something go wrong? IANA class it 'NXNAME', the PR seems to have NXDOMAIN?

mworsley-cloudflare:
> Did something go wrong? IANA class it 'NXNAME', the PR seems to have NXDOMAIN?

You're 100% right, it should be NXNAME to match the IANA reserved

src/base/iana/exterr.rs
```diff
@@ -149,6 +149,9 @@ int_enum! {
     /// is otherwise configured to support. Examples of this include
     /// its most recent zone being too old or having expired.
     (INVALID_DATA => 24, b"Invalid Data")
+
+    /// The requested resource record should not appear in query
```
SebastiaanYN, line 153: tiniest nit: comment should end with a `.`

SebastiaanYN (approved):

src/base/iana/exterr.rs
mworsley-cloudflare: Haha no worries, I've added it

mworsley-cloudflare:
Hi there, is there anything else the team would like out of this PR?

src/base/iana/exterr.rs
```diff
@@ -149,6 +149,9 @@ int_enum! {
     /// is otherwise configured to support. Examples of this include
     /// its most recent zone being too old or having expired.
     (INVALID_DATA => 24, b"Invalid Data")
+
+    /// The requested resource record should not appear in query.
```
ximon18, line 153: Just a couple of nits about the comment:

- Replace "in query" with "in a query" or "in queries".

- The RFC refers to "any RR _type_" while the comment refers to "resource record" (without type). I think "resource record type" would be more correct.

ximon18 (changes requested):

ximon18:
I notice that https://datatracker.ietf.org/doc/draft-ietf-dnsop-compact-denial-of-existence/ has not yet finished moving through the IETF process nor has the RFC been updated yet to reflect the new IANA EDE code point. Given that the IANA code point has now been allocated (at least it appears at https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#extended-dns-error-codes) I assume it is low risk to merge this PR while the IETF process is ongoing.

src/base/iana/exterr.rs
mworsley-cloudflare: They definitely add clarity, I've made the changes now

mworsley-cloudflare:
> I notice that https://datatracker.ietf.org/doc/draft-ietf-dnsop-compact-denial-of-existence/ has not yet finished moving through the IETF process nor has the RFC been updated yet to reflect the new IANA EDE code point. Given that the IANA code point has now been allocated (at least it appears at https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#extended-dns-error-codes) I assume it is low risk to merge this PR while the IETF process is ongoing.

This is our assumption as well given that they've been allocated, in the small chance that anything changes I'll be back

ximon18 (approved):

Chreo:
https://datatracker.ietf.org/doc/draft-ietf-dnsop-compact-denial-of-existence/ is in the final stages for RFC publication and while that may still result in changes to the final document, the allocation of the NXNAME type and the new EDE code is not going to be changed so is safe to import. IANA does not have a habit of changing allocated types meaning that worst case the allocated code point is burned and considered deprecated in the future.
Anything marked as allocated on https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml can be considered stable and not subject to rash changes.
It would be very useful to get this merged.

PR #388, opened 2024-09-12
Overhaul parsing from the presentation format

bal-e:

PR #394, opened 2024-09-16
Rewrite `NameBuilder`

bal-e:
This is primarily a performance enhancement.  The idea is to require the storage backing `NameBuilder` to always provide a fixed 256-byte array, rather than performing any dynamic allocations.  Since `NameBuilder`s are transient objects, allocating a few more bytes won't hurt, especially if they can be put on the stack.

I initially intended to make `NameBuilder` just take a reference to a `[u8; 256]` array somewhere in the caller's code.  However, this would make it harder to store `NameBuilder` (because you also need to store the backing buffer somewhere, and you can quickly run into self-referential lifetime issues).  Instead, I use a `Buffer` parameter that should implement `Borrow` and `BorrowMut` (not `AsRef` / `AsMut` as they don't have the blanket `impl Borrow<T> for T`).

This work will lead up to #388, making it easier to build domain names quickly.  Note that I've dropped all support for `Symbol` here.

### DIFF
--- a/src/base/name/absolute.rs
+++ b/src/base/name/absolute.rs
@@ -243,7 +243,8 @@ impl Name<Bytes> {
 
     /// Creates a domain name atop a Bytes from its string representation.
     pub fn bytes_from_str(s: &str) -> Result<Self, ScanError> {
-        FromStr::from_str(s)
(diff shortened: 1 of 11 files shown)


bal-e:
Closing this in favor of a `base::new_name` module.

PR #397, opened 2024-10-01
Usability tweaks

ximon18:
Various trait impls and other tweaks to enable more use cases.

Based on code that was originally in the `xfr` branch for now closed PR #335 which used these changes to implement the `ZoneMaintainer`.


src/tsig/mod.rs
```diff
@@ -389,6 +393,15 @@ impl AsRef<Key> for Key {
     }
 }
 
+//--- Display
+
+#[cfg(feature = "std")]
+impl fmt::Display for Key {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_fmt(format_args!("{}", self.name))
```
partim, line 401: This is equivalent to

```
write!(f, "{}", self.name)
```

or probably even just `self.name.fmt(f)`.

src/zonetree/update.rs
```diff
@@ -1309,6 +1309,21 @@ pub enum Error {
...
 
+//--- Display
+
+impl std::fmt::Display for Error {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            Error::OutOfZone => f.write_str("OutOfZone"),
+            Error::NotSoaRecord => f.write_str("NotSoaRecord"),
+            Error::IoError(err) => {
+                f.write_fmt(format_args!("I/O error: {err}"))
```
partim, line 1320: As above: Probably clearer to use `write!()`.

src/zonetree/zone.rs
```diff
@@ -13,9 +13,47 @@ use super::traits::WritableZone;
...
 use super::{parsed, ReadableZone, ZoneStore};
 
+//------------ ZoneKey -------------------------------------------------------
+
+/// A key that uniquely identifies a zone.
+///
+/// A zone is identified by the owner name of the apex and its class. Every
+/// record in a zone must be at or under the apex owner name and be of the
+/// same class.
+pub type ZoneKey = (StoredName, Class);
```
partim, line 23: `ZoneKey` is potentially confusing what with DNSSEC also using zone keys …

ximon18: Good point.

ximon18: Renamed to `ZoneTuple`.

partim (approved):

PR #400, opened 2024-10-02
Set the AA flag and include glue in responses.

ximon18:
This PR attempts to fix two issues and adds a Stelline test based on the scenario in RFC 1034 section 6 to demonstrate compliance.

1. AA flag not set on authoritative responses. This PR sets the AA flag on zone queries whose response is not a referral. Answers at zone cuts are constructed by calling `NodeAnswer::authority()` which has been updated to set an `authoritative` flag to false, while other `NodeAnswer` constructors set the flag to true. `NodeAnswer::into_answer()` then carries the flag through to `Answer` and `Answer::to_message()` sets the AA header flag based on this new `authoritative` flag.

2. Missing glue in responses. This PR extends `AnswerAuthority::new()` to take additional records (provided from glue) as well as other data, and `Answer::to_message()` is extended to add any additional records to the response.

This PR also extends Stelline to support `MATCH ANY_ANSWER` whereby if additional answers are supplied using the `EXTRA_PACKET` Stelline directive then the expected response will be matched against any of them, rather than the default behaviour of treating multiple answers as one large answer (which was supported already and used for unknown ordered and split large XFR answers across multiple responses).


src/zonetree/answer.rs
```diff
@@ -40,28 +41,35 @@ pub struct Answer {
...
+
+    /// Should the answer be flagged as authoritative?
+    authoritative: bool,
 }
 
 impl Answer {
     /// Creates an "empty" answer.
     ///
-    /// The answer, authority and additinal sections will be empty.
+    /// The answer, authority and additional sections will be empty.
```
partim, line 52: You might also want to mention that the answer will be treated as non-authoritative.

src/zonetree/answer.rs
```diff
@@ -87,6 +95,11 @@ impl Answer {
         self.authority = Some(authority)
     }
 
+    /// Marks the response authoritative for the QNAME.
```
partim, line 98: Documentation nit-pickery: The method marks _whether_ the response (also, this should probably be “answer”) is authoritative – i.e., the text should reflect the presence of the boolean argument.

src/zonetree/answer.rs
```diff
@@ -247,8 +274,13 @@ pub struct AnswerAuthority {
     /// The NS record set if it should be included.
     ns: Option<SharedRrset>,
 
-    /// The DS record set if it should be included..
+    /// The DS record set if it should be included.
     ds: Option<SharedRrset>,
+
+    /// Any additional address records to include.
```
partim, line 280: This is conceptionally weird: The type represents the authority section, but these records go into the additional section.

partim: On re-reading this: It might be good to split out additional records into its own field in `Answer` and have a type `AnswerAdditional` that collects all the additional records? We need to distinguish between necessary and optional additional RRsets (the latter can be dropped if we lack space in the response).

src/zonetree/answer.rs
ximon18: Ah, it didn't have the argument when I first wrote the documentation! Good catch.

ximon18: Done. Please check.

src/zonetree/answer.rs
ximon18: Done. Please check.

src/zonetree/answer.rs
ximon18: Done. Please check.

partim (approved):

PR #406, opened 2024-10-07
Loading, storing, and generating DNSSEC keys

bal-e:
This PR defines an ergonomic basis for DNSSEC signing.

The following features are supported:

- [x] Keys can be generated.
- [x] Keys can be parsed from or serialized to the conventional BIND format.
- [x] Keys can be used to sign byte sequences.
- [x] Public keys can be parsed from or serialized to DNSKEY records.

Cryptographic primitives are implemented using Ring or OpenSSL, at the user's choice.  Note that OpenSSL supports more algorithms.  For simplicity, a combined key type is provided which uses Ring where possible and falls back to OpenSSL for other algorithms.

Signatures can be created using the following algorithms:

- [ ] RSA/SHA-1: support is prohibited as per RFC 8624.
- [x] RSA/SHA-256 (also with Ring, for 2048-bit or larger keys).
- [ ] RSA/SHA-512: support is not recommended as per RFC 8624.
- [x] ECDSA P-256/SHA-256 (also with Ring).
- [x] ECDSA P-384/SHA-384 (also with Ring).
- [x] Ed25519 (also with Ring).
- [x] Ed448.

In DNSSEC, keys are associated with important metadata, such as who they belong to and how they can be used (Zone Signing Keys sign resource records, while Key Signing Keys sign Zone Signing Keys).  This implementations provides low-level or "raw" types which do not include this metadata, as well as higher-level types which do include it.

- `sign::generic::SecretKey`: A generic representation of a secret key.  It does not support any cryptographic operations itself.
- `sign::{openssl,ring,common}::SecretKey`: A secret key that supports cryptographic operations and that can be used for signing.
- `sign::Signer`: A secret key associated with important metadata.
- `validate::RawPublicKey`: A generic representation of a public key.
- `validate::Key`: A public key associated with important metadata.

The actual signing functionality is provided by the `sign::SignRaw` trait.  This provides a *synchronous* `sign_raw()` function, which should be used for on-CPU signing operations.  In the future, an *asynchronous* signing interface will be provided, for use with off-CPU signing operations (e.g. via HSMs).


src/sign/mod.rs
```diff
@@ -1,10 +1,57 @@
 //! DNSSEC signing.
 //!
 //! **This module is experimental and likely to change significantly.**
+//!
+//! Signatures are at the heart of DNSSEC -- they confirm the authenticity of a
+//! DNS record served by a secure-aware name server.  But name servers are not
```
ximon18, line 6: RFC 9499 defines "DNSSEC-aware" and refers to "security-aware" but doesn't mention "secure-aware".

ximon18, line 6: I was taught, though it seems it is not incorrect, not to start sentences with "But". Grammarly advises gently against it too: https://www.grammarly.com/blog/sentences/starting-a-sentence-with-a-conjunction/

src/sign/mod.rs
```diff
@@ -1,10 +1,57 @@
...
+//! usually creating those signatures themselves.  Within a DNS zone, it is the
```
ximon18, line 7: I'm not sure what you mean by a name server not signing a zone itself. Some name servers do indeed do this themselves.

src/sign/mod.rs
```diff
@@ -1,10 +1,57 @@
...
 pub mod key;
-//pub mod openssl;
+pub mod openssl;
 pub mod records;
 pub mod ring;
+
+/// Signing DNS records.
+///
+/// Implementors of this trait own a private key and sign DNS records for a zone
+/// with that key.  Signing is a synchronous operation performed on the current
```
ximon18, line 26: I don't think it being synchronous rules out I/O, it just makes it blocking but could be done from a background thread or slow running task pool.

src/sign/mod.rs
```diff
@@ -1,10 +1,57 @@
...
+    /// - [`SecAlg::DSA_NSEC3_SHA1`] (highly insecure, do not use)
+    /// - [`SecAlg::RSASHA1_NSEC3_SHA1`] (insecure, not recommended)
+    /// - [`SecAlg::RSASHA256`]
+    /// - [`SecAlg::RSASHA512`] (not recommended)
+    /// - [`SecAlg::ECC_GOST`] (do not use)
+    /// - [`SecAlg::ECDSAP256SHA256`]
+    /// - [`SecAlg::ECDSAP384SHA384`]
+    /// - [`SecAlg::ED25519`]
+    /// - [`SecAlg::ED448`]
+    fn algorithm(&self) -> SecAlg;
```
ximon18, line 48: I don't the trait documentation is the place to limit the set of algorithms supported, that is dependent on the implementers of the trait right?

src/sign/mod.rs
```diff
@@ -1,10 +1,57 @@
...
-//pub mod openssl;
+pub mod openssl;
 pub mod records;
 pub mod ring;
+
+/// Signing DNS records.
+///
+/// Implementors of this trait own a private key and sign DNS records for a zone
+/// with that key.  Signing is a synchronous operation performed on the current
+/// thread; this rules out implementations like HSMs, where I/O communication is
```
ximon18, line 27: Given that this trait has `fn sign()` which defines an action, a verb, I would think `Signer` might be a better name for this trait than `Sign`.

ximon18: Also, have you seen https://github.com/NLnetLabs/domain/pull/217/files#diff-af737f2dd1178f239bcc7823d35afed2a2db6947676918377c4e8787c122bf89R28 ?

src/sign/mod.rs
```diff
@@ -1,10 +1,57 @@
...
+    /// - [`SecAlg::ECC_GOST`] (do not use)
+    /// - [`SecAlg::ECDSAP256SHA256`]
+    /// - [`SecAlg::ECDSAP384SHA384`]
+    /// - [`SecAlg::ED25519`]
+    /// - [`SecAlg::ED448`]
+    fn algorithm(&self) -> SecAlg;
+
+    /// Compute a signature.
+    ///
+    /// A regular signature of the given byte sequence is computed and is turned
```
ximon18, line 52: What would an "irregular" signature be?

src/sign/generic.rs
```diff
@@ -0,0 +1,512 @@
+use core::{fmt, mem, str};
```
ximon18, line 1: General tip: please follow the style guidelines at https://github.com/NLnetLabs/domain/blob/contributing-doc/CONTRIBUTING.md.

src/sign/openssl.rs
```diff
@@ -1,58 +1,373 @@
...
+    /// The private key.
+    pkey: PKey<Private>,
+}
+
+impl SecretKey {
+    /// Use a generic secret key with OpenSSL.
+    ///
+    /// # Panics
+    ///
+    /// Panics if OpenSSL fails or if memory could not be allocated.
```
ximon18, line 32: I don't think we want an entire name server to crash if there is a failure signing one zone so I'm not sure that this should panic if OpenSSL fails.

src/sign/openssl.rs
```diff
@@ -1,58 +1,373 @@
...
+            pkey,
+        })
+    }
+
+    /// Export this key into a generic secret key.
+    ///
+    /// # Panics
+    ///
+    /// Panics if OpenSSL fails or if memory could not be allocated.
+    pub fn export<B>(&self) -> generic::SecretKey<B>
```
ximon18, line 113: I tend to think of import and export as referring to something outside the application, e.g. a file. I would suggest naming this "as_generic()`, or could this perhaps be written as a From impl?

src/sign/openssl.rs
```diff
@@ -1,58 +1,373 @@
...
+            _ => unreachable!(),
+        }
+    }
+
+    /// Export this key into a generic public key.
+    ///
+    /// # Panics
+    ///
+    /// Panics if OpenSSL fails or if memory could not be allocated.
+    pub fn export_public<B>(&self) -> generic::PublicKey<B>
```
ximon18, line 159: Same concern as above about the term "export".

ximon18: Same concern as above about panicking if OpenSSL fails.

ximon18 (commented):
Some incomplete initial review feedback.

Cargo.toml
```diff
@@ -32,6 +32,7 @@ heapless       = { version = "0.8", optional = true }
 libc           = { version = "0.2.153", default-features = false, optional = true } # 0.2.79 is the first version that has IP_PMTUDISC_OMIT
 parking_lot    = { version = "0.12", optional = true }
 moka           = { version = "0.12.3", optional = true, features = ["future"] }
+openssl        = { version = "0.10.57", optional = true } # 0.10.57 upgrades to 'bitflags' 2.x
```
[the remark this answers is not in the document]
bal-e: Well, ring and openssl offer equivalent functionality; users of `domain` can choose which one they prefer.  So we will have to add feature gates for enabling one or the other.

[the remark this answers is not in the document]
ximon18: At present they don't offer equivalent functionality though, you need openssl to do signing because `ECDSAP256SHA256` is `MUST` for signing according to https://datatracker.ietf.org/doc/html/rfc8624#section-3.1 but we can currently only use externally created ECDSAP256SHA256 keys with OpenSSL, not with Ring.

src/sign/generic.rs
```diff
@@ -0,0 +1,512 @@
...
+        }
+    }
+
+    /// Parse a key from the conventional DNS format.
+    ///
+    /// - For RSA, see RFC 5702, section 6.
+    /// - For ECDSA, see RFC 6605, section 6.
+    /// - For EdDSA, see RFC 8080, section 6.
+    pub fn from_dns(data: &str) -> Result<Self, DnsFormatError>
+    where
```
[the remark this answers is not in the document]
bal-e: I intentionally left out a lot of information here in the same spirit as `ring::error::Unspecified` -- it could potentially leak information about the private key.  In any case, providing specific error messages is useless from an end-user perspective: there's an infinitesimal chance they actually opened the private key file and edited it by hand.  At best, any error messages visible to the end user should link to a specification of the format (not that I can find any).

src/sign/generic.rs
```diff
@@ -0,0 +1,512 @@
...
+            .next()
+            .ok_or(DnsFormatError::Misformatted)?
+            .parse::<u8>()
+            .map_err(|_| DnsFormatError::Misformatted)?;
+        let name = words.next().ok_or(DnsFormatError::Misformatted)?;
+        if words.next().is_some() {
+            return Err(DnsFormatError::Misformatted);
+        }
+
+        match (code, name) {
```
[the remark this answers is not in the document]
bal-e: Yes, this is deliberately made a subset of all keys that we should validate -- validation related code belongs in the `validate` submodule.  I'm probably going to restructure this so that `sign` depends on `validate` and uses their public key type definitions.

src/sign/generic.rs
```diff
@@ -0,0 +1,512 @@
...
+    ///
+    /// The public key is a 57-byte encoding of the public point.
+    Ed448([u8; 57]),
+}
+
+impl<B: AsRef<[u8]>> PublicKey<B> {
+    /// The algorithm used by this key.
+    pub fn algorithm(&self) -> SecAlg {
+        match self {
+            Self::RsaSha1(_) => SecAlg::RSASHA1,
```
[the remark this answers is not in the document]
bal-e: Storing the `SecAlg` alongside the enum values would just introduce an additional match statement with a fallback `_ => unreachable!()`.  This seemed the simpler and more explicit option.

src/sign/generic.rs
```diff
@@ -0,0 +1,512 @@
...
+            }
+        }
+    }
+
+    /// Parse a key from the conventional DNS format.
+    ///
+    /// - For RSA, see RFC 5702, section 6.
+    /// - For ECDSA, see RFC 6605, section 6.
+    /// - For EdDSA, see RFC 8080, section 6.
+    pub fn from_dns(data: &str) -> Result<Self, DnsFormatError>
```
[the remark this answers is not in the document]
ximon18: Nothing great, just the following:
- https://kb.isc.org/docs/dnssec-key-and-signing-policy
- https://lists.nlnetlabs.nl/pipermail/ldns-users/2011-November/000447.html
- https://github.com/search?q=repo%3Aisc-projects%2Fbind9+v1.3&type=code

src/sign/generic.rs
```diff
@@ -0,0 +1,512 @@
...
+            Self::RsaSha512(_) => SecAlg::RSASHA512,
+            Self::EcdsaP256Sha256(_) => SecAlg::ECDSAP256SHA256,
+            Self::EcdsaP384Sha384(_) => SecAlg::ECDSAP384SHA384,
+            Self::Ed25519(_) => SecAlg::ED25519,
+            Self::Ed448(_) => SecAlg::ED448,
+        }
+    }
+
+    /// Construct a DNSKEY record with the given flags.
+    pub fn into_dns<Octs>(self, flags: u16) -> Dnskey<Octs>
```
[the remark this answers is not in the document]
bal-e: I want to avoid the same public key being transformed into DNSKEY records with different flags values (since that is almost certainly incorrect).  I'm going to address this a bit by storing that metadata in the primary public key type directly in an overhaul of this PR.

src/sign/mod.rs
bal-e: This documentation is to inform async users that they can call the `sign` function without worrying about blocking the thread.  It's unclear, though -- I'll fix that.

src/sign/mod.rs
bal-e: Oops, typo.  I meant security-aware.

src/sign/generic.rs
bal-e: Right, I should have seen that.  Will fix.

src/sign/generic.rs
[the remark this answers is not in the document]
bal-e: cc @Philip-NLnetLabs, it looks like BIND actually stores key rollover information within the private key files themselves (see Ximon's second link).

src/sign/generic.rs
ximon18: It's not that easy to find as it's in a branch so I'm not surprised you hadn't seen it, we should approve and merge it.

src/sign/generic.rs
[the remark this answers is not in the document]
bal-e: I'll put a rigorous definition of the private key format somewhere for others to refer to.

src/sign/generic.rs
```diff
@@ -0,0 +1,512 @@
...
+    pub e: B,
+}
+
+impl<B> From<RsaSecretKey<B>> for RsaPublicKey<B>
+where
+    B: AsRef<[u8]> + AsMut<[u8]> + Default,
+{
+    fn from(mut value: RsaSecretKey<B>) -> Self {
+        Self {
+            n: mem::take(&mut value.n),
```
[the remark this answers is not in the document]
bal-e: I can't move out of `RsaSecretKey` because it implements `Drop`.  `mem::take` seemed like the next best thing.

src/sign/generic.rs
```diff
@@ -0,0 +1,512 @@
...
+    // Trim any pending newlines.
+    let data = data.trim_start();
+
+    // Stop if there's no more data.
+    if data.is_empty() {
+        return Ok(None);
+    }
+
+    // Get the first line (NOTE: CR LF is handled later).
+    let (line, rest) = data.split_once('\n').unwrap_or((data, ""));
```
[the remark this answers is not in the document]
bal-e: I'm not sure which function you're referring to.  `str::split_once()` and `Option::unwrap_or` don't panic.

src/sign/mod.rs
bal-e: I was trying to fight back my tendency to write run-on sentences :slightly_smiling_face: I'll rewrite it.

src/sign/mod.rs
bal-e: I didn't phrase it well -- I meant that signing seems to usually be done offline.  I'll update that if online signing is a common convention.

src/sign/mod.rs
bal-e: If I'm reading the contributing docs correctly:

> Traits are named using verb phrases that describe what actions the trait represents. An exception are conversion traits that follow the usual Rust convention for naming conversion functions.

That should mean that `Sign` is the appropriate name.  Also see most of the core traits in the standard library like `Copy`, `Clone`, etc.

bal-e: RE PR #217: Most of the time you can get away with `&[u8]`, but I think it's a bit difficult for RR sets (because you might have to reorder them).  I'll replace this with an init-update-finalize interface.

src/sign/mod.rs
bal-e: A signature that is context-specific, e.g. that prepends a fixed string to the data being signed, or that uses an unconventional digest algorithm.  I'm going to make it more explicit, though, especially after discovering that signatures can be fixed-length or DER-encoded.

src/sign/openssl.rs
bal-e: OpenSSL exposes very little information about possible failures, but since we are only performing CPU-based operations here, the only reasonable failure case is out-of-memory -- which will probably cause panics anyway.  Issues with malformed keys should result in an `Err` instead, but that's hard to verify -- I might have to do something silly like sign a `[u8; 0]` here just to make OpenSSL perform those checks.

src/sign/openssl.rs
bal-e: Yeah, I'll turn this into a `From` impl.  The primary reason to "export" to a generic key is indeed to serialize it on disk, but the additional layer of indirection makes the name confusing.

Cargo.toml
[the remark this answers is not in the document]
bal-e: That's true: Ring's interface makes it impossible to import ECDSA keys from the BIND private key format.  However, users who would like to avoid OpenSSL can still opt to disable it and only use other kinds of keys.  As long as Ring's limitations are adequately described in the documentation, I think we should offer both.

src/sign/generic.rs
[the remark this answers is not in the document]
ximon18: My mistake, I misread the code.

.github/workflows/ci.yml
```diff
@@ -10,13 +10,26 @@ jobs:
         rust: [1.76.0, stable, beta, nightly]
     env:
         RUSTFLAGS: "-D warnings"
+        VCPKG_ROOT: "${{ github.workspace }}\\vcpkg"
+        VCPKGRS_TRIPLET: x64-windows-release
+        VCPKGRS_DYNAMIC: 1
```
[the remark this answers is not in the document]
bal-e: Yep, will do; it just ensures that OpenSSL is dynamically linked.

src/sign/mod.rs
bal-e: Actually, Ed25519 and Ed448 don't support init-update-finalize interfaces -- both Ring and OpenSSL only offer signing functions which take the entire message as a single slice.  Rather than exposing a streaming interface where the underlying implementation performs allocations, I'm going to let the caller organize the data themselves (they may be able to do so more efficiently).

ximon18: > RE PR #217: Most of the time you can get away with `&[u8]`, but I think it's a bit difficult for RR sets (because you might have to reorder them). I'll replace this with an init-update-finalize interface.

I'm a bit confused by this. Surely the caller decides what needs signing, you just sign what you're given? Which re-ordering are you referring to?

src/sign/openssl.rs
ximon18: The problem with panicking in a library is you don't know what context you are being used in. The signing might be a tiny unimportant part of an application with much more important objectives for which it is imperative that it not abort due to panicking. I'm also wary of assuming reasons why calls to OpenSSL may fail, are you sure you can say it is only OOM that will cause signing failure?

Cargo.toml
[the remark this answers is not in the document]
ximon18: They can't avoid OpenSSL and do compliant signing or validation as that requires `ECDSAP256SHA256`. Perhaps indeed there are use cases which wouldn't need full compliance, but we should be careful not to make a footgun for users of the library. IMO.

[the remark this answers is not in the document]
bal-e: Actually, I've rewritten the interface so that ECDSA works on Ring.  Unlike `ldns`, our API now requires key signing to load _both_ the public and private keys from files at startup.  I was trying to avoid this since it seemed too brittle to require both files to be present (especially since most people would assume you only need the private key file), but it has the added benefit of letting us check for corruption or mismatches.

src/validate.rs
```diff
@@ -10,14 +10,361 @@ use crate::base::name::Name;
...
 use bytes::Bytes;
 use octseq::builder::with_infallible;
 use ring::{digest, signature};
+use std::boxed::Box;
 use std::vec::Vec;
 use std::{error, fmt};
 
+/// A generic public key.
+#[derive(Clone, Debug)]
+pub enum PublicKey {
```
ximon18, line 25: I'm not sure about PublicKey being in the validate module. For example I want to construct a key for signing and for that I need the public part as well as the private part, and to use validate for that seems strange. Could this not go in src/sign/key.rs?

src/validate.rs
```diff
@@ -10,14 +10,361 @@ use crate::base::name::Name;
...
+    /// The first three fields are simple integers, while the last field is
+    /// Base64 encoded data (with or without padding).  The [`from_dnskey()`]
+    /// and [`to_dnskey()`] read from and serialize to the Base64-decoded data
+    /// format.
+    ///
+    /// [`from_dnskey()`]: Self::from_dnskey()
+    /// [`to_dnskey()`]: Self::to_dnskey()
+    ///
+    /// The `<comment>` is any text starting with an ASCII semicolon.
+    pub fn from_dnskey_text(
```
ximon18, line 164: Point of consistency: we have `SecretKey::parse_from_bind()` and `PublicKey::from_dnskey_text()` - why is one called parse and the other is not?

src/validate.rs
bal-e: The DNSKEY record conversion that `PublicKey` offers is also relevant to validation.  Furthermore, I want to put most of the NSEC3 signing logic (everything up to `Sign::sign()`) in `validate` since they'll use it for validation too.

src/validate.rs
bal-e: Right, will rename.

src/sign/mod.rs
bal-e: I meant that init-update-finalize paradigms are nice when the caller has RR data lying in non-contiguous slices, which is almost certainly the case if the RRs have to be sorted first.  That's why callers would likely benefit from having an init-update-finalize interface which lets them avoid allocations.  Unfortunately, EdDSA makes that impossible.

src/sign/mod.rs
bal-e: These are the official recommendations as per RFC 8624 -- I've placed it here so that users know which algorithms to focus on.

ximon18: Such advice won't age well, I'd just link to the RFC as that will get marked as obsoleted/updated by a later RFC.

src/validate.rs
```diff
@@ -1,23 +1,723 @@
...
+    ) -> Result<Self, ParseDnskeyTextError>
+    where
+        Octs: FromBuilder,
+        Octs::Builder: EmptyBuilder + Composer,
+    {
+        // Ensure there is a single line in the input.
+        let (line, rest) = dnskey.split_once('\n').unwrap_or((dnskey, ""));
+        if !rest.trim().is_empty() {
+            return Err(ParseDnskeyTextError::Misformatted);
+        }
```
ximon18, line 247: This code failed to parse files produced by BIND `dnssec-keygen` which have leading comment lines, e.g.:

```
$ cat /tmp/x/Kexample.com.+013+04622.key 
; This is a key-signing key, keyid 4622, for example.com.
; Created: 20241015203641 (Tue Oct 15 22:36:41 2024)
; Publish: 20241015203641 (Tue Oct 15 22:36:41 2024)
; Activate: 20241015203641 (Tue Oct 15 22:36:41 2024)
example.com. IN DNSKEY 257 3 13 e9Usj72J/s5jY9Rxi+kc9M7n3f1osxzVjD3HaS9GYTHqjDNolHraAevV RjH8pj+I7JtcT/pE1oqwxnxC0pKeAw==
```

The following modified code worked for me:

```suggestion
        // Skip leading comment lines (BIND uses these to record key timing
        // data).
        let mut line = dnskey;
        while let Some((this_line, rest)) = line.split_once('\n') {
            if !this_line.starts_with(';') {
                // Ensure there is a single data line in the input.
                if !rest.trim().is_empty() {
                    return Err(ParseDnskeyTextError::Misformatted);
                } else {
                    line = this_line;
                    break;
                }
            } else {
                line = rest;
            }
        }
```

src/validate.rs
```diff
@@ -1,23 +1,723 @@
...
+    ///
+    /// From RFC 4034, section 2.1.1:
+    ///
+    /// > Bit 7 of the Flags field is the Zone Key flag.  If bit 7 has value
+    /// > 1, then the DNSKEY record holds a DNS zone key, and the DNSKEY RR's
+    /// > owner name MUST be the name of a zone.  If bit 7 has value 0, then
+    /// > the DNSKEY record holds some other type of DNS public key and MUST
+    /// > NOT be used to verify RRSIGs that cover RRsets.
+    pub fn is_zone_signing_key(&self) -> bool {
+        self.flags & (1 << 8) != 0
```
ximon18, line 86: This didn't correctly detect the ZSK flag for me when the input file contained a flags value of 256, for which RFC 4034 says _"Value 256 indicates that the Zone Key bit (bit 7) in the Flags field has value 1".

The following modified code worked for me:

```suggestion
        self.flags & 0b0000_0001_0000_0000 != 0
```

src/validate.rs
```diff
@@ -1,23 +1,723 @@
...
+    /// > intended to be a hint to zone signing or debugging software as to
+    /// > the intended use of this DNSKEY record; validators MUST NOT alter
+    /// > their behavior during the signature validation process in any way
+    /// > based on the setting of this bit.  This also means that a DNSKEY RR
+    /// > with the SEP bit set would also need the Zone Key flag set in order
+    /// > to be able to generate signatures legally.  A DNSKEY RR with the SEP
+    /// > set and the Zone Key flag not set MUST NOT be used to verify RRSIGs
+    /// > that cover RRsets.
+    pub fn is_secure_entry_point(&self) -> bool {
+        self.flags & 1 != 0
```
ximon18, line 118: This didn't correctly detect the SEP flag for me when the input file contained a flags value of 257 (which should be ZONE key flag set AND SEP key flag set).

The following modified code worked for me:

```suggestion
        self.flags & 0b0000_0000_0000_0001 != 0
```

src/validate.rs
bal-e: Thanks!  I'll make it a bit more resilient too (to skip whitespace preceding the colon).

Cargo.toml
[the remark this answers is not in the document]
bal-e: I've now introduced separate features for OpenSSL and Ring.  Users can choose to disable both (... once `unstable-validate` doesn't unconditionally require Ring) and bring their own cryptography.

src/sign/openssl.rs
bal-e: I've removed the vast majority of panics and replaced them with opaque "implementation error" variants.

src/lib.rs
```diff
@@ -36,14 +36,14 @@
 #![cfg_attr(not(feature = "resolv"), doc = "* resolv:")]
 //!   An asynchronous DNS resolver based on the
 //!   [Tokio](https://tokio.rs/) async runtime.
-#![cfg_attr(feature = "sign", doc = "* [sign]:")]
-#![cfg_attr(not(feature = "sign"), doc = "* sign:")]
+#![cfg_attr(feature = "unstable-sign", doc = "* [sign]:")]
```
ximon18, line 39: The sign on the right should also be changed to unstable-sign. Same issue below for other renamed features.

src/lib.rs
```diff
@@ -86,8 +86,8 @@
 //!   [ring](https://github.com/briansmith/ring) crate.
 //! * `serde`: Enables serde serialization for a number of basic types.
 //! * `sign`: basic DNSSEC signing support. This will enable the
-#![cfg_attr(feature = "sign", doc = "  [sign]")]
-#![cfg_attr(not(feature = "sign"), doc = "  sign")]
+#![cfg_attr(feature = "unstable-sign", doc = "  [sign]")]
```
ximon18, line 89: Update: I have moved this GH comment down to the correct part of the file.

The `sign` on the right should also be changed to `unstable-sign`. Same issue below for other renamed features. Also, now unstable features that previoulsy were not marked as such should be moved down the doc to the unstable features section.

src/sign/mod.rs
```diff
@@ -1,10 +1,387 @@
...
+
+// TODO: Conversion to and from key files
+
+//----------- SignRaw --------------------------------------------------------
+
+/// Low-level signing functionality.
+///
+/// Types that implement this trait own a private key and can sign arbitrary
+/// information (for zone signing keys, DNS records; for key signing keys,
+/// subsidiary public keys).
```
ximon18, line 238: What is a "subsidiary public key"? There's only one reference to "subsidiary" in https://www.rfc-editor.org/rfc/rfc9499.html and https://www.rfc-editor.org/rfc/rfc4033.html, and it's not clear to me from that.

src/sign/mod.rs
```diff
@@ -1,10 +1,387 @@
...
+///
+/// Before a key can be used for signing, it should be validated.  If the
```
ximon18, line 240: Is this validation something the user is expected to do? Or is this saying that such validation is done by the implementation because it should be done?

ximon18 (commented):
I've looked at the RustDoc and given some feedback in comments on this PR.

I have not yet reviewed the actual code in its current form.

src/sign/ring.rs
```diff
@@ -1,140 +1,442 @@
-//! Key and Signer using ring.
+//! DNSSEC signing using `ring`.
+//!
+//! This backend supports the following algorithms:
+//!
+//! - RSA/SHA-256 (2048-bit keys or larger)
```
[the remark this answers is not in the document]
Philip-NLnetLabs: I think Ring is fine for validation. I didn't find any issues.

src/lib.rs
bal-e: The `sign` on the right refers to the module, whose name has not changed.  I'll move these down to the unstable feature list.

.github/workflows/ci.yml
```diff
@@ -10,13 +10,28 @@ jobs:
         rust: [1.76.0, stable, beta, nightly]
     env:
         RUSTFLAGS: "-D warnings"
+        # We use 'vcpkg' to install OpenSSL on Windows.
```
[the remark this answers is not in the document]
bal-e: The `openssl` crate relies on the `vcpkg` crate to find OpenSSL on Windows.  The `VCPKGRS_*` environment variable configures the `vcpkg` crate to try dynamically linking to OpenSSL (the `openssl` crate doesn't do this itself, but will fail if OpenSSL is statically linked).

.github/workflows/ci.yml
```diff
@@ -37,6 +52,8 @@ jobs:
       uses: hecrj/setup-rust-action@v2
       with:
         rust-version: "1.68.2"
+    - name: Install OpenSSL
+      run: sudo apt install libssl-dev
```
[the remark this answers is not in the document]
bal-e: It works fine on CI, but yes, I need to fix this.

src/sign/bytes.rs
```diff
@@ -0,0 +1,531 @@
+//! A generic representation of secret keys.
```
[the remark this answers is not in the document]
bal-e: The names exported by this module are explicit, and the module is never directly exposed to users so it cannot confuse them.  The previous name for this module was `generic`, but that doesn't quite communicate its usage either.  I think `bytes` is a relatively clear name once somebody looks at its contents, but I'm happy to take better suggestions.

src/sign/mod.rs
```diff
@@ -1,10 +1,387 @@
...
 //!
 //! **This module is experimental and likely to change significantly.**
-#![cfg(feature = "sign")]
-#![cfg_attr(docsrs, doc(cfg(feature = "sign")))]
+//!
+//! Signatures are at the heart of DNSSEC -- they confirm the authenticity of
+//! a DNS record served by a security-aware name server.  Signatures can be
+//! made "online" (in an authoritative name server while it is running) or
+//! "offline" (outside of a name server).  Once generated, signatures can be
+//! serialized as DNS records and stored alongside the authenticated records.
```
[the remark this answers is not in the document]
bal-e: This documentation is here to provide some context to users as to how this module should be used.  Instead of removing this, perhaps I should add a second paragraph which explains this more thoroughly?

src/sign/mod.rs
```diff
@@ -1,10 +1,387 @@
...
+//! - ECDSA P-256/SHA-256
+//! - ECDSA P-384/SHA-384
+//! - Ed25519
+//! - Ed448
+//!
+//! # Importing and Exporting
+//!
+//! The [`SecretKeyBytes`] type is a generic representation of a secret key as
+//! a byte slice.  While it does not offer any cryptographic functionality, it
+//! is useful to transfer secret keys stored in memory, independent of any
```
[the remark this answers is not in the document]
bal-e: Transfer between functions, or to and from the disk.  I don't think "communicate" fits either.

src/sign/mod.rs
```diff
@@ -1,10 +1,387 @@
...
+//!
+//! A DNSSEC key actually has two components: a cryptographic key, which can
+//! be used to make and verify signatures, and key metadata, which defines how
+//! the key should be used.  These components are brought together by the
+//! [`SigningKey`] type.  It must be instantiated with a cryptographic key
+//! type, such as [`common::KeyPair`], in order to be used.
+//!
+//! # Example Usage
+//!
+//! At the moment, only "low-level" signing is supported.
```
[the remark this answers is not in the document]
bal-e: Rather than making users copy that example whenever they want to perform an import, we should just expose an `std`-gated function on `SigningKey` to do it for them (given the file name(s) to lead from).  What do you think?

src/sign/mod.rs
```diff
@@ -1,10 +1,387 @@
...
+//!
+//! Users can choose to bring their own cryptography by providing their own
+//! `KeyPair` type that implements [`SignRaw`].  Note that `async` signing
+//! (useful for interacting with cryptographic hardware like HSMs) is not
+//! currently supported.
+//!
+//! While each cryptographic backend can support a limited number of signature
+//! algorithms, even the types independent of a cryptographic backend (e.g.
+//! [`SecretKeyBytes`] and [`GenerateParams`]) support a limited number of
+//! algorithms.  They are:
```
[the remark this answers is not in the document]
bal-e: I wanted some module-level documentation so that readers don't need to jump between declarations to get a complete high-level understanding of the module.  I guess it's a balance between concision and readability.

src/sign/mod.rs
```diff
@@ -1,10 +1,387 @@
...
+//----------- GenerateParams -------------------------------------------------
+
+/// Parameters for generating a secret key.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum GenerateParams {
+    /// Generate an RSA/SHA-256 keypair.
+    RsaSha256 {
+        /// The number of bits in the public modulus.
+        ///
+        /// A ~3000-bit key corresponds to a 128-bit security level.  However,
```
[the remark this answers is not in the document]
bal-e: Nice find with NIST publication!  I'm adding a link to it.

src/sign/common.rs
```diff
@@ -0,0 +1,274 @@
...
+
+#[cfg(feature = "ring")]
+use super::ring;
+
+//----------- KeyPair --------------------------------------------------------
+
+/// A key pair based on a built-in backend.
+///
+/// This supports any built-in backend (currently, that is OpenSSL and Ring).
+/// Wherever possible, the Ring backend is preferred over OpenSSL -- but for
```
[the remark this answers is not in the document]
bal-e: I'm updating the language used, but I'm not sure where a discussion of why Ring is preferred should go.  We could place it on this type, on the module-level documentation for `common`, at the module-level documentation for `sign`, or in the `ring` module.

src/sign/bytes.rs
```diff
@@ -0,0 +1,531 @@
...
+use crate::validate::RsaPublicKeyBytes;
+
+//----------- SecretKeyBytes -------------------------------------------------
+
+/// A secret key expressed as raw bytes.
+///
+/// This is a low-level generic representation of a secret key from any one of
+/// the commonly supported signature algorithms.  It is useful for abstracting
+/// over most cryptographic implementations, and it provides functionality for
+/// importing and exporting keys from and to the disk.
```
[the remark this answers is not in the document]
bal-e: Well, the BIND private key format exists for the purpose of storage on disk.  I think directly talking about the BIND format in this paragraph is more confusing than just mentioning storage functionality.

src/sign/bytes.rs
```diff
@@ -0,0 +1,531 @@
...
+/// [RFC 6605]: https://www.rfc-editor.org/rfc/rfc6605
+/// [RFC 8080]: https://www.rfc-editor.org/rfc/rfc8080
+///
+/// In this format, a private key is a line-oriented text file.  Each line is
+/// either blank (having only whitespace) or a key-value entry.  Entries have
+/// three components: a key, an ASCII colon, and a value.  Keys contain ASCII
+/// text (except for colons) and values contain any data up to the end of the
+/// line.  Whitespace at either end of the key and the value will be ignored.
+///
+/// Every file begins with two entries:
```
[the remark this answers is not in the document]
bal-e: I've never heard of BIND private key files being able to store comments.  The DNSKEY record format used for public key files can, however.  Can you find a reference which demonstrates support for comments here?

src/sign/ring.rs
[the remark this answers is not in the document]
bal-e: This is the `sign` module, not the `validate` module.  While RSA-SHA1 is required for validation, it is prohibited for signing.  Any discussion of the need of RSA-SHA1 needs to be moved to the `validate` module documentation instead.

src/sign/mod.rs
bal-e: I'm referring to the ZSKs that would be signed by a KSK; I'd describe that relationship as subsidiary, but I don't know if there's any conventional term for it.  I believe the ZSKs being signed must be part of the same zone, but I'm not sure about that either.  Should I replace "subsidiary public key" with "zone signing key" instead?

src/sign/mod.rs
bal-e: That validation should be performed by the implementation (and it is, for the backends provided by this module).  I'll make this more clear.

src/sign/bytes.rs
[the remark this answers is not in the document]
ximon18:
```
$ dnssec-keygen -V
dnssec-keygen 9.18.30

$ dnssec-keygen -K /tmp/ -a ECDSAP256SHA256 -f KSK example.com
Generating key pair.
Kexample.com.+013+63913

$ cat /tmp/Kexample.com.+013+63913.key 
; This is a key-signing key, keyid 63913, for example.com.
; Created: 20241104100724 (Mon Nov  4 11:07:24 2024)
; Publish: 20241104100724 (Mon Nov  4 11:07:24 2024)
; Activate: 20241104100724 (Mon Nov  4 11:07:24 2024)
example.com. IN DNSKEY 257 3 13 qDC4ClR00gf7S+rdj1Q9Fl/8MmKc3qaFhF9FbALxcXWYoGViaD4byU5+ FnI3vgn2VMe9HWV6LCIFnpW0pjWaoQ==
```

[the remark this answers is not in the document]
bal-e: Once again -- that is a DNSKEY record showing a *public* key file.  This code in `sign::bytes` is describing the format of *private* key files.  See the `.private` file that `dnssec-keygen` generated.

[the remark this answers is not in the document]
ximon18: Fair point. As the one output file of BIND can have comments, is it maybe tolerating them as well in the other file?

src/sign/mod.rs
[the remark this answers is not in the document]
ximon18: I'm not sure what you are suggesting. If you are suggesting an fn to load from a file on disk I would say no don't do that as we had said we were not going to put file based APIs into domain.

src/sign/mod.rs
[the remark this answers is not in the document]
ximon18: I can't think of anything better. Leave it for now.

src/sign/mod.rs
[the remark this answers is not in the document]
ximon18: I'd be worried about the potential for it becoming out of sync with the underlying implementation.

src/sign/common.rs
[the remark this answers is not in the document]
ximon18: There's not really a good apex point in the docs indeed. I suggested to mention it here because this is where a preference is expressed.

src/sign/mod.rs
ximon18: AFAIK the KSK has to be part of the same zone as the ZSKs that it signs, with the exception that the DS RR in the parent zone "refers to a DNSKEY RR" in the child zone (and includes the entire RDATA of the child zone KSK DNSKEY RR). In the case of a CSK (combined signing key) the key signs itself and the zone RRs. Saying that zone signnig keys sign DNS records is a bit misleading because KSKs do that as well. Maybe just drop the text in parentheses?

src/sign/bytes.rs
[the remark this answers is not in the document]
ximon18: Raw?

src/sign/mod.rs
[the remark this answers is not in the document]
ximon18: I'd rather drop it, but let's leave it for now as it's not wrong, just my 10c as they say.

src/sign/bytes.rs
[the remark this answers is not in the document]
ximon18: Resolving this comment because this is bordering on nit-picking.

src/sign/bytes.rs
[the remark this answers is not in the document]
bal-e: But `RawSecretKey` could also mean that it's part of a cryptographic backend.  `SecretKeyBytes` feels more specific about it just being a bunch of bytes with no cryptographic functionality.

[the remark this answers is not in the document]
bal-e: And I think defining `SecretKeyBytes` in a module called `raw` would be even more confusing.

src/sign/mod.rs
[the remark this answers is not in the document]
bal-e: I think that's unfortunate, but alright.  I'll add an example here.

src/sign/mod.rs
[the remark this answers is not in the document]
bal-e: That's understandable.  I'll fix this.

src/sign/mod.rs
bal-e: Right, removing the additional information is probably the easiest way forward.

ximon18 (approved):

ximon18:
> I've looked at the RustDoc and given some feedback in comments on this PR.
> 
> I have not yet reviewed the actual code in its current form.

This code is getting more eyes on it and being exercised at points of use in other work we are doing so with that combined with previous reviews I've done of this PR I'm fine to merge this.

bal-e:
Awesome, thanks for all the reviews @ximon18!

bal-e:
I'll leave the branch up for a bit because it could mess with other PRs iirc.

ximon18:
> I'll leave the branch up for a bit because it could mess with other PRs iirc.

There was a PR targetting this branch but that now targets `main` so I will delete this PR branch.

PR #413, opened 2024-10-14
Fix serde impls to work with nostd.

partim:
This PR fixes issues that block building the `"serde"´ feature without `"std"`.

Specifically:
* Change the `Serialize` impls of `ParsedName` and `ExtendedError` to not build strings first.
* Added a `Deserialize` impl to `base::net:nostd::Ipv6Addr` and fixed the impl for `Ipv4Addr`.

### DIFF
--- a/src/base/opt/exterr.rs
+++ b/src/base/opt/exterr.rs
@@ -45,12 +45,12 @@ pub struct ExtendedError<Octs> {
     ///
     /// See `text` for the interpretation of the result.
     #[cfg_attr(feature = "serde", serde(serialize_with = "lossy_text"))]
-    text: Option<Result<Str<Octs>, Octs>>,
+    text: Option<Result<Str<Octs>, LossyOctets<Octs>>>,
 }
 
 #[cfg(feature = "serde")]
 fn lossy_text<S, Octs: AsRef<[u8]>>(
-    text: &Option<Result<Str<Octs>, Octs>>,
+    text: &Option<Result<Str<Octs>, LossyOctets<Octs>>>,
     serializer: S,
 ) -> Result<S::Ok, S::Error>
 where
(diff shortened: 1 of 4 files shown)


src/base/opt/exterr.rs
```diff
@@ -117,15 +120,15 @@ impl<Octs> ExtendedError<Octs> {
...
+        self.text.as_ref().map(|res| res.as_ref().map_err(|err| err.as_ref()))
     }
 
     /// Returns the text as an octets slice.
     pub fn text_slice(&self) -> Option<&[u8]>
     where Octs: AsRef<[u8]> {
         match self.text {
             Some(Ok(ref text)) => Some(text.as_slice()),
-            Some(Err(ref text)) => Some(text.as_ref()),
+            Some(Err(ref text)) => Some(text.as_ref().as_ref()),
```
ximon18, line 131: It's not required but adding `LossyOctets::as_slice()` would make this more readable ;-)

partim: And drop the `AsRef<_>` impl instead?

src/base/opt/exterr.rs
```diff
@@ -364,6 +339,59 @@ impl<'a, Target: Composer> OptBuilder<'a, Target> {
     }
 }
 
+
+//------------ LossyOctets ---------------------------------------------------
+
+/// A octets wrapper that displays its content as a lossy UTF-8 sequence.
```
ximon18, line 345: A_n_ octets wrapper/

src/base/opt/exterr.rs
```diff
@@ -364,6 +339,65 @@ impl<'a, Target: Composer> OptBuilder<'a, Target> {
...
+                Err(err) => {
+                    let (good, mut tail) = source.split_at(err.valid_up_to());
+                    f.write_str(
+                        // Safety: valid UTF8 for this part was confirmed
+                        // above.
+                        unsafe {
+                            str::from_utf8_unchecked(good)
+                        }
+                    )?;
+                    f.write_str("\u{fffd}")?;
```
ximon18, line 385: While not caused by this PR, I really would have appreciated a tiny comment to say "Output the unicode <?> replacement character" as I had to Google unicode fffd...

ximon18 (approved):

src/base/opt/exterr.rs
partim: Oh, sorry, missed this comment over the big green “approved“ line.

PR #418, opened 2024-10-17
Enhanced zone signing.

ximon18:
This PR builds upon #416 to extend the signing support in `domain` to support signing with multiple keys and key types using the new key management functionality in #406.


src/sign/records.rs
```diff
@@ -74,34 +77,122 @@ impl<N, D> SortedRecords<N, D> {
...
+            })
+            .collect();
+
+        let mut zsks: Vec<&(SigningKey, validate::Key<Octets>)> = keys
+            .iter()
+            .filter(|(k, _)| !unsupported_algorithms.contains(&k.algorithm()))
+            .filter(|(_, dk)| {
+                dk.is_zone_signing_key() && !dk.is_secure_entry_point()
+            })
+            .collect();
```
bal-e, line 132: `Iterator::partition()` might be a nicer way of separating ZSKs and KSKs.

src/sign/records.rs
```diff
@@ -74,34 +77,122 @@ impl<N, D> SortedRecords<N, D> {
...
     {
-        let mut res = Vec::new();
-        let mut buf = Vec::new();
+        // Per RFC 8624 section 3.1 "DNSSEC Signing" column guidance.
+        let unsupported_algorithms = [
+            SecAlg::RSAMD5,
+            SecAlg::DSA,
+            SecAlg::DSA_NSEC3_SHA1,
+            SecAlg::ECC_GOST,
+        ];
```
bal-e, line 116: I don't think this step is necessary.  Keys with must-not-implement algorithms aren't supported by `SigningKey` anyway.  It is the user's responsibility to select appropriate keys and to be aware of algorithm security.

src/sign/records.rs
```diff
@@ -74,34 +77,122 @@ impl<N, D> SortedRecords<N, D> {
...
-    pub fn sign<Octets, Key, ApexName>(
+    pub fn sign<Octets, SigningKey>(
         &self,
-        apex: &FamilyName<ApexName>,
+        apex: &FamilyName<N>,
         expiration: Timestamp,
         inception: Timestamp,
-        key: Key,
-    ) -> Result<Vec<Record<N, Rrsig<Octets, ApexName>>>, Key::Error>
+        keys: &[(SigningKey, validate::Key<Octets>)], // private, public key pair
```
bal-e, line 99: I'm going to introduce a new type to combine these for you.

src/sign/records.rs
```diff
@@ -74,34 +77,122 @@ impl<N, D> SortedRecords<N, D> {
...
+                dk.is_zone_signing_key() && !dk.is_secure_entry_point()
+            })
+            .collect();
+
+        // CSK?
+        if !ksks.is_empty() && zsks.is_empty() {
+            zsks = ksks.clone();
+        } else if ksks.is_empty() && !zsks.is_empty() {
+            ksks = zsks.clone();
+        }
```
bal-e, line 139: What happens if there are no ZSKs or KSKs?  Shouldn't the function stop entirely?

src/validate.rs
```diff
@@ -240,10 +241,21 @@ impl<Octs: AsRef<[u8]>> Key<Octs> {
...
+                // Ensure there is a single data line in the input.
+                if !rest.trim().is_empty() {
+                    return Err(ParseDnskeyTextError::Misformatted);
+                } else {
+                    line = this_line;
+                    break;
+                }
+            } else {
+                line = rest;
+            }
```
bal-e, line 258: I've updated my PR to account for this better.  Let me know if my new code works.

src/sign/records.rs
```diff
@@ -75,34 +79,134 @@ impl<N, D> SortedRecords<N, D> {
...
+    /// ZSK).
     #[allow(clippy::type_complexity)]
     pub fn sign<Octets, ConcreteSecretKey>(
         &self,
         apex: &FamilyName<N>,
         expiration: Timestamp,
         inception: Timestamp,
-        key: SigningKey<Octets, ConcreteSecretKey>,
-    ) -> Result<Vec<Record<N, Rrsig<Octets, N>>>, ErrorTypeToBeDetermined>
+        keys: &[SigningKey<Octets, ConcreteSecretKey>],
```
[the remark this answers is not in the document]
ximon18: There's a version of this code that takes an iterator and works with a domain ZoneTree, so yes changes to this API are intended at some point, but at the moment I'm just making stupid minimal changes to the API and will return to the topic of API design later.

src/sign/records.rs
```diff
@@ -74,34 +76,108 @@ impl<N, D> SortedRecords<N, D> {
...
+    ///
+    /// Keys with a supported algorithm with the ZONE flag set will be used as
+    /// ZSKs.
+    ///
+    /// Keys with a supported algorithm with the ZONE flag AND the SEP flag
+    /// set will be used as KSKs.
+    ///
+    /// If only one key has a supported algorithm and has the ZONE flag set
+    /// AND has the SEP flag set, it will be used as a CSK (i.e. both KSK and
+    /// ZSK).
```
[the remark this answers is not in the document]
ximon18: Thanks. Note: None of this behaviour / interface is set in stone yet as I'm still working my way through the support needed for all of the `ldns-signzone` arguments and behaviours. Once I'm done with that I'll start reconsidering the API and then update comments to match.

PR #420, opened 2024-10-24
Internal message printer exposed as `display_dig_style`

tertsdiepraam:
For use in `dnsi` and `dnst` (`update` in particular). Heavily based on the dig output in `dnsi`: https://github.com/NLnetLabs/dnsi/blob/main/src/output/dig.rs. I'm not really sure about the error handling here, so feel free to scrutinize.

### DIFF
--- a/src/base/message.rs
+++ b/src/base/message.rs
@@ -13,6 +13,7 @@
 use super::header::{Header, HeaderCounts, HeaderSection};
 use super::iana::{Class, OptRcode, Rcode, Rtype};
 use super::message_builder::{AdditionalBuilder, AnswerBuilder, PushError};
+use super::message_printer::MessagePrinter;
 use super::name::ParsedName;
 use super::opt::{Opt, OptRecord};
 use super::question::Question;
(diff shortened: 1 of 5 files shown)


bal-e (commented):
`MessagePrinter` (and the module) should be renamed to `DigPrinter`, I think.  It can also probably be placed directly inside `display_dig_style()`, but some of the code may become too nested.

It's unfortunate that printing stops the moment an error is discovered, but I don't think there's a way around that.  I'm slightly concerned that users won't notice that they're printing invalid messages -- at least some documentation is necessary on `display_dig_style()`, which mentions the behavior for messages containing parse errors.

tertsdiepraam:
> It's unfortunate that printing stops the moment an error is discovered, but I don't think there's a way around that. I'm slightly concerned that users won't notice that they're printing invalid messages -- at least some documentation is necessary on display_dig_style(), which mentions the behavior for messages containing parse errors.

Note that this is actually not super common, because it would be when a message is incomplete and doesn't have the right number of records. If some record data is invalid then it will continue.

src/base/message.rs
```diff
@@ -665,6 +666,13 @@ impl<Octs: Octets + ?Sized> Message<Octs> {
     }
 }
 
+/// # Printing
+impl<Octs: AsRef<[u8]>> Message<Octs> {
+    pub fn display_dig_style(&self) -> impl core::fmt::Display + '_ {
```
bal-e, line 671: Small thing, but could you add a doc comment here?  It's part of the public API, after all.

tertsdiepraam: Fair enough :)

bal-e (approved):
LGTM!

PR #421, opened 2024-10-27
HINFO display CPU and OS using quoted format

weilence:
This is the behavior of dig:
image


tertsdiepraam (approved):
Looks good to me. This will make some names that includes spaces more readable. Thank you!

PR #423, opened 2024-10-28
Use base16 for DS (de)serialization.

ximon18:
https://rfc-annotations.research.icann.org/rfc4034.html#section-5.3 says _"The Digest MUST be represented as a sequence of case-insensitive hexadecimal digits"_ so use base 16 for the related (de)serialiazation to be consistent with the display format.


tertsdiepraam (approved):
Just making @bal-e's approval official (and I approve too)

Issue #426, opened 2024-10-28
Add support for NAPTR record type


weilence:
I noticed that some DNS record types, such as NAPTR, SPF, and CAA, are not yet supported.
I’ve implemented the parsing for NAPTR to prevent it from appearing as a binary string in outputs.
Could you review the code to ensure it follows the project’s coding standards?

ximon18:
Hi @weilence,

Thankyou very much for this contribution!

At first glance it looks very nice. I have some minor observations which I'll add to the PR.

Ximon

PR #427, opened 2024-10-28
Add support for NAPTR record type

weilence:
#426

### DIFF
--- a/src/rdata/mod.rs
+++ b/src/rdata/mod.rs
@@ -132,4 +132,9 @@ rdata_types! {
             Zonemd<O>,
         }
     }
+    naptr::{
+        zone {
+            Naptr<O, N>,
+        }
+    }
 }
--- a/src/rdata/naptr.rs
+++ b/src/rdata/naptr.rs
@@ -1,0 +1,545 @@
+//! Record data from [RFC 3403]: NAPTR records.
+//!
+//! This RFC defines the Naptr record type.
+//!
+//! [RFC 3403]: https://tools.ietf.org/html/rfc3403
+
+use crate::base::{
+    name::FlattenInto,
+    rdata::ComposeRecordData,
+    scan::{Scan, Scanner},
+    wire::{Compose, Parse, ParseError},
+    zonefile_fmt::{self, Formatter, ZonefileFmt},
+    CanonicalOrd, CharStr, ParseRecordData, ParsedName, RecordData, Rtype,
+    ToName,
+};
+use core::{cmp::Ordering, fmt, hash};
+#[cfg(feature = "serde")]
+use octseq::builder::{EmptyBuilder, FromBuilder, OctetsBuilder};
+use octseq::{Octets, OctetsFrom, OctetsInto, Parser};
+
+//------------ Naptr ---------------------------------------------------------
+
+#[derive(Clone)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(bound(
+        serialize = "
+            Octs: octseq::serde::SerializeOctets + AsRef<[u8]>,
+            Name: serde::Serialize,
+        ",
+        deserialize = "
+            Octs: FromBuilder + octseq::serde::DeserializeOctets<'de>,
+            <Octs as FromBuilder>::Builder:
+                OctetsBuilder + EmptyBuilder
+                + AsRef<[u8]> + AsMut<[u8]>,
+            Name: serde::Deserialize<'de>,
+        ",
+    ))
+)]
+pub struct Naptr<Octs, Name> {
+    order: u16,
+    preference: u16,
+    flags: CharStr<Octs>,
+    services: CharStr<Octs>,
+    regexp: CharStr<Octs>,
+    replacement: Name,
+}
+
+impl Naptr<(), ()> {
+    /// The rtype of this record data type.
+    pub(crate) const RTYPE: Rtype = Rtype::NAPTR;
+}
+
+impl<Octs, Name> Naptr<Octs, Name> {
+    pub fn new(
+        order: u16,
+        preference: u16,
+        flags: CharStr<Octs>,
+        services: CharStr<Octs>,
+        regexp: CharStr<Octs>,
+        replacement: Name,
+    ) -> Self {
+        Naptr {
+            order,
+            preference,
+            flags,
+            services,
+            regexp,
+            replacement,
+        }
+    }
+
+    pub fn order(&self) -> u16 {
+        self.order
+    }
+
+    pub fn preference(&self) -> u16 {
+        self.preference
+    }
+
+    pub fn flags(&self) -> &CharStr<Octs> {
+        &self.flags
+    }
+
+    pub fn services(&self) -> &CharStr<Octs> {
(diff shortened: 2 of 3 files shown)


src/rdata/naptr.rs
```diff
@@ -0,0 +1,545 @@
+//! Record data from [RFC 3403]: NAPTR records.
+//!
+//! This RFC defines the Naptr record type.
+//!
+//! [RFC 3403]: https://tools.ietf.org/html/rfc3403
```
ximon18, line 5: While not consistently done throughout the domain code base yet, in newer code we have been opting to use rfc-editor.org links instead of ietf.org links.

```suggestion
//! [RFC 3403]: https://www.rfc-editor.org/info/rfc3403
```

src/rdata/naptr.rs
```diff
@@ -0,0 +1,545 @@
+//! Record data from [RFC 3403]: NAPTR records.
+//!
+//! This RFC defines the Naptr record type.
```
ximon18, line 3:
```suggestion
//! This RFC defines the NAPTR record type.
```

src/rdata/naptr.rs
```diff
@@ -0,0 +1,545 @@
...
+            order,
+            preference,
+            flags,
+            services,
+            regexp,
+            replacement,
+        }
+    }
+
+    pub fn order(&self) -> u16 {
```
ximon18, line 74: Please add RustDoc comments to each of these public functions, e.g. see `src/rdata/rfc1035/soa.rs`.

src/rdata/naptr.rs
```diff
@@ -0,0 +1,545 @@
...
+        match self.services.canonical_cmp(&other.services) {
+            Ordering::Equal => {}
+            other => return other,
+        }
+        match self.regexp.canonical_cmp(&other.regexp) {
+            Ordering::Equal => {}
+            other => return other,
+        }
+
+        self.replacement.name_cmp(&other.replacement)
```
ximon18, line 291: In other rdata implementations names in `impl CanonicalOrd` are compared using `lowercase_composed_cmp()` so I wonder if that should be the case here as well, e.g.

```suggestion
        self.replacement.lowercase_composed_cmp(&other.replacement)
```

src/rdata/naptr.rs
```diff
@@ -0,0 +1,545 @@
...
+        self.regexp.hash(state);
+        self.replacement.hash(state);
+    }
+}
+
+//--- RecordData, ParseRecordData, ComposeRecordData
+
+impl<Octs, Name> RecordData for Naptr<Octs, Name> {
+    fn rtype(&self) -> Rtype {
+        Rtype::NAPTR
```
ximon18, line 347: In other rdata implementations the style used here is `Naptr::RTYPE` rather than `Rtype::NAPTR`.

```suggestion
        Naptr::RTYPE
```

src/rdata/naptr.rs
```diff
@@ -0,0 +1,545 @@
...
+}
+
+impl<Octs, Name> ComposeRecordData for Naptr<Octs, Name>
+where
+    Octs: AsRef<[u8]>,
+    Name: ToName,
+{
+    fn rdlen(&self, _compress: bool) -> Option<u16> {
+        Some(
+            2 + 2
```
ximon18, line 373: In other rdata implementations there rather than using hard-coded numbers `COMPOSE_LEN` constants are used instead and `checked_add()` is used, e.g. see `fn rdlen` in `src/rdata/dnssec.rs`  so it might be good to do the same here.

src/rdata/naptr.rs
```diff
@@ -0,0 +1,545 @@
...
+    fn compose_rdata<Target: crate::base::wire::Composer + ?Sized>(
+        &self,
+        target: &mut Target,
+    ) -> Result<(), Target::AppendError> {
+        self.order.compose(target)?;
+        self.preference.compose(target)?;
+        self.flags.compose(target)?;
+        self.services.compose(target)?;
+        self.regexp.compose(target)?;
+        self.replacement.compose_canonical(target)
```
ximon18, line 390: Is `compose_canonical()` supposed to be called here?

src/rdata/naptr.rs
```diff
@@ -0,0 +1,545 @@
...
+    }
+
+    fn compose_canonical_rdata<
+        Target: crate::base::wire::Composer + ?Sized,
+    >(
+        &self,
+        target: &mut Target,
+    ) -> Result<(), Target::AppendError> {
+        self.compose_rdata(target)
```
ximon18, line 399: Should `compose_canonical_rdata()` be called here?

src/rdata/naptr.rs
```diff
@@ -0,0 +1,545 @@
...
+            .field("services", &self.services)
+            .field("regexp", &self.regexp)
+            .field("replacement", &self.replacement)
+            .finish()
+    }
+}
+
+//--- ZonefileFmt
+
+impl<Octs, Name> ZonefileFmt for Naptr<Octs, Name>
```
ximon18, line 445: @tertsdiepraam: Can you review this fn please?

src/rdata/mod.rs
```diff
@@ -132,4 +132,9 @@ rdata_types! {
             Zonemd<O>,
         }
     }
+    naptr::{
```
ximon18, line 135: @partim: Should this be ordered alphabetically (with rfc1035 as an exception first) ?

ximon18:
I think it would be good to add a test that regexp escape handling works as expected, e.g. as discussed in https://www.rfc-editor.org/rfc/rfc3403.html#section-4.3, https://www.rfc-editor.org/rfc/rfc3403.html#section-6.1 and https://www.rfc-editor.org/rfc/rfc3403.html#section-7.

src/rdata/naptr.rs
tertsdiepraam: Looks good!

src/rdata/naptr.rs
weilence: I think it should be case-insensitive in comparisons because it's a <domain-name>.

src/rdata/naptr.rs
weilence: I think the processing method should be the same as that of the PTR record.

src/rdata/naptr.rs
weilence: So, the SOA documentation uses the URL from https://tools.ietf.org/html/rfc1035#section-3.3.13 in the file located at https://github.com/NLnetLabs/domain/blob/984d16b6d8f6eb925124b927d58a8c5ddf50d08f/src/rdata/rfc1035/soa.rs#L30. Should I use the https://tools.ietf.org/html/rfc3403 or https://www.rfc-editor.org/rfc/rfc3403.html in doc?

ximon18: We prefer the https://www.rfc-editor.org/rfc/rfc3403 format.

ximon18 (approved):

src/rdata/mod.rs
ximon18: I'm going to re-order this because for me it's like staring at a wonky painting on the wall ;-)

PR #428, opened 2024-10-29
Restore 'Display' behaviour for 'SecAlg' etc

bal-e:
The doc comment for 'int_enum_str_decimal' explicitly states that the 'FromStr' impl expects a decimal number and that the 'Display' impl will also output a decimal number.  However, 'Display' was also adding the mnemonic for the number.


ximon18:
This fixes the issue I encountered, but I can't say if it's a safe change to make to the core of domain, but as you say it restores the behaviour to match the doc strnig. This PR makes more changes than just the one I needed, I don''t know if the change to `fn int_enum_zonefile_fmt_decimal()` is correct or not, I'll leave that to @tertsdiepraam to comment on.

PR #432, opened 2024-11-01
Add missing required features to Cargo.toml for examples

ximon18:
Other examples have similar issues. This PR aims to fix them all.

Note to self: I should investigate why `tokio-stream` needs to be specified explicitly, but for now this at least makes the example easier to run.


tertsdiepraam (approved):
I confirmed that this checks all the examples correctly. Thanks!

PR #434, opened 2024-11-06
Add support for CAA record type

weilence:


weilence:
@ximon18 The Caa `value` field should be displayed as a `CharStr`, but it isn't actually a `CharStr`. I used `CharStr::from_octets_unchecked` in the `Display` and `Debug` implementations, but I’m not sure if this is acceptable or safe.

src/base/charstr.rs
```diff
@@ -380,6 +380,22 @@ impl<Octs: AsRef<[u8]> + ?Sized> CharStr<Octs> {
             .compose(target)?;
         target.append_slice(self.0.as_ref())
     }
+
+    /// Appends the canonical wire format representation to an octets builder.
+    pub fn compose_canonical<Target: OctetsBuilder + ?Sized>(
```
partim, line 385: I think this method produces the same result as `compose` only slower?

I don’t think we need an extra method here – I think using `compose` is actually better since it makes it clear that we don’t do anything special. (This is less important for char strings that don’t anything special, but for domain names it is important to signal that they are not lower cased and for consistency, using the same strategy here seems better to me.)

weilence:
@partim RFC 8659 for CAA mentions three standard tag values: `issue`, `issuewild`, and `iodef`.
So when parsing CAA records, should the parser treat any tag value other than these three as a parsing failure?
Or should it allow custom tags, which might have some internal or special-purpose meaning in certain cases?

partim:
While the RFC doesn’t suggest specifically say that there may be more tags, the existence of the IANA registry (and mentioning special rules for IANA-registered tags) suggests that other tags are allowed.

I would probably not special-case the registered tags and have the tag type just wrap the char string (and only enforce the more relaxed character set of any ASCII alphanumeric). Comparing the string against the tag name should be cheap enough.

src/rdata/caa.rs
```diff
@@ -0,0 +1,649 @@
...
+    pub const RTYPE: Rtype = Rtype::CAA;
+}
+
+impl<Octs> Caa<Octs> {
+    /// Creates a new CAA record data from the flags, tag, and value.
+    pub fn new(flags: u8, tag: CaaTag<Octs>, value: Octs) -> Self {
+        Caa { flags, tag, value }
+    }
+
+    /// If the value is set to "1", the Property is critical.
```
partim: This docstring lacks a lead-in line, e.g., “Returns the flags.” Yes, this is obvious, but we still have one for every function.

src/rdata/caa.rs
```diff
@@ -0,0 +1,649 @@
...
+    /// A CA MUST NOT issue certificates for any FQDN if the
+    /// Relevant RRset for that FQDN contains a CAA critical
+    /// Property for an unknown or unsupported Property Tag.
+    pub fn flags(&self) -> u8 {
+        self.flags
+    }
+
+    /// The Property identifier
```
partim: This should be an active phrase, e.g., “Returns the property identifier.”

The same goes for the other methods. They should all have a docstring at least with the leading line, even the private ones.

PR #435, opened 2024-11-07
Key sizes for DNSSEC keys

bal-e:
Introduce a `.key_size()` method for public key types, which measures how many bits are in a key (usually, this is just the size of the key).  This is useful for noting in e.g. the output of `ldns-keygen`.

I've also thrown in a small bug fix, which ensures that RSA public keys don't have leading zeros when parsing DNSKEY records.

### DIFF
--- a/src/validate.rs
+++ b/src/validate.rs
@@ -97,6 +97,11 @@ impl<Octs> Key<Octs> {
         self.key.algorithm()
     }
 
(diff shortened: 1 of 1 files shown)


ximon18 (approved):
LGTM, I think 🤷‍♂️  Some tests would be nice.

ximon18 (approved):

PR #438, opened 2024-11-08
Check example compilation in the CI workflow.

ximon18:
This PR adds a job to the `CI` workflow which `cargo check`s the crate examples to make sure they don't fail immediately to compile because the set of features specified for them in `Cargo.toml` is incorrect.

As it uses `jq` and `tr` I limited it to running on Linux (maybe it would work on Mac?).

I included all the versions of Rust we use for the main build and test job, maybe that is excessive?

An example failing (correctly) run can be seen here: https://github.com/NLnetLabs/domain/actions/runs/11740551147/job/32707334314#step:4:1 _(which I cancelled as soon as it failed)_

The summary on the workflow run handily shows what failed, e.g.:

afbeelding

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,7 +76,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macOS-latest]
+        os: [ubuntu-latest]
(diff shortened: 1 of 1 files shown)


ximon18:
I'll merge this to `main` so that the checks will run and fail for PR #432 and hopefully start passing as the issues get fixed.

PR #439, opened 2024-11-08
Don't run CI checks twice per push to PR branches.

ximon18:
We currently suffer from this: https://github.com/orgs/community/discussions/26940

This PR hopes to prevent this dual execution of the same checks per PR commit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,7 @@
 name: ci
-on: [push, pull_request]
+# Don't run on pull_request as well otherwise the jobs will be run twice.
+# See: https://github.com/orgs/community/discussions/26940
+on: push
 jobs:
   test:
     name: test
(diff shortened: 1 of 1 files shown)


bal-e:
From what I understand, the `pull_request` version will merge the branch to `main` (or whatever the target branch is) before running the tests.  This is an important distinction from the `push` version, since some bugs are caused by freshly-merged PRs that the PR being tested did not integrate yet.  I ran into this case with the `ZonefileFmt` PR, which changed a `Display` impl I was relying upon at some point.

bal-e:
Looking at the CI failure: Should `unstable-server-transport` enable the `unstable-client-transport` feature?

ximon18:
> Looking at the CI failure: Should `unstable-server-transport` enable the `unstable-client-transport` feature?

I would have expected that to be fixed by #432. Perhaps this PR just needs rebasing against/merging with `main`. I'll check.

ximon18:
True, so we could use `pull_request` instead of `push` but then the CI would not be run for commits to branches that don't yet have a PR associated with them (which is maybe fine, you can always create a draft PR if you're not ready yet but want CI feedback).

bal-e (approved):
LGTM!

partim (approved):

PR #440, opened 2024-11-13
[resolver/stub] Allow adding connections dynamically

bal-e:
This can be used to side-step 'ResolvConf' and the underlying default 'multi_stream' connection type with a custom one, e.g. for testing.  We need this right now in order to support Stelline testing for `ldns-update`.


Philip-NLnetLabs (approved):
Looks good to me

ximon18 (approved):

PR #441, opened 2024-11-14
Derive Arbitrary (via new optional feature) for some types to enable fuzz testing.

ximon18:
Useful for applications that use `domain` types in their own fuzz tests.


tertsdiepraam (approved):
Looks good!

PR #444, opened 2024-11-20
Implement FromStr for zonemd Scheme and Algorithm

mozzieongit:
While writing the dnst signzone zonemd argument parsing, I noticed that `zonemd::Scheme` and `zonemd::Algorithm` implement neither `FromStr` nor a `from_mnemonic` function (as with the enums created with the iana int_enum macro). 
I was unsure which one should be used, but settled on `FromStr` with both number and mnemonic, as it allows an easy use of `str::parse`.

### DIFF
--- a/src/base/iana/zonemd.rs
+++ b/src/base/iana/zonemd.rs
@@ -1,0 +1,50 @@
+//! ZONEMD IANA parameters.
+
+//------------ ZonemdScheme --------------------------------------------------
+
+int_enum! {
+    /// ZONEMD schemes.
+    ///
+    /// This type selects the method by which data is collated and presented
+    /// as input to the hashing function for use with [ZONEMD].
+    ///
+    /// For the currently registered values see the [IANA registration]. This
+    /// type is complete as of 2024-11-29.
+    ///
+    /// [ZONEMD]: ../../../rdata/zonemd/index.html
+    /// [IANA registration]: https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#zonemd-schemes
+    =>
+    ZonemdScheme, u8;
+
+    /// Specifies that the SIMPLE scheme is used.
+    (SIMPLE => 1, "SIMPLE")
+}
+
+int_enum_str_decimal!(ZonemdScheme, u8);
+int_enum_zonefile_fmt_decimal!(ZonemdScheme, "scheme");
+
+//------------ ZonemdAlg -----------------------------------------------------
+
+int_enum! {
+    /// ZONEMD algorithms.
+    ///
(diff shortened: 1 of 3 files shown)


ximon18:
After discussing internally I have reworked this to implement the ZONEMD IANA parameters using the IANA macros as we do for other IANA types and I have labelled this as a breaking change.

By using the macros we get a `FromStr` impl and `to/from_mnemonic()` functions, and consistency with the rest of domain.

This does however mean that types are renamed, moved in the module tree, and the "reserved", "unassigned" and "private" enum variants are gone (as we don't do that for other IANA parameters) and the `ZonefileFmt` impl no longer produces comments for the scheme (as it's a bit of a pain to code cleanly to match the way it was before with "unassigned", "private" and "reserved" ranges and without them there's only a single scheme so the comment has little value, nor does LDNS produce a comment for this type) and algorithm (as we don't output comments for the algorithm on other types).

mozzieongit:
Noting it down, to not forget it: This (and other types in future PRs) should potentially change to use `int_enum_str_with_decimal!` instead of `int_enum_str_decimal`; the difference being that `..._with_decimal` offers `from_str` with a mnemonic AND a number (as a string), while `_str_decimal` only allows using a number.

src/base/iana/zonemd.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// This type selects the algorithm used to hash domain names for use with
+    /// the [ZONEMD].
+    ///
+    /// For the currently registered values see the [IANA registration]. This
+    /// type is complete as of 2024-11-29.
+    ///
+    /// [ZONEMD]: ../../../rdata/zonemd/index.html
+    /// [IANA registration]: https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#zonemd-hash-algorithms
+    =>
+    ZonemdAlg, u8;
```
Philip-NLnetLabs, line 40: Should this be ZonemdAlgorithm?

mozzieongit: I think this was for consistency with the other algorithm type names (SecAlg, DigestAlg, Nsec3HashAlg, ...)

Philip-NLnetLabs: Usually we avoid abbreviation unless something is frequently used. In wonder what is different in this case.

mozzieongit: Do you mean why the other type names are abbreviated in the first place?

Philip-NLnetLabs: I created a PR that renames them all. So we can leave it as is for the moment.

Philip-NLnetLabs (approved):

PR #445, opened 2024-11-22
Support zone file formatting using tabs instead of spaces.

ximon18:
Based on initial internal feedback this is expected to be updated to use a separate `FormatWriter`.

It also currently wrongly uses tabs between record fields AND between rdata values, while rdata values should be padded by spaces.

This PR is about matching current LDNS zone file output format.

It may make more sense for this not to be part of `domain` at all if possible as I can imagine that formatting is an opinionated thing?


ximon18:
Closing in favour of PR #446.

PR #446, opened 2024-11-22
Add TabbedWriter for ZonefileFmt

tertsdiepraam:
An alternative to https://github.com/NLnetLabs/domain/pull/445/files.

I'm not sure about the interface, because the `pretty: bool` doesn't work anymore, and it was already not very good. Maybe these should be separate methods: `display_zonefile`, `display_zonefile_multiline` & display_zonefile_aligned`? Feedback welcome!

### DIFF
--- a/src/base/dig_printer.rs
+++ b/src/base/dig_printer.rs
@@ -2,7 +2,7 @@ use core::fmt;
 
 use crate::rdata::AllRecordData;
 
-use super::zonefile_fmt::ZonefileFmt;
+use super::zonefile_fmt::{DisplayKind, ZonefileFmt};
 use super::ParsedRecord;
 use super::{opt::AllOptData, Message, Rtype};
 
@@ -24,7 +24,7 @@ impl<'a, Octs: AsRef<[u8]>> fmt::Display for DigPrinter<'a, Octs> {
         writeln!(
             f,
             ";; ->>HEADER<<- opcode: {}, rcode: {}, id: {}",
-            header.opcode().display_zonefile(false),
+            header.opcode().display_zonefile(DisplayKind::Simple),
             header.rcode(),
             header.id()
         )?;
@@ -161,7 +161,9 @@ fn write_record_item(
     let parsed = item.to_any_record::<AllRecordData<_, _>>();
 
     match parsed {
-        Ok(item) => writeln!(f, "{}", item.display_zonefile(false)),
+        Ok(item) => {
+            writeln!(f, "{}", item.display_zonefile(DisplayKind::Simple))
+        }
         Err(_) => writeln!(
             f,
             "; {} {} {} {} <invalid data>",
--- a/src/base/zonefile_fmt.rs
+++ b/src/base/zonefile_fmt.rs
@@ -11,21 +11,32 @@ impl From<fmt::Error> for Error {
 
 pub type Result = core::result::Result<(), Error>;
 
+pub enum DisplayKind {
+    Simple,
+    Aligned,
+    Multiline,
+}
+
 pub struct ZoneFileDisplay<'a, T: ?Sized> {
     inner: &'a T,
-    pretty: bool,
+    kind: DisplayKind,
 }
 
 impl<T: ZonefileFmt + ?Sized> fmt::Display for ZoneFileDisplay<'_, T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        if self.pretty {
-            self.inner
-                .fmt(&mut MultiLineWriter::new(f))
-                .map_err(|_| fmt::Error)
-        } else {
-            self.inner
+        match self.kind {
+            DisplayKind::Simple => self
+                .inner
                 .fmt(&mut SimpleWriter::new(f))
-                .map_err(|_| fmt::Error)
+                .map_err(|_| fmt::Error),
+            DisplayKind::Aligned => self
+                .inner
+                .fmt(&mut AlignedWriter::new(f))
+                .map_err(|_| fmt::Error),
+            DisplayKind::Multiline => self
+                .inner
+                .fmt(&mut MultiLineWriter::new(f))
+                .map_err(|_| fmt::Error),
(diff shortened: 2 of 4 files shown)


ximon18:
I am about to review this, but realized that while you have proposed this PR as an alternative to PR #445 you haven't actually said here or there what you think should be different or why.

ximon18:
> I'm not sure about the interface, because the `pretty: bool` doesn't work anymore, and it was already not very good. Maybe these should be separate methods: `display_zonefile`, `display_zonefile_multiline` & display_zonefile_aligned`? Feedback welcome!

I think you are right. What does "pretty" mean anyway? What is pretty to one person may be ugly for another. Perhaps any pretty mode should be a separate formatter with a more descriptive style name than the subjective "pretty".

ximon18:
> > I'm not sure about the interface, because the `pretty: bool` doesn't work anymore, and it was already not very good. Maybe these should be separate methods: `display_zonefile`, `display_zonefile_multiline` & display_zonefile_aligned`? Feedback welcome!
> 
> I think you are right. What does "pretty" mean anyway? What is pretty to one person may be ugly for another. Perhaps any pretty mode should be a separate formatter with a more descriptive style name than the subjective "pretty".

Hmm, I see in this PR that you already have separate styles as a "kind" enum so you have already addressed the issue with the term "pretty". I guess though by using an enum you preclude the ability for users to make their own implementations, this is a closed set extensible only by the crate authors or forking. Would it make sense instead for this to be a trait and the behaviour depends on which trait impl you use?

src/base/dig_printer.rs
```diff
@@ -24,7 +24,7 @@ impl<'a, Octs: AsRef<[u8]>> fmt::Display for DigPrinter<'a, Octs> {
         writeln!(
             f,
             ";; ->>HEADER<<- opcode: {}, rcode: {}, id: {}",
-            header.opcode().display_zonefile(false),
+            header.opcode().display_zonefile(DisplayKind::Simple),
```
ximon18, line 27: This looks really strange to me, but also the original as well: How can one display a zone file file an opcode? I guess it means `header.opcode().display_in_zonefile_presentation_format()` but that would be a bit long ;-)

src/base/zonefile_fmt.rs
```diff
@@ -11,21 +11,32 @@ impl From<fmt::Error> for Error {
 
 pub type Result = core::result::Result<(), Error>;
 
+pub enum DisplayKind {
+    Simple,
```
ximon18, line 15: I think "Simple" has the same problem as the term "pretty" in that it is not clear what it means and what one person considers simple another might not.

src/base/zonefile_fmt.rs
```diff
@@ -11,21 +11,32 @@ impl From<fmt::Error> for Error {
...
-                .map_err(|_| fmt::Error)
+                .map_err(|_| fmt::Error),
+            DisplayKind::Aligned => self
+                .inner
+                .fmt(&mut AlignedWriter::new(f))
+                .map_err(|_| fmt::Error),
+            DisplayKind::Multiline => self
+                .inner
+                .fmt(&mut MultiLineWriter::new(f))
+                .map_err(|_| fmt::Error),
```
ximon18, line 39: Maybe the match arms could just invoke the appropriate `new()` and then after the match do `self.inner.fmt(&mut new_writer).map_err(|_| fmt::Error)`?

src/base/zonefile_fmt.rs
```diff
@@ -128,6 +142,59 @@ impl<W: fmt::Write> FormatWriter for SimpleWriter<W> {
...
+    fn fmt_comment(&mut self, _args: fmt::Arguments<'_>) -> Result {
+        Ok(())
+    }
+
+    fn newline(&mut self) -> Result {
+        self.writer.write_char('\n')?;
+        self.first = true;
+
+        // Little sanity thing, it _should_ already be 0
+        self.blocks = 0;
```
ximon18, line 192: Should this perhaps then be a `dbg_assert()` instead?

src/base/zonefile_fmt.rs
```diff
@@ -338,7 +405,7 @@ mod test {
             more like a silly monkey with a typewriter accidentally writing \
             some shakespeare along the way but it feels like I have to type \
             e\" \"ven longer to hit that limit!\"",
-            record.display_zonefile(false).to_string()
+            record.display_zonefile(DisplayKind::Simple).to_string()
```
ximon18, line 408: Unrelated to this PR I like the text content above hehe, but shouldn't that be "Shakespeare" with a capital S? And sadly: https://edition.cnn.com/2024/11/01/science/monkeys-cannot-type-shakespeare-study-intl-scli-scn/index.html ;-)

ximon18 (approved):
While there may be room for small improvements like "Simple" and "Aligned" naming being unclear, and we need to resolve the discussion about separate fns per style or a trait or whatever, I would still be happy to merge this PR as is because I think this is an improvement over PR #445.

src/base/zonefile_fmt.rs
ximon18: Maybe in RustDoc on this type, or at the module level, we should show examples of the different formats side-by-side to make it easier for a user to compare them?

src/base/zonefile_fmt.rs
```diff
@@ -128,6 +142,59 @@ impl<W: fmt::Write> FormatWriter for SimpleWriter<W> {
     }
 }
 
+/// A simple writer that aligns some bits
```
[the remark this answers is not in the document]
tertsdiepraam: I like `TabbedWriter`!

src/base/zonefile_fmt.rs
tertsdiepraam: Sure!

src/base/zonefile_fmt.rs
tertsdiepraam: I don't think that works without `dyn` unfortunately.

PR #447, opened 2024-11-25
BYO signing sorter impl.

ximon18:
Allows consumers to e.g. use Rayon for faster multi-threaded sorting instead of the default slow single-threaded sorting, without placing a hard-dependency on Rayon in the `domain` crate.

Also changes the NSEC3 logic to no longer insert into a self-sorting collection while collecting NSEC3s, instead it now pushes to an unsorted vec then sorts that before iterating, as post-sorting is much faster.


ximon18:
> I like the idea, but I'm not sure whether this is the right API? I would expect something like `records.sort_with(some_function_to_sort) -> SortedRecords`, where the sorter is not part of the `SortedRecords` type, but this introduces a bunch of additional trait bounds (and we already have so many). Another option to reduce the trait bounds is to implement `extend` on the `Sorter`, so you make a `Sorter` and then let that insert into the `SortedRecords`.

The reason I didn't do that was I found it to be a foot gun, it was too easy to not use the sort in all of the places it should be used, I felt that it's a fundamental characteristic of the instance which sorting algorithm/impl it uses, not just something for one method call.

ximon18:
One issue for example is that `nsec3()` and `sign()` do sorting themselves so there's no opportunity to do the sorting from outside, except to pass the sorter in at multiple points which would be both unwieldy, and feels strange to potentially let you use different sort impls at different stages of using the same object.

Maybe the problem is that the fns should be broken down into pieces, or not be methods on a collection or in some other way that a different design entirely than SortedRecords is called for.

I don't know at this point, I just wanted one place to define the sort impl to use consistently with the SortedRecords type once created.

PR #448, opened 2024-11-25
Report zonefile parsing & signing progress to the caller.

ximon18:
This allows callers to report progress, which is especially useful on very large zone files.

For example usage see https://github.com/NLnetLabs/dnst/pull/35.

I'm not very happy with the `inplace::Zonefile` progress support, but due to the replacement of self via the Bytes `split()` operation, reset of the `start` variable and the iterator being mutably consumed by the caller, using a new position tracking variable and returning it via `Entry` seemed like one easy way of getting the progress information out to the iterating caller code.

Perhaps something like `Iterator::enumerate()` could be used instead to return the progress offset to the iteration loop?

If we keep the extra var returned with `Entry` it should probably be made into named struct fields instead of anonymous tuple fields, to be more self-evident what the `usize` value is. Also, the extra tuple field infects all call sites which is annoying as most don't care about it.

Note: This PR lacks progress notification for NSEC hashing, that should also be added.

Thoughts:
- Progress reporting should be optional, so perhaps the callback should be `Option`.
- Progress reporting has a cost, perhaps it should be emitted only every N units instead of for every unit change?


tertsdiepraam:
@ximon18 this has some conflicts. Do we need to keep this around?

partim:
@ximon18, would you then mind fixing the conflicts and, if you feel the PR is ready for review, remove the draft status?

PR #450, opened 2024-11-27
Attempt to give more context on unknown record type zone file parsing errors.

ximon18:
This came about because I encountered error messages like:

```
Invalid zone file: 58:9: unknown record type
```

Which record type? If I have access to the source file still at the time of inspecting the error then I can determine the RTYPE from the input file, but if the file is gone or was never available due to being XFR or taken from a database or something then the lack of the actual RTYPE is quite annoying.

I suspect this is a breaking change, as I think the zone file parsing was released before we added unstable support for parsing into a zone tree.

Note: This PR doesn't fully extend `StrError` as in my scenario I don't seem to use it, perhaps we should augment that more, and doesn't update any tests.

### DIFF
--- a/src/base/scan.rs
+++ b/src/base/scan.rs
@@ -270,6 +270,9 @@ macro_rules! declare_error_trait {
 
             /// Creates an error when there are trailing tokens.
             fn trailing_tokens() -> Self;
+
+            /// Unknown record type.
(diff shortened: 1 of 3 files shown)


Philip-NLnetLabs (approved):

tertsdiepraam:
This is probably a good idea, but needs to be reimplemented, because the error type changed in the meantime. Presumably as part of new base-related stuff.

PR #457, opened 2024-11-27
FIX: Don't loop forever if an RR CLASS doesn't match the apex CLASS.

### LINKED ISSUE #456: Sign module gets stuck in infinite loop on RFC3597 SOA CLASS
At this point: https://github.com/NLnetLabs/domain/blob/874d44b99b90ee9abe7fea82e2cf7bf722e85585/src/sign/records.rs#L493

ximon18:
Fixes #456 in my local testing.

### DIFF
--- a/src/sign/records.rs
+++ b/src/sign/records.rs
@@ -492,11 +492,10 @@ impl<'a, N, D> RecordsIter<'a, N, D> {
     where
         N: ToName,
     {
-        while let Some(first) = self.slice.first() {
-            if first.class() != apex.class() {
-                continue;
-            }
-            if apex == first || first.owner().ends_with(apex.owner()) {
+        for rr in self.slice {
+            if rr.class() == apex.class()
+                && (apex == rr || rr.owner().ends_with(apex.owner()))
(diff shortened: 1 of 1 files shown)


src/sign/records.rs
```diff
@@ -492,11 +492,10 @@ impl<'a, N, D> RecordsIter<'a, N, D> {
     where
         N: ToName,
     {
-        while let Some(first) = self.slice.first() {
-            if first.class() != apex.class() {
-                continue;
-            }
-            if apex == first || first.owner().ends_with(apex.owner()) {
+        for rr in self.slice {
```
partim, line 495: For-looping on a thing that you are also currently modifying sounds scary and unintuitive. I’m kind of surprised the compiler lets you do this.

Philip-NLnetLabs: In my opinion, signing needs to be changed to not allow multiple classes. There is no need to that feature and it just make the code more complex.

Philip-NLnetLabs:
Seems obsolete.

PR #460, opened 2024-11-29
Apply fixes for new issues found by Clippy 1.83.0

ximon18:
For example there are now many occurrences of "the following explicit lifetimes could be elided".

These new issues are causing our CI builds to fail.

### DIFF
--- a/src/net/client/request.rs
+++ b/src/net/client/request.rs
@@ -78,9 +78,6 @@ pub trait ComposeRequestMulti: Debug + Send + Sync {
     /// Create a message that captures the recorded changes.
     fn to_message(&self) -> Result<Message<Vec<u8>>, Error>;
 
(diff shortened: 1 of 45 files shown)


src/net/client/request.rs
```diff
@@ -80,7 +80,7 @@ pub trait ComposeRequestMulti: Debug + Send + Sync {
 
     /// Create a message that captures the recorded changes and convert to
     /// a Vec.
-
+    ///
```
tertsdiepraam, line 83: These lines should've been removed too I think

tertsdiepraam (changes requested):
I think some of the doc fixes are not correct, but otherwise it's all good if it compiles!

tertsdiepraam (approved):
Fixed the issue I requested a change on

PR #462, opened 2024-12-04
Rewrite `net::client::redundant`

bal-e:
This is a low-priority PR for simplifying the `Redundant` client transport.  I've managed to replace a large and difficult state machine with a much simpler control flow, making good use of the `futures_util` crate.  The module has become significantly smaller and easier to understand.

### DIFF
--- a/examples/client-transports.rs
+++ b/examples/client-transports.rs
@@ -191,19 +191,12 @@ async fn main() {
     drop(request);
 
     // Create a transport connection for redundant connections.
-    let (redun, transp) = redundant::Connection::new();
-
-    // Start the run function on a separate task.
-    let run_fut = transp.run();
-    tokio::spawn(async move {
-        run_fut.await;
-        println!("redundant run terminated");
-    });
+    let redun = redundant::Connection::new();
 
     // Add the previously created transports.
-    redun.add(Box::new(udptcp_conn.clone())).await.unwrap();
-    redun.add(Box::new(tcp_conn.clone())).await.unwrap();
-    redun.add(Box::new(tls_conn.clone())).await.unwrap();
+    redun.add(Box::new(udptcp_conn.clone()));
+    redun.add(Box::new(tcp_conn.clone()));
+    redun.add(Box::new(tls_conn.clone()));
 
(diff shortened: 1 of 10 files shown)


src/net/client/redundant.rs
```diff
@@ -101,697 +99,379 @@ impl Config {
...
+
+        let result = stream::iter(transports)
+            // For each transport, request from it, then wait out its timeout.
+            .scan(Duration::ZERO, |timeout, transport| {
+                let request = request.clone();
+                let timeout = mem::replace(timeout, transport.timeout);
+                tokio::time::sleep(timeout)
+                    .map(|()| Some(transport.transport.request(request)))
+            })
+            // Execute all requests concurrently, as they are produced.
```
Philip-NLnetLabs, line 221: I think this is wrong. A next request starts only after the previous one had enough time to complete.

bal-e: Each transport request is only "produced" after the respective timeout is executed.  The overall flow looks like:

```
transport 1 | timeout | transport 2 | timeout | ...
 \------------------------------------------------- request 1
                         \------------------------- request 2
                                                  \- ...
```

Should I add a comment at the top of this function to make it more explicit?

bal-e: I've added some comments; let me know if they make the code easier to understand.

examples/client-transports.rs
```diff
@@ -191,19 +191,14 @@ async fn main() {
...
-
-    // Start the run function on a separate task.
-    let run_fut = transp.run();
-    tokio::spawn(async move {
-        run_fut.await;
-        println!("redundant run terminated");
-    });
+    let redun = redundant::Connection::<
+        Box<dyn SendRequest<RequestMessage<Vec<u8>>> + Send + Sync>,
+    >::new();
```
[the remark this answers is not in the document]
bal-e: Yeah, that's fair.  I was trying out the fully strongly typed interface, but I'm happy to revert it to always using `Box<dyn SendRequest>`, or to just set `Box<dyn SendRequest>` as the default for that generic parameter.

PR #463, opened 2024-12-05
Add a tab before the RDATA as well as within it, to match LDNS tabbed output format.

ximon18:


src/base/zonefile_fmt.rs
```diff
@@ -145,6 +145,7 @@ impl<W: fmt::Write> FormatWriter for SimpleWriter<W> {
 /// A single line writer that puts tabs between ungrouped tokens
 struct TabbedWriter<W> {
     first: bool,
+    first_block: bool,
```
tertsdiepraam, line 148: Maybe `first_in_block` is a bit more clear?

tertsdiepraam (approved):
I tried a slightly different approach, where I write the separator on `begin_block` and then set `first` to true. But that didn't turn out much better. So this seems fine to me.

src/base/zonefile_fmt.rs
ximon18: But it's not first in any block, it's the start of the first block.

PR #464, opened 2024-12-05
Make key selection more flexible.

ximon18:
E.g. LDNS seems to consider DS and CDS and CDNSKEY resource record types as well as DNSKEY when selecting keys.


ximon18:
I tried that in an earlier attempt but couldn't get it to compile. I guess you'd need Hash as well. I think the issue was lifetimes as the keys would need to be present in multiple collections but cannot be Copy'd or Cloned, and thus have to be stored by reference.

mozzieongit:
That seems unreasonably complicated for now

mozzieongit (approved):

PR #469, opened 2024-12-12
Introduce a new `base` module

bal-e:
This introduces a `new_base` module as a potential candidate for replacing `base`.  While it has a similar structure (although much is left to be filled out), it has three distinguishing features:

1. It builds the foundation for a `derive`-based interface for easily and efficiently parsing (and possibly building) DNS messages.

2. It introduces `RevName`, a more efficient type for domain names (where labels are stored in reverse order).

3. It relies on the `zerocopy` crate to offload basic byte parsing, and to represent big-endian integers where necessary.  This saves a fair amount of parsing code.

There's a lot more work to be done:

- [x] Define some record data types (probably those from RFC1035).
- [ ] Define the OPT record data, and some EDNS option types.
- [ ] Define some common message structures (including a generic fallback).
- [ ] Update existing parts of the codebase to show the PR's benefits.
  - [ ] The client transports (... while remaining compatible?).
  - [ ] The stub resolver.
- [ ] Try defining some derive macros for the parsing/building traits.


bal-e:
Closing in favor of #474.

Issue #468, opened 2024-12-10
Double quote parsing issue for record data?


ximon18:
Given this input line from a zone file:

```
xx.example.     3600    IN      HINFO   "KLH-10" "TOPS-20"
```

When loading and dumping the zone file using `examples/query-zone.rs` the HINFO record is output with escaped quotes:

```
$ cargo run -q --example query-zone --features tracing-subscriber,unstable-zonetree,zonefile -- -v /tmp/t A example.
Reading zone file '/tmp/t'...
Constructing zone...
Inserting zone for example class IN...
Dumping zone example class IN...
...
> xx.example. 3600 IN HINFO "KLH-10\"" "TOPS-20\""
...
```

Wheeas if the same zone file is signed by `ldns-signzone` the signed output includes the `HINFO` record with the same quoting as the input file, i.e. without escaping.

So:
- Is the escaped output valid?
- Is the escaped output expected?
- If not is it a problem with reading the zone?
- Or is it a problem with the dumping of the zone?

Full input file (based on RFC 1035 Appendix A):
```
example.        3600    IN      SOA     ns1.example. bugs.x.w.example. 1081539377 3600 300 3600000 3600
xx.example.     3600    IN      AAAA    2001:db8::f00:baaa
xx.example.     3600    IN      HINFO   "KLH-10" "TOPS-20"
xx.example.     3600    IN      A       192.0.2.10
x.y.w.example.  3600    IN      MX      1 xx.example.
x.w.example.    3600    IN      MX      1 xx.example.
*.w.example.    3600    IN      MX      1 ai.example.
ns2.example.    3600    IN      A       192.0.2.2
ns1.example.    3600    IN      A       192.0.2.1
ns2.b.example.  3600    IN      A       192.0.2.8
ns1.b.example.  3600    IN      A       192.0.2.7
b.example.      3600    IN      NS      ns2.b.example.
b.example.      3600    IN      NS      ns1.b.example.
ai.example.     3600    IN      AAAA    2001:db8::f00:baa9
ai.example.     3600    IN      HINFO   "KLH-10" "ITS"
ai.example.     3600    IN      A       192.0.2.9
ns2.a.example.  3600    IN      A       192.0.2.6
ns1.a.example.  3600    IN      A       192.0.2.5
a.example.      3600    IN      DS      57855 5 1 b6dcd485719adca18e5f3d48a2331627fdd3636b
a.example.      3600    IN      NS      ns2.a.example.
[...]

PR #470, opened 2024-12-13
Handle quoted octets. (#468)

ximon18:
This seems to solve the issue I encountered in #468, but not being very familiar with the zone parsing code I am not at all sure that this is the correct fix. Review needed!


bal-e (approved):
Good catch, @ximon18!

PR #473, opened 2024-12-19
Alg to algorithm

Philip-NLnetLabs:
Rename *Alg to *Algorithm various DNSSEC types.


src/base/opt/algsig.rs
```diff
@@ -377,8 +377,8 @@ impl<'a, Variant, Octs> IntoIterator for &'a Understood<Variant, Octs>
 where
     Octs: AsRef<[u8]> + ?Sized
 {
-    type Item = SecAlg;
-    type IntoIter = SecAlgsIter<'a>;
+    type Item = SecurityAlgorithm;
+    type IntoIter = SecurityAlgorithmsIter<'a>;
```
tertsdiepraam, line 381: Nitpicky, but I would write this in singular form.

tertsdiepraam (commented):
Looks good! I have one tiny suggestion, but this PR doesn't introduce the thing I would change, so it's fine like this too in my opinion.

src/base/opt/algsig.rs
Philip-NLnetLabs: Done

tertsdiepraam (approved):

bal-e:
Merged manually by @Philip-NLnetLabs.

PR #477, opened 2025-01-09
Require single class zonefiles by default, and give context if possible on parsing errors.

ximon18:
Extend the inplace zone parser to detect violations of the RFC 1035 section 5.2 requirement that a zonefile consist of exactly one class, enabled by default, and extend its error reporting to say what the wrong and expected classes were (and also use this contextual error support for other errors where applicable).

It also modifies the zone file parser to reflect the lack of default class specification in RFC 1035 by no longer making IN the default class and adding `set_default_class()` to allow the original behaviour to be re-enabled or overridden.

This is a breaking change as it enables the validation checks by default and removes the previous default IN class.


bal-e (approved):
Looks good @ximon18!

src/zonefile/inplace.rs
```diff
@@ -342,12 +358,27 @@ impl<'a> EntryScanner<'a> {
...
+
+            (Some(class), None) => {
+                self.zonefile.last_class = Some(class);
                 class
             }
-            None => self.zonefile.last_class,
+
+            (None, Some(last_class)) => last_class,
+
+            (None, None) => {
```
Philip-NLnetLabs, line 378: I think this is not correct. There is no default class according to RFC 1035. Do other parsers accept leaving out the class entirely?

ximon18: `ldns-read-zone` will happily read a zone that lacks the class and print it out as having class `IN`.

ximon18: @Philip-NLnetLabs: I updated this PR to raise an error if no last class exists and added the ability to set the last class explicitly. I also added some simple tests.

ximon18: One issue with this is it makes the library harder to use for the most common case, the IN class, and often we argue make it easy to use in the most common case.

PR #478, opened 2025-01-13
Introduce feature unstable-client-cache to make the client cache optional.

Philip-NLnetLabs:
The feature resolv depends on unstable-client-transport and the client cache depends on moka. This in not ideal because currently the resolver does not use the cache.

Introduce the feature unstable-client-cache for the client cache and remove the client cache from unstable-client-transport.

### DIFF
--- a/examples/client-transports.rs
+++ b/examples/client-transports.rs
@@ -5,9 +5,11 @@ use domain::net::client::request::{
     RequestMessage, RequestMessageMulti, SendRequest,
 };
 use domain::net::client::{
-    cache, dgram, dgram_stream, load_balancer, multi_stream, redundant,
+    dgram, dgram_stream, load_balancer, multi_stream, redundant,
     stream,
 };
+#[cfg(feature = "unstable-client-cache")]
+use domain::net::client::cache;
 use std::net::{IpAddr, SocketAddr};
 use std::str::FromStr;
 #[cfg(feature = "unstable-validator")]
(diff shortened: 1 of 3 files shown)


WhyNotHugo (approved):
Works fine when using the `StubResolver` :+1:

examples/client-transports.rs
tertsdiepraam, line 1: In my opinion, this could also just require the feature. Or maybe there should be two examples, one with and one without cache. That would make the examples a bit more approachable.

Philip-NLnetLabs: I created a client cache example. Though if we go down this route there are more examples that could be factored out.

examples/client-cache-transport.rs
```diff
@@ -0,0 +1,85 @@
...
+    // Get the reply
+    println!("Wating for cache reply");
+    let reply = request.get_response().await;
+    println!("Cache reply: {reply:?}");
+
+    // Send the request message again.
+    let mut request = cache.send_request(req.clone());
+
+    // Get the reply
+    println!("Wating for cached reply");
```
tertsdiepraam, line 82:
```suggestion
    println!("Waiting for cached reply");
```

tertsdiepraam (commented):
I like the separate example. Looks good to me apart from the small typos that we might as well fix now.

tertsdiepraam (approved):

Issue #322, opened 2024-05-30
Latest version of domain 0.10.0 is not compatible with triomphe v0.1.12


ravalikakurumillla:
Tried using the latest domain crate `domain = "0.10.0"` 

```
[[package]]
name = "domain"
version = "0.10.0"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "5cd50aea158e9a57c9c9075ca7a3dfa4c08d9a468b405832383876f9df85379b"
dependencies = [
 "bytes",
 "futures-util",
 "moka",
 "octseq",
 "pin-project-lite",
 "rand 0.8.5",
 "smallvec",
 "time",
 "tokio",
 "tracing",
]
```

```
[[package]]
name = "moka"
version = "0.12.7"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "9e0d88686dc561d743b40de8269b26eaf0dc58781bde087b0984646602021d08"
dependencies = [
 "async-lock 3.3.0",
 "async-trait",
 "crossbeam-channel",
 "crossbeam-epoch",
 "crossbeam-utils",
 "event-listener 5.3.1",
 "futures-util",
 "once_cell",
 "parking_lot",
 "quanta",
 "rustc_version",
 "smallvec",
 "tagptr",
 "thiserror",
 "triomphe",
 "uuid",
]
```

```
[[package]]
name = "triomphe"
version = "0.1.12"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "1b2cb4fbb9995eeb36ac86fadf24031ccd58f99d6b4b2d7b911db70bddb80d90"
```

Failing with the below error 
```
error[E0425]: cannot find function `addr_eq` in module `ptr`
   --> /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/triomphe-0.1.12/src/arc.rs:282:14
    |
282 |         ptr::addr_eq(this.ptr(), other.ptr())
    |              ^^^^^^^ not found in `ptr`

   Compiling tagptr v0.2.0
   Compiling httpdate v1.0.3
   Compiling clang-sys v1.8.2
   Compiling aho-corasick v1.1.3
error[E0658]: use of unstable library feature 'pointer_byte_offsets'
   --> /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/triomphe-0.1.12/src/arc.rs:201:33
    |
201 |         let arc_inner_ptr = ptr.byte_sub(offset_of_data);
[...]

PR #480, opened 2025-01-13
Change the MSRV policy and bump to 1.79.0

bal-e:
The MSRV is being bumped so that `domain` can access newer language features.  In particular:

- `<[u8]>::utf8_chunks()` (1.79) is very useful for processing byte strings that will _mostly_ contain valid UTF-8 text.
- `core::net` (1.77) allows us to remove our polyfills of `Ipv4Addr` etc., used when the `std` feature is disabled.
- `async fn` and return-position `impl Trait` in traits (1.75) allow us to define traits using `async fn`, avoiding the overhead of `Box<dyn Future>` for un-nameable future types.
- `Option::is_some_and()` (1.70) simplifies a very common pattern when inspecting optional values.

The MSRV bump policy is also being updated; it sets a maximum MSRV (relative to the latest stable Rust version) and documents the conditions under which a bump will occur.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,9 +2,8 @@
 name = "domain"
 version = "0.10.3"
 
-# This is the mimimal Rust version with minimal dependencies as specified
-# below. With latest dependency versions 1.76.0 is currently required.
-rust-version = "1.68.2"
+# We keep our MSRV 4 versions behind stable (about half a year).
+rust-version = "1.80.1"
 
 edition = "2021"
 authors = ["NLnet Labs <dns-team@nlnetlabs.nl>"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,8 +2,8 @@
 name = "domain"
 version = "0.10.3"
 
-# We keep our MSRV 4 versions behind stable (about half a year).
-rust-version = "1.80.1"
+# The MSRV is at least 4 versions behind stable (about half a year).
+rust-version = "1.79.0"
 
 edition = "2021"
 authors = ["NLnet Labs <dns-team@nlnetlabs.nl>"]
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -2,9 +2,8 @@
 name = "domain"
 version = "0.10.3"
 
-# This is the mimimal Rust version with minimal dependencies as specified
-# below. With latest dependency versions 1.76.0 is currently required.
-rust-version = "1.68.2"
+# We keep our MSRV 4 versions behind stable (about half a year).
```
Philip-NLnetLabs, line 5: I think the 4 versions behind rule is fine if we want to use new features. But if we don't need any features from 1.80.1 then er is no reason to bump it to that version.

bal-e: I agree.  For reference, here's a list of notable features since 1.68.2:
- 1.80: `<[u8]>::trim_ascii()`, `LazyLock`, exclusive ranges in patterns
- 1.79: associated type bounds (`T: Trait<Assoc: Bound>`), better borrow checking, `<[u8]>::utf8_chunks`
- 1.78: custom diagnostic messages
- 1.77: C string literals, recursive `async fn`s, `core::net::{Ipv4Addr,Ipv6Addr}`
- 1.75: `async fn` and return position `impl Trait` in traits, pointer byte offsets
- 1.70: `Option::is_some_and()`

I don't think 1.80 is strictly necessary, but `utf8_chunks` from 1.79 would be quite helpful (for formatting labels and domain names, which are usually -- but not always -- ASCII text).

bal-e: I'd rather keep MSRV bumps separate from the code that's using them, but I absolutely agree that the MSRV bump commit should document why the bump is happening and what features will be used.  I'll add a note about it.

Philip-NLnetLabs (approved):

bal-e:
There has been previous discussion about MSRV bumps at #322.  @vavrusa, I just wanted to check that the proposed MSRV policy (i.e. MSRV is at least 4 versions behind stable, or older) is compatible with your use case.

partim (approved):

PR #486, opened 2025-01-30
Zone updater should work with any record type.

ximon18:
This PR merges into another one because I'm using the stacked work in my nameshed PoC work.

The changes in this PR compared to the base branch are however of use independent of that work and could/should be made into a PR to main, if/once we're happy with these changes.


ximon18:
Replaced by https://github.com/NLnetLabs/domain/pull/535.

PR #488, opened 2025-02-05
The great rename plus cleanup.

Philip-NLnetLabs:
Move DNSSEC modules around. This PR removes the validate module. The validate module becomes dnssec::validator::base.

Remove apex_owner from sign_rrset_in.

Move DNSKEY generation and signing to the application.


src/crypto/common.rs
```diff
@@ -0,0 +1,256 @@
...
+}
+
+//--- Display, Error
+
+impl fmt::Display for FromDnskeyError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(match self {
+            Self::UnsupportedAlgorithm => "unsupported algorithm",
+            Self::UnsupportedProtocol => "unsupported protocol",
+            Self::InvalidKey => "malformed key",
```
ximon18, line 251: I see that this has only been moved so is not new, but looking at it I wonder are "invalid" and "malformed" really the same thing? Something can be well formed but still invalid, right? So maybe it would be better to make these consistent, in which case what is it: invalid or malformed?

src/crypto/common.rs
```diff
@@ -0,0 +1,256 @@
...
+        return Err(AlgorithmError::InvalidData);
+    };
+
+    // Check for minimum supported key size
+    if public_key[pos..].len() < min_len {
+        return Err(AlgorithmError::Unsupported);
+    }
+
+    let (e, n) = public_key[pos..].split_at(exp_len);
+    Ok((e.to_vec(), n.to_vec()))
```
ximon18, line 197: This function appears to be based on the original `src/validate.rs` function `RsaPublicKeyBytes::from_dnskey_format()` but appears to lack the _"Empty values and leading zeros are not allowed."_ checks. Is that intentional? (I haven't worked out yet if it is in effect not possible to have the zero values that the old code checks for, maybe that is why?)

Also, why does it skip over `pos` bytes on line 187 when getting the length of the exponent bytes but then include `pos` in the `pos + exp_len` calculation of the number of exponents bytes to expect?

src/crypto/common.rs
```diff
@@ -0,0 +1,256 @@
...
+        len => (1, usize::from(len)),
+    };
+
+    // Check if there's enough space for exponent and modulus.
+    if public_key[pos..].len() < pos + exp_len {
+        return Err(AlgorithmError::InvalidData);
+    };
+
+    // Check for minimum supported key size
+    if public_key[pos..].len() < min_len {
```
ximon18, line 192: I don't know how inefficient it is, if at all, but you obtain the length of the same slice twice at lines 187 and 192.

src/crypto/openssl.rs
ximon18, line 1: A general comment about the code in `openssl.rs`: Does mapping OpenSSL `ErrorStack` to `AlgorithmError::InvalidData` discard useful error context?

src/crypto/common.rs
ximon18: Ah, correction: I see that `rsa_exponent_modulus()` was taken from the original function by the same name in `src/validate.rs` but I'm still left with the same questions.

src/crypto/openssl.rs
```diff
@@ -0,0 +1,941 @@
...
+
+/// Context for computing a message digest.
+pub struct DigestContext(Hasher);
+
+impl DigestContext {
+    /// Create a new context for a specified digest type.
+    pub fn new(digest_type: DigestType) -> Self {
+        Self(
+            match digest_type {
+                DigestType::Sha1 => Hasher::new(MessageDigest::sha1()),
```
ximon18, line 122: I can't find any trace of `MessageDigest::sha1()` in the original code. Is it intentional to add support for it here in the new code?

src/crypto/openssl.rs
```diff
@@ -0,0 +1,941 @@
...
+        self.0.as_ref()
+    }
+}
+
+//----------- PublicKey ------------------------------------------------------
+
+/// A public key for verifying a signature.
+pub enum PublicKey {
+    /// Default variant, used for RSA.
+    Default(MessageDigest, PKey<Public>, u16),
```
ximon18, line 159: Suggest to rename this variant to `Rsa` as it is confusing reading code such as `verify()` and `dnskey()` when encountering the `Default` variant as it is not clear that it actually refers to RSA, as the comment even says on this variant.

src/crypto/openssl.rs
ximon18, line 1: Another general comment about this file: uses of `.expect("should not fail")` will cause panics with not very helpful messages. and in functions that return `Result` like `from_dnskey()` could instead be handled as an error return. Can (some of) these be converted to error returns instead of expects?

src/crypto/openssl.rs
```diff
@@ -0,0 +1,941 @@
...
+            }
+            SecAlg::ED25519 => {
+                let public_key = PKey::public_key_from_raw_bytes(
+                    dnskey.public_key().as_ref(),
+                    Id::ED25519,
+                )
+                .map_err(|_| AlgorithmError::InvalidData)?;
+                Ok(PublicKey::NoDigest(public_key, dnskey.flags()))
+            }
+            _ => Err(AlgorithmError::Unsupported),
```
ximon18, line 247: The original `src/validate.rs` function `from_dnskey()` could handle ED448 but this function does not. Is that intentional? It seems especially odd given that other functions do handle the ED448 case.

src/crypto/openssl.rs
```diff
@@ -0,0 +1,941 @@
...
+            }
+            PublicKey::NoDigest(public_key, _) => {
+                let mut verifier =
+                    Verifier::new_without_digest(public_key.as_ref())
+                        .map_err(|_| AlgorithmError::InvalidData)?;
+                verifier
+                    .verify_oneshot(signature, signed_data)
+                    .map_err(|_| AlgorithmError::InvalidData)?
+            }
+            PublicKey::EcDsa(digest_algorithm, public_key, _) => {
```
ximon18, line 274: The original `src/validate.rs` function `verify_signed_data()` would raise an unsupported error for the ED448 case but this code does not. Is that intentional?

src/crypto/openssl.rs
```diff
@@ -0,0 +1,941 @@
...
+    }
+
+    /// Convert to a [`Dnskey`].
+    pub fn dnskey(&self) -> Dnskey<Vec<u8>> {
+        match self {
+            PublicKey::Default(message_digest, public_key, flags) => {
+                let alg = if *message_digest == MessageDigest::sha256() {
+                    SecAlg::RSASHA256
+                } else {
+                    unreachable!();
```
ximon18, line 309: The `Default` variant can be constructed for SHA1 and SHA512 as well. Is it intentional that this code will panic if given such a digest?

src/crypto/openssl.rs
```diff
@@ -0,0 +1,941 @@
...
+                if let Ok(exp_len) = u8::try_from(e.len()) {
+                    key.reserve_exact(1 + e.len() + n.len());
+                    key.push(exp_len);
+                } else if let Ok(exp_len) = u16::try_from(e.len()) {
+                    key.reserve_exact(3 + e.len() + n.len());
+                    key.push(0u8);
+                    key.extend(&exp_len.to_be_bytes());
+                } else {
+                    unreachable!(
+                        "RSA exponents are (much) shorter than 64KiB"
```
ximon18, line 327: I'm not sure but I think this error message will be confusing as I assume it will print out what looks like a message that a too short exponent was encountered while actually it wants to say the opposite.

src/crypto/openssl.rs
```diff
@@ -0,0 +1,941 @@
...
+                    unreachable!();
+                };
+                let rsa = public_key.rsa().expect("should not fail");
+                let e = rsa.e().to_vec();
+                let n = rsa.n().to_vec();
+
+                let mut key = Vec::new();
+
+                // Encode the exponent length.
+                if let Ok(exp_len) = u8::try_from(e.len()) {
```
ximon18, line 318: A reference to RFC 3110 section 2. "RSA Public KEY Resource Records" which defines the byte format of the key vec could be useful to readers here.

src/crypto/openssl.rs
```diff
@@ -0,0 +1,941 @@
...
+                key.extend(&*e);
+                key.extend(&*n);
+
+                Dnskey::new(*flags, 3, alg, key).expect("should not fail")
+            }
+            PublicKey::NoDigest(public_key, flags) => {
+                let alg = match public_key.id() {
+                    Id::ED25519 => SecAlg::ED25519,
+                    Id::ED448 => SecAlg::ED448,
+                    _ => todo!(),
```
ximon18, line 340: As no variants other than ED25519 and ED448 are supported elsewhere in this code, perhaps this should be `unreachable!()` instead of `todo!()` ?

src/crypto/openssl.rs
```diff
@@ -0,0 +1,941 @@
...
+                    unreachable!();
+                };
+
+                let key = public_key.public_key();
+                let group = public_key.group();
+                let mut ctx = BigNumContext::new().expect("should not fail");
+                let key = key
+                    .to_bytes(
+                        group,
+                        PointConversionForm::UNCOMPRESSED,
```
ximon18, line 362: Adding a reference here to RFC 6605 section 4 "DNSKEY and RRSIG Resource Records for ECDSA" which defines the byte format of teh key vec as being the "uncompressed form" would be nice here to explain the use of `PointConversionForm::UNCOMPRESSED` in the call to the OpenSSL `to_bytes()` function.

src/crypto/openssl.rs
```diff
@@ -0,0 +1,941 @@
...
+                let (digest_algorithm, min_bytes) = match sec_alg {
+                    SecAlg::RSASHA1 | SecAlg::RSASHA1_NSEC3_SHA1 => {
+                        (MessageDigest::sha1(), 1024 / 8)
+                    }
+                    SecAlg::RSASHA256 => (MessageDigest::sha256(), 1024 / 8),
+                    SecAlg::RSASHA512 => (MessageDigest::sha512(), 1024 / 8),
+                    _ => unreachable!(),
+                };
+
+                // The key isn't available in either PEM or DER, so use the
```
ximon18, line 188: This comment doesn't apply here as no verification is being done. I assume it was copy-pasted from the original verify_signed_data() function.

src/crypto/openssl.rs
```diff
@@ -0,0 +1,941 @@
...
+        /// Import a key pair from bytes into OpenSSL.
+        pub fn from_bytes<Octs>(
+            secret: &SecretKeyBytes,
+            public: &Dnskey<Octs>,
+        ) -> Result<Self, FromBytesError>
+        where
+            Octs: AsRef<[u8]>,
+        {
+            /// Create a [`BigNum`] from a slice.
+            fn num(slice: &[u8]) -> Result<BigNum, FromBytesError> {
```
ximon18, line 429: Why not use `BigNum::from_slice()` instead of `BigNum::new() and BigNum::copy_from_slice()`?

src/crypto/openssl.rs
```diff
@@ -0,0 +1,941 @@
...
+            /// Create a [`BigNum`] from a slice with secure storage.
+            fn secure_num(slice: &[u8]) -> Result<BigNum, FromBytesError> {
+                let mut v = BigNum::new_secure()?;
+                v.copy_from_slice(slice)?;
+                Ok(v)
+            }
+
+            let pkey = match secret {
+                SecretKeyBytes::RsaSha256(s) => {
+                    let n = num(&s.n)?;
```
ximon18, line 444: Why not call `PublicKey::from_dnskey()` here which has this code and also would make it more consistent with the other variants handled here, and could also be used to support for RSASHA1 and RSASHA512?

src/crypto/common.rs
Philip-NLnetLabs: I think this should be rewritten. Especially if we do an ldns-verify-zone that checks signatures. But I think that should be a separate PR.

src/crypto/openssl.rs
Philip-NLnetLabs: NSEC3 uses SHA1...

src/crypto/openssl.rs
Philip-NLnetLabs: Why? If it is not expected to fail, then returning an error only makes things worse. Now the caller has an error that is not supposed to happen. What would the caller do?

ximon18: If you think this can only fail due to major issues that would also affect the parent application then I can agree that panicking is okay.

But if it could fail for some issue specific to OpenSSL then that would not be a reason to abort the parent application.

src/crypto/openssl.rs
Philip-NLnetLabs: Done

src/crypto/openssl.rs
Philip-NLnetLabs: I added SHA1 and SHA512 but there are some limitations for SHA1.

src/dnssec/sign/denial/config.rs
```diff
@@ -0,0 +1,59 @@
...
+where
+    O: AsRef<[u8]> + From<&'static [u8]>,
+{
+    /// The zone already has the necessary NSEC(3) records.
+    AlreadyPresent,
+
+    /// The zone already has NSEC records.
+    Nsec(GenerateNsecConfig),
+
+    /// The zone already has NSEC3 records, possibly more than one set.
```
ximon18, line 29:
```suggestion
    /// The zone already has NSEC3 records.
    ///
    /// Note: While zones can have multiple NSEC3 chains, only the configuraton
    /// for a single chain can be expressed using this type.
```

Cargo.toml
```diff
@@ -72,11 +72,12 @@ zonefile    = ["bytes", "serde", "std"]
 
 # Unstable features
 unstable-client-transport = ["moka", "net", "tracing"]
+unstable-crypto = ["bytes"]
+unstable-crypto-sign = ["dep:secrecy", "unstable-crypto"]
 unstable-server-transport = ["arc-swap", "chrono/clock", "libc", "net", "siphasher", "tracing"]
-unstable-sign = ["std", "dep:secrecy", "dep:smallvec", "dep:serde", "time/formatting", "tracing", "unstable-validate"]
+unstable-sign = ["std", "dep:secrecy", "dep:smallvec", "dep:serde", "time/formatting", "tracing", "unstable-crypto-sign"]
 unstable-stelline = ["tokio/test-util", "tracing", "tracing-subscriber", "tsig", "unstable-client-transport", "unstable-server-transport", "zonefile"]
```
ximon18, line 79: The secrecy dependency is already part of the depended on unstable-crypto-sign dependency and so can be removed.

```suggestion
unstable-sign = ["std", "dep:smallvec", "dep:serde", "time/formatting", "tracing", "unstable-crypto-sign"]
```

src/lib.rs
```diff
@@ -152,16 +149,19 @@
...
+//!   or more feature flags that enable cryptographic backends. This feature
+//!   flag enables all parts of the crypto module.
 //! * `unstable-sign`: basic DNSSEC signing support. This will enable the
-#![cfg_attr(feature = "unstable-sign", doc = "  [sign]")]
-#![cfg_attr(not(feature = "unstable-sign"), doc = "  sign")]
+//!   `dnssec::sign`
 //!   module and requires the `std` feature. In order to actually perform any
 //!   signing, also enable one or more cryptographic backend modules (`ring`
 //!   and `openssl`).
 //! * `unstable-validate`: basic DNSSEC validation support. This enables the
```
ximon18, line 164:
```suggestion
//!   and `openssl`). Enabling this will also enable `unstable-crypto-sign`.
```

src/dnssec/validator/base.rs
```diff
@@ -0,0 +1,746 @@
...
+                    // a user could choose to put a length less than 256 over here.
+                    let exp_len =
+                        u16::from_be_bytes(data[1..3].try_into().unwrap());
+                    (exp_len as usize, 3)
+                };
+                let n = &data[off + exp_len..];
+                Ok(n.len() * 8 - n[0].leading_zeros() as usize)
+            }
+            SecAlg::ECDSAP256SHA256 | SecAlg::ECDSAP384SHA384 => {
+                // ECDSA public keys have two points.
```
ximon18, line 121: The original code said that ECDSA public keys have a marker byte as well and subtracted one from the len to account for this. Is that no longer the case?

src/crypto/ring.rs
```diff
@@ -0,0 +1,796 @@
...
+    ) -> Result<(), AlgorithmError> {
+        match self {
+            PublicKey::Rsa(algorithm, public_key) => {
+                public_key.verify(algorithm, signed_data, signature)
+            }
+            PublicKey::Unparsed(_, public_key) => {
+                public_key.verify(signed_data, signature)
+            }
+        }
+        .map_err(|_| AlgorithmError::BadSig)
```
ximon18, line 232: Why is map_err() here twice?

ximon18 (approved):

src/crypto/openssl.rs
```diff
@@ -0,0 +1,941 @@
...
+                let group = openssl::nid::Nid::X9_62_PRIME256V1;
+                let group = openssl::ec::EcGroup::from_curve_name(group)?;
+                PKey::from_ec_key(openssl::ec::EcKey::generate(&group)?)?
+            }
+            GenerateParams::EcdsaP384Sha384 => {
+                let group = openssl::nid::Nid::SECP384R1;
+                let group = openssl::ec::EcGroup::from_curve_name(group)?;
+                PKey::from_ec_key(openssl::ec::EcKey::generate(&group)?)?
+            }
+            GenerateParams::Ed25519 => PKey::generate_ed25519()?,
```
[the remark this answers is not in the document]
Philip-NLnetLabs: Done

src/crypto/openssl.rs
```diff
@@ -0,0 +1,941 @@
...
+                    k.check_key().map_err(|_| FromBytesError::InvalidKey)?;
+                    PKey::from_ec_key(k)?
+                }
+
+                SecretKeyBytes::EcdsaP384Sha384(s) => {
+                    use openssl::{ec, nid};
+
+                    let group = nid::Nid::SECP384R1;
+                    let group = ec::EcGroup::from_curve_name(group)?;
+                    let n = secure_num(s.expose_secret().as_slice())?;
```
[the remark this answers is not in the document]
Philip-NLnetLabs: Done

src/crypto/ring.rs
```diff
@@ -0,0 +1,796 @@
...
+                        .expect("new should not fail")
+                } else if addr_eq(
+                    *algorithm as *const _,
+                    &signature::ED25519 as *const _,
+                ) {
+                    let alg = SecAlg::ED25519;
+
+                    Dnskey::new(flags, 3, alg, unparsed.as_ref().to_vec())
+                        .expect("new should not fail")
+                } else {
```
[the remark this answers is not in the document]
Philip-NLnetLabs: Ring does not have ED448 (as far is I can tell). So this cannot happen.

src/crypto/openssl.rs
```diff
@@ -0,0 +1,941 @@
...
+                    PKey::from_rsa(key)?
+                }
+
+                SecretKeyBytes::EcdsaP256Sha256(s) => {
+                    use openssl::{ec, nid};
+
+                    let group = nid::Nid::X9_62_PRIME256V1;
+                    let group = ec::EcGroup::from_curve_name(group)?;
+                    let n = secure_num(s.expose_secret().as_slice())?;
+
```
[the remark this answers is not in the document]
Philip-NLnetLabs: Done

src/crypto/openssl.rs
bal-e: While OpenSSL is quite a black box, I'm confident that we are currently only relying on OpenSSL for CPU-local signing (i.e. without any HSMs) and that it will come with all the necessary algorithms.  Failure here is incredibly unlikely -- I had discussed some of these concerns in the documentation for `SignError`.

bal-e (commented):
I haven't reviewed everything yet, but I'm pretty happy with it overall.

src/crypto/common.rs
bal-e: I expect it to have exactly zero performance impact.  We can rewrite it in any form, preferably the most ergonomic / easy to understand one.

src/lib.rs
Philip-NLnetLabs: Done

src/crypto/ring.rs
Philip-NLnetLabs: Done

src/crypto/common.rs
```diff
@@ -0,0 +1,256 @@
...
+    /// [FIPS Secure Hash Standard] Section 6.5.
+    ///
+    /// [FIPS Secure Hash Standard]: http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf
+    Sha384,
+}
+
+//----------- DigestContext --------------------------------------------------
+
+/// Context for computing a message digest.
+pub enum DigestContext {
```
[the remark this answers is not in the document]
Philip-NLnetLabs: Done

src/crypto/common.rs
```diff
@@ -0,0 +1,256 @@
...
+impl AsRef<[u8]> for Digest {
+    fn as_ref(&self) -> &[u8] {
+        match self {
+            #[cfg(feature = "ring")]
+            Digest::Ring(digest) => digest.as_ref(),
+            #[cfg(feature = "openssl")]
+            Digest::Openssl(digest) => digest.as_ref(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
Philip-NLnetLabs: It seems that that would require an extra Vec. No it just gets AsRef from the underlying ring or openssl type.

src/crypto/common.rs
```diff
@@ -0,0 +1,256 @@
...
+    ) -> Result<Self, AlgorithmError> {
+        #[cfg(feature = "ring")]
+        return Ok(Self::Ring(ring::PublicKey::from_dnskey(dnskey)?));
+
+        #[cfg(feature = "openssl")]
+        return Ok(Self::Openssl(openssl::PublicKey::from_dnskey(dnskey)?));
+
+        #[cfg(not(any(feature = "ring", feature = "openssl")))]
+        compile_error!("Either feature \"ring\" or \"openssl\" must be enabled for this crate.");
+    }
```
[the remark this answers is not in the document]
Philip-NLnetLabs: How does that work if both ring and openssl are enabled?

src/crypto/common.rs
```diff
@@ -0,0 +1,256 @@
...
+            #[cfg(feature = "openssl")]
+            PublicKey::Openssl(public_key) => {
+                public_key.verify(signed_data, signature)
+            }
+        }
+    }
+}
+
+/// Return the RSA exponent and modulus components from DNSKEY record data.
+pub fn rsa_exponent_modulus(
```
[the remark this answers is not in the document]
Philip-NLnetLabs: I want to keep crypto self contained and not depend on dnssec. So crypto uses the types from base and dnssec uses crypto.

src/crypto/common.rs
```diff
@@ -0,0 +1,256 @@
...
+    /// Create a new context for a specified digest type.
+    pub fn new(digest_type: DigestType) -> Self {
+        #[cfg(feature = "ring")]
+        return Self::Ring(ring::DigestContext::new(digest_type));
+
+        #[cfg(feature = "openssl")]
+        return Self::Openssl(openssl::DigestContext::new(digest_type));
+
+        #[cfg(not(any(feature = "ring", feature = "openssl")))]
+        compile_error!("Either feature \"ring\" or \"openssl\" must be enabled for this crate.");
```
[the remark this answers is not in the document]
Philip-NLnetLabs: Done

src/crypto/ring.rs
```diff
@@ -0,0 +1,796 @@
...
+//----------- PublicKey ------------------------------------------------------
+
+/// A public key for verifying a signature.
+pub enum PublicKey {
+    /// Variant for RSA public keys.
+    Rsa(&'static RsaParameters, PublicKeyComponents<Vec<u8>>),
+
+    /// Variant for elliptic-curve public keys.
+    Unparsed(
+        &'static dyn VerificationAlgorithm,
```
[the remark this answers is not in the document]
Philip-NLnetLabs: Done

src/crypto/ring.rs
```diff
@@ -0,0 +1,796 @@
...
+                    key.push(exp_len);
+                } else if let Ok(exp_len) = u16::try_from(e.len()) {
+                    key.reserve_exact(3 + e.len() + n.len());
+                    key.push(0u8);
+                    key.extend(&exp_len.to_be_bytes());
+                } else {
+                    unreachable!(
+                        "RSA exponents are (much) shorter than 64KiB"
+                    )
+                }
```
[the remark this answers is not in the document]
Philip-NLnetLabs: I extracted the common RSA encoding code.

src/dnssec/common.rs
```diff
@@ -0,0 +1,361 @@
...
+    let hash = mk_hash(owner, iterations, salt)
+        .map_err(|_| Nsec3HashError::AppendError)?;
+
+    let owner_hash = OwnerHash::from_octets(hash)
+        .map_err(|_| Nsec3HashError::OwnerHashError)?;
+
+    Ok(owner_hash)
+}
+
+//------------ parse_from_bind -----------------------------------------------
```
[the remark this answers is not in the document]
Philip-NLnetLabs: Done

src/dnssec/sign/keys/signingkey.rs
```diff
@@ -1,14 +1,20 @@
...
 //----------- SigningKey -----------------------------------------------------
 
 /// A signing key.
 ///
 /// This associates important metadata with a raw cryptographic secret key.
-pub struct SigningKey<Octs, Inner: SignRaw> {
+#[derive(Debug)]
+pub struct SigningKey<Octs, Inner: SignRaw>
+where
+    Octs: AsRef<[u8]> + Debug,
```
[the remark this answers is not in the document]
Philip-NLnetLabs: It is very convenient to be able to print keys when debugging. I added a comment.

src/crypto/openssl.rs
```diff
@@ -0,0 +1,941 @@
...
+        Ok(KeyPair {
+            algorithm,
+            flags,
+            pkey,
+        })
+    }
+
+    #[cfg(test)]
+    mod tests {
```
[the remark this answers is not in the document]
Philip-NLnetLabs: Yes, the tests were wrong. It is now one module per file.

src/crypto/common.rs
Philip-NLnetLabs: This code is almost identical to rsa_exponent_modulus. I create an issue.

src/crypto/common.rs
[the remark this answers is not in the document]
bal-e: Oh.  Yeah, I didn't think about that.  Whoops!

src/crypto/common.rs
bal-e: I've rewritten this function to be much more comprehensive.

bal-e (approved):

PR #498, opened 2025-03-18
Remove RequestMessageMulti and friends

tertsdiepraam:
This PR removes the distinction between `RequestMessage` and `RequestMessageMulti` and deals with the consequences. The reason is that this distinction leads to code duplication and unused type arguments.

Specifically, this PR makes the following changes:
- Remove `RequestMessageMulti`
- Remove `ComposeRequestMulti`
- Remove the `ReqMulti` type parameter from `Connection` and `ChanReq`
- Remove the `TsigClient` type and always use `ClientSequence` since its behaviour is (at least should be) identical to `ClientTransaction` for a single message.

This is left as a draft for two reasons:
- The documentation needs to be updated.
- There are still the traits `SendRequest` and `SendRequestMulti`, which both have a method `send_request`. These are now more ambiguous because a `RequestMessage` implements both. There are two easy ways to fix this and I'd like to discuss which one is the best:
  - Give them separate names
  - Or make 1 `SendRequest` trait with 2 methods on it, which might be more convenient.

This is a breaking change.


Philip-NLnetLabs:
I don't understand how this should work. A stream should handle multiple response messages for AXFR and IXFR and only a single response message for other requests. I can't find in the code where this is handled.

Then there is also the confusion if the user tries to use a request with AXFR or IXFR on a SendRequest that doesn't support them.

tertsdiepraam:
This changes only the request part. There is still a distinction between `SendMessage` and `SendMessageMulti`, `GetResponse` and `GetResponseMulti`, which do different things for those two cases. But the type for the request can be the same. You are right that this does remove the (runtime) sanity checks for the right Rtypes, but I could add those back in somewhere later. If we do that then I don't think this regresses in any way.

Philip-NLnetLabs:
RequestMessageMulti follows the 'parse, don’t validate' paradigm. With this change we have to check for XFR in many transports of users can get very confusing error if they pass, for example, AXFR or IXFR to dgram_stream.

Can we make RequestMessageMulti the default for stream and that way prevent passing extra parameters?

ximon18:
Disclaimer: I have **NOT** looked at the code in this PR or even read the description and comments above closely.

Historical note: Not that long ago we had an implementation that worked for both single and multi response using a single interface and it was rejected and replaced by the current separate single and multi approach.

I don't disagree with the aim of simplifying the user interface, but we've repeatedly seen that the multi-response/streaming/XFR case is special and supporting it together with the single-response use case negatively influences the usability of the single-response use case which is by far the most commonly used case.

So, if we can make improvements here, great, but we should reflect carefully on why this hasn't been done till now and whether it really is an improvement.

tertsdiepraam:
Not important enough to keep around.

PR #500, opened 2025-03-20
Simplify the implementation of `crypto::common`.

ximon18:
By removing the boilerplate code that just passes through to either Ring or OpenSSL, with the same logic as before that Ring will be used if both Ring and OpenSSL are enabled.

I was triggered to try removing this boiler plate when I saw that all of the impls are just wrappers around either `super::ring` or `super::openssl` functions with identical signatures.

I haven't looked further at the `dnssec-restructure` PR yet, this was the first thing I saw when starting to review it, so maybe this doesn't make sense for some reason I haven't thought of yet or due to something I haven't seen in that PR yet.


ximon18:
Maybe this could even be done just via imports without the type definitions in this PR.

ximon18:
This PR is a bit old now, not critical and probably out-dated. I'll close it.

PR #507, opened 2025-03-31
XFR processing fixes and improvements:

ximon18:
- FIX: Also instruct the caller to delete all records on fallback when fallback from IXFR to AXFR is detected and add a test for this case.
- Return an error if the caller attempts to process records if the end of the XFR has already been reached and add a test for this case.
- Return an error if the first record of the XFR is not a SOA record.
- Convert member fields that don't need to be public into private fields.
- Don't use assert!(it.eq(..)) as it doesn't reveal which element didn't match.


bal-e (approved):
The logic makes sense to me.  The `(update, Option<update>)` style is unfortunate, but I think we can try to address that more thoroughly when we have more time to restructure this properly.

PR #516, opened 2025-04-15
zonefile_fmt: fix handling of tabs when printing only rdata

tertsdiepraam:
Needed for https://github.com/NLnetLabs/dnst/pull/48.

Specifically this fixes how RDATA only zone file fragments are `display`ed, e.g.

```
5414 15 2 DEADBEEF
```

Should not be displayed as:

```
5414\t15 2 DEADBEEF
```

But that is what happens without this fix.

The use case in `dnst` that is affected is generation of a `DS` RR by the `key2ds` subcommand equivalent of the `ldns-key2ds` command.


partim:
I think it is a bug fix not a breaking change.

PR #518, opened 2025-04-24
Implement retrieving dns servers in windows

nagua:
Currently the `ResolvConf` does only read the '/etc/resolv.conf' file to find the active DNS servers. This does not work on windows. In windows you can use the GetAdaptersAddresses win32 api function to find the required information.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -276,6 +276,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "webpki-roots",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


partim:
Thank you for the PR!

First of all, apologies for the ridiculous delay in responding. We’ve been quite busy with Cascade and the unrelated parts of domain fell behind the couch.

While the PR looks good, we feel a bit uncomfortable to add code with quite a bit of unsafe blocks for a system that we are not experienced with and only have limited testing capabilities. I had a quick look at [what Hickory does](https://github.com/hickory-dns/hickory-dns/blob/main/crates/resolver/src/system_conf/windows.rs) and they seem to be using a crate called [ipconfig](https://github.com/liranringel/ipconfig) which provides safe abstractions and, from a cursory look, seems well maintained even if it looks odd on docs.rs.

So, I think I would prefer if we based the server discovery on Windows on that crate. I don’t know if you are interested in converting your PR to that or if you would rather we did the implementation.

Once again, sorry for the delay!

PR #521, opened 2025-05-08
Remove incorrect logic for determining the apex from signing function.

ximon18:
Instead require the apex name to be passed in.

This fixes issues such as signing wrongly identifying an out-of-zone name (such as a glue that sorts earlier than the zone) as the apex instead of the zone itself.

Also update Rust Doc strings to more closely match current behaviour of the signing APIs.

This is a DRAFT PR because it still needs additional tests.


src/dnssec/sign/denial/nsec3.rs
```diff
@@ -470,6 +486,10 @@ where
         last_nent_stack.push(name.clone());
     }
 
+    let Some(nsec3param_ttl) = nsec3param_ttl else {
+        return Err(SigningError::SoaRecordCouldNotBeDetermined);
```
Philip-NLnetLabs, line 490: Is this the wrong error here?

ximon18: No, it can only happen if for some reason the SOA wasn't seen.

Philip-NLnetLabs: Ah, you are right. It is about the TTL.

Philip-NLnetLabs (approved):

PR #527, opened 2025-05-13
FIX: Use an initial absolute owner name if present as the origin.

ximon18:
Thereby supporting parsing of `$ORIGIN`-less zone files also without calling `set_origin()`, for compatibility with LDNS zone parsing.

Includes a unit test.

### DIFF
--- a/src/zonefile/inplace.rs
+++ b/src/zonefile/inplace.rs
@@ -21,7 +21,7 @@ use octseq::str::Str;
 
 use crate::base::charstr::CharStr;
 use crate::base::iana::{Class, Rtype};
-use crate::base::name::{Chain, Name, RelativeName, ToName};
+use crate::base::name::{Chain, FlattenInto, Name, RelativeName, ToName};
 use crate::base::record::Record;
 use crate::base::scan::{
     BadSymbol, ConvertSymbols, EntrySymbol, Scan, Scanner, ScannerError,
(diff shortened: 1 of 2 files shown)


Philip-NLnetLabs:
This seems wrong to me. If it is needed for ldns compatibility then it should be an option.

Philip-NLnetLabs:
I think it is wrong to just take the first owner name in a zone file to be the origin of the zone. If a zone does not have an origin then parsing the zone should just fail for relative names. In many zones a relative name is  a mistake. It would be wrong if we implicitly define an origin for those zones.

ximon18:
Ok. The same reasoning that I made the apex an explicit argument to signing now that I think about it, thanks for pointing that out.Op 14 mei 2025 om 10:45 heeft Philip-NLnetLabs ***@***.***> het volgende geschreven:﻿
Philip-NLnetLabs left a comment (NLnetLabs/domain#527)
I think it is wrong to just take the first owner name in a zone file to be the origin of the zone. If a zone does not have an origin then parsing the zone should just fail for relative names. In many zones a relative name is  a mistake. It would be wrong if we implicitly define an origin for those zones.

ximon18:
I discussed this with @Philip-NLnetLabs and we concluded that rather than put extra origin detection logic into `domain` zone file parsing it's better to force the caller to explicitly define the origin (via the existing `$ORIGIN` and `set_origin()` mechanisms), if their zone file needs it, and in most cases we expect that it is not needed. So this PR should be abandoned.

PR #531, opened 2025-05-21
Keys are small enough that they don't need to be boxed.

Philip-NLnetLabs:
Undo boxing of ring keys trigger by a clippy warning. Keys a re small enough that they don't need to be boxed. If there is a security reason then this needs to be documented and addressed in a separate PR.


tertsdiepraam:
Seems to have been fixed on `main` already.

PR #535, opened 2025-06-02
Generalize `ZoneUpdater` to support any `Record` type, not just `ParsedRecord`.

ximon18:


bal-e (commented):
Looks good, but can we set a default value for the new name parameter so that this isn't (as much of) a breaking change?

PR #541, opened 2025-06-18
Trim leading modulus and public exponent zeroes per RFC 3110 section 2.

ximon18:


src/crypto/common.rs
```diff
@@ -288,3 +300,35 @@ impl fmt::Display for FromDnskeyError {
...
+
+//----------- Tests ----------------------------------------------------------
+
+#[cfg(test)]
+mod tests {
+    use crate::crypto::common::rsa_encode;
+
+    #[test]
+    fn test_rsa_encode() {
+        assert_eq!(rsa_encode(&[], &[]), &[0]);
```
bal-e, line 312: I'm a little surprised that `&[0]` works here -- I expected a compilation error because about one too many references.  But if it works, that's great.

bal-e (approved):

src/crypto/common.rs
```diff
@@ -288,3 +300,35 @@ impl fmt::Display for FromDnskeyError {
...
+        assert_eq!(rsa_encode(&[0, 1, 1], &[0, 0, 1]), &[2, 1, 1, 1]);
+
+        assert_eq!(rsa_encode(&[1, 2], &[]), &[2, 1, 2]);
+        assert_eq!(rsa_encode(&[], &[1, 2]), &[0, 1, 2]);
+        assert_eq!(rsa_encode(&[1, 2], &[1, 2]), &[2, 1, 2, 1, 2]);
+
+        assert_eq!(rsa_encode(&[0, 1, 2], &[1]), &[2, 1, 2, 1]);
+        assert_eq!(rsa_encode(&[1], &[0, 1, 2]), &[1, 1, 1, 2]);
+        assert_eq!(rsa_encode(&[0, 1, 2], &[0, 1, 2]), &[2, 1, 2, 1, 2]);
+    }
```
ximon18, line 333: Hmm, I didn't test the all zeroes case.

Philip-NLnetLabs: I guess we can assume that keys are at least 512 bits big. If something weird happens if something is zero then that is fine.

PR #543, opened 2025-06-24
Disallow elided lifetimes in paths

bal-e:
This language feature is deprecated, but the lint for it is not enabled by default.  Enabling the lint and fixing all occurrences also resolves some new Clippy lints.  This does not affect the API or behaviour of the code in any way.


bal-e:
Some files underwent formatting that `cargo fmt` (even in CI) had not previously picked up for us.  This appears to be because of a `rustfmt` bug where `mod` statements created by macro expansions aren't detected as files to format.  We'll fix this (and open an issue upstream) shortly.

PR #548, opened 2025-06-27
Add a to_system_time method to Timestamp.

Philip-NLnetLabs:
Add a method that converts a Timestamp into a SystemTime given a SystemTime references. This can be used for sorting.

### DIFF
--- a/src/rdata/dnssec.rs
+++ b/src/rdata/dnssec.rs
@@ -707,16 +707,19 @@ impl Timestamp {
         self.0.into_int()
     }
 
-    /// Return a SystemTime that meets two requirements:
-    /// 1) The SystemTime value has a duration since UNIX_EPOCH that
-    ///    modulo 2**32 is equal to our Timestamp value.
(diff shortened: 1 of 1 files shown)


src/rdata/dnssec.rs
```diff
@@ -704,6 +706,48 @@ impl Timestamp {
...
+    /// Return a SystemTime that meets two requirements:
+    /// 1) The SystemTime value has a duration since UNIX_EPOCH that
+    ///    modulo 2**32 is equal to our Timestamp value.
+    /// 2) The time difference between the SystemTime value and the
+    ///    reference time fits in an i32.
+    ///
+    /// This can be used to sort Timestamp values.
+    #[must_use]
+    #[cfg(feature = "std")]
+    pub fn to_system_time(&self, reference: SystemTime) -> SystemTime {
```
tertsdiepraam: @partim says that this should take `self` by value since the type is copy.

PR #553, opened 2025-06-30
Overhaul feature flags for DNSSEC / crypto

bal-e:
- `ring` and `openssl` now enable a shared `unstable-crypto-backend` feature, which is used internally to test whether a common backend is available.

- Examples / doc tests in `crypto`, relying on `crypto::common`, have been moved to the submodule to avoid needing more `cfg` magic.

- `unstable-crypto` now enables `std`; this requirement was previously undetected (and the CI will be adjusted to try to catch more of these over time), but compilation would fail without it.

- `unstable-sign` and `unstable-validator` now fail to compile if `ring` and/or `openssl` are not enabled.  Previously, those modules would remain configured out.  Its status as a breaking change is debatable, but in any case it only affects unstable features.


partim:
I’ll leave the name as `unstable-crypto-backend` for now because that keeps it in line with the other `unstable-crypto-*` features that use it. We should rename it when we de-unstable the crypto stuff.

partim (approved):

PR #568, opened 2025-09-02
Fix dropped responses when DNS stream encounters EOF

TheJokr:
The `stream::Transport` loop did not account for responses that had been
added to the reply channel prior to `reader_fut` returning. If a server
sent a response and immediately closed the stream afterwards, that
response was lost. I fixed the bug by adding an explicit check for any
remaining items from `reply_receiver` in the `reader_fut` select branch.

Additionally, the `Transport::error` function did not await the futures
returned by `ReplySender::send`. This meant the errors were never sent
and instead the channel closed without any message. I fixed that by
making `Transort::error` an async fn.

The PR also contains a regression test. Running it without the
fix fails, but with the fixes it passes. I did not use stelline for this
test as it requires a very specific order of IO events on the
connection.

### DIFF
--- a/src/net/client/stream.rs
+++ b/src/net/client/stream.rs
@@ -630,7 +630,8 @@ where
                             Self::error(
                                 Error::StreamReadTimeout,
                                 &mut query_vec,
-                            );
+                            )
+                            .await;
                             status.state = ConnState::ReadTimeout;
                             break;
                         }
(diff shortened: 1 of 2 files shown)


Philip-NLnetLabs:
Thanks. I'll take a look in a few days.

TheJokr:
Hi @Philip-NLnetLabs, any progress on this? The bug blocks us from upgrading past version 0.9, which has a future-incompatibility warning with the latest stable rustc.

PR #569, opened 2025-09-04
Implement support for various record types

### LINKED ISSUE #451: Add support for SSHFP records in zone parsing.
Currently attempting to parse a zone file using the `inplace` parser causes an `unknown record type` error.

### LINKED ISSUE #452: Add support for TLSA records in zonefile parsing.
Currently attempting to parse a zone file containing a TLSA record using the `inplace` parser causes an `unknown record type` error.

### LINKED ISSUE #453: Add support for OPENPGPKEY records in zonefile parsing.
Currently attempting to parse a zone file containing a OPENPGPKEY record using the `inplace` parser causes an `unknown record type` error.

### LINKED ISSUE #454: Add support for IPSECKEY records in zonefile parsing.
Currently attempting to parse a zone file containing a IPSECKEY record using the `inplace` parser causes an `unknown record type` error.

mozzieongit:
TODO:
- ~~[ ] Figure out how to deal with `@` in HTTPS/SVCB target name ([cascade#190](https://github.com/nlnetlabs/cascade/issues/190))~~
- document (in code and where applicable) that parsing the @ symbol in SVCB/HTTPS RRs is not implemented (TODO moved to Issue #595)

Added types:
- [x] TLSA
- [x] OPENPGPKEY
- [x] SSHFP
- [x] IPSECKEY

Further planned:
- [x] SVCB (needed for nlnetlabs.nl)
- [x] HTTPS (because it's basically SVCB)
- CAA [RFC 8659](https://www.rfc-editor.org/rfc/rfc8659.html#name-syntax) (postponed)
- SMIMEA (because it's basically TLSA) [RFC 8162](https://datatracker.ietf.org/doc/html/rfc8162#page-4) (postponed)

There _was_ a FIXME in `src/base/iana/sshfp.rs`: "These types don't actually have a mnemonic, only a description."
- [x] How should we handle that? Use the mnemonics as they are in code right now, create a variant of inet_enum that doesn't have mnemonics, or remove the types all together?

I based the code on the `Zonemd` types, and don't know if there are functions that could be removed or should be added.

I tried adding a test in `test-data/zonefiles/` but get the error that de-serialization of bytes is not supported in YAML.
- [x] How should I add tests for the types, if at all?

Also, there are no (too short) length checks in the parser as I didn't find a direct answer in the RFCs, and didn't want to think too much about what a minimum would be given the possible parameters.

Fixes: #451
Fixes: #452
Fixes: #453
Fixes: #454


src/rdata/svcb/value.rs
```diff
@@ -608,6 +697,45 @@ impl<Octs: AsRef<[u8]>> Alpn<Octs> {
...
+                return Err(S::Error::custom(
+                    "expected at least one alpn-id value",
+                ));
+            }
+            Ok(Some(Self::from_octets(tmp.freeze()).map_err(|_| {
+                S::Error::custom("invalid svc param value for alpn")
+            })?))
+        } else {
+            // TODO: why is it ok none if the key is wrong? (stolen from parse_value)
+            Ok(None)
```
mozzieongit, line 734: @partim I copied this behavior from parse_value. What was the reason to do this Result<Option<>> and check the key thing?

src/rdata/svcb/params.rs
```diff
@@ -186,6 +194,132 @@ impl<Octs: AsRef<[u8]>> SvcParams<Octs> {
...
+                .compose(&mut builder)
+                .map_err(|_| S::Error::short_buf())?;
+
+            // - an octet string of this length whose contents are the
+            // SvcParamValue in a format determined by the SvcParamKey.
+            param_value
+                .compose_value(&mut builder)
+                .map_err(|_| S::Error::short_buf())?;
+
+            // TODO: [...] Other automatically mandatory keys SHOULD NOT
```
Philip-NLnetLabs: In general the presentation format parser cannot know which keys are automatically mandatory. Except for the Mandatory key which is always automatically mandatory.

src/rdata/sshfp.rs
```diff
@@ -0,0 +1,352 @@
...
+        fmt::Display::fmt(self, f)?;
+        f.write_str(")")
+    }
+}
+
+impl<Octs: AsRef<[u8]>> ZonefileFmt for Sshfp<Octs> {
+    fn fmt(&self, p: &mut impl Formatter) -> zonefile_fmt::Result {
+        p.block(|p| {
+            p.write_token(self.algorithm)?;
+            p.write_show(self.fingerprint_type)?;
```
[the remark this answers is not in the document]
mozzieongit: Not sure, probably a left over from copying zonemd. I'll change the show to token.

src/rdata/ipseckey.rs
```diff
@@ -0,0 +1,935 @@
...
+impl Ipseckey<(), ()> {
+    /// The rtype of this record data type.
+    pub(crate) const RTYPE: Rtype = Rtype::IPSECKEY;
+}
+
+impl<Octs, N> Ipseckey<Octs, N> {
+    /// Create a Ipseckey record data from provided parameters.
+    pub fn new(
+        precedence: u8,
+        gateway_type: IpseckeyGatewayType,
```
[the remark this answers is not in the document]
mozzieongit: My initial thinking was to allow the user to do whatever they like, but now I think `domain` is designed to be easily correct for users, so yes, removing the type parameter is a good idea.

src/rdata/svcb/params.rs
mozzieongit: You are correct, but the TODO is in the context of parsing as SVCB and HTTPS, and when being parsed as an HTTPS RR additional mandatory keys are known.

Philip-NLnetLabs (approved):

PR #571, opened 2025-09-17
Copy LICENSE file into domain-macros crate

pemensik:
Separate crate needs at least one license file in the archive too.


pemensik:
At least minimalist README.md pointing it to primary domain crate would be nice too, but license would be needed for separate crate packages.

PR #576, opened 2025-09-29
Make RelativeName::from_slice const fn

WhyNotHugo:
Allow defining relative names which are validated resolved at compile time. This is mostly useful for infallible functions which need to use a RelativeName which is not taken as input. 

This is also useful for traits where each implementation needs to define
a specific RelativeName. E.g.:

    trait DiscoverableService {
        const RELATIVE_NAME: RelativeName<u8>;
    }

Example (also included in test module):

    const VALID_NAME: &RelativeName<[u8]> =
        match RelativeName::from_slice(b"\x03www\x07example") {
            Ok(name) => name,
            Err(_) => panic!("VALID_NAME failed at compile time"),
        };

A few other functions in the call tree of from_slice had to be made const fn too, some with small changes (like `match` incited of `?`, or inlining `.into()` calls for errors).

### DIFF
--- a/src/base/name/label.rs
+++ b/src/base/name/label.rs
@@ -51,7 +51,7 @@ impl Label {
     /// # Safety
     ///
     /// The `slice` must be at most 63 octets long.
-    pub(super) unsafe fn from_slice_unchecked(slice: &[u8]) -> &Self {
+    pub(super) const unsafe fn from_slice_unchecked(slice: &[u8]) -> &Self {
         // SAFETY: Label has repr(transparent)
         mem::transmute(slice)
     }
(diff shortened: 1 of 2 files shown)


WhyNotHugo:
This is also usable for patterns like:


```rs
  pub struct DiscoverableService {
      pub relative_domain: &'static RelativeName<[u8]>,
      pub scheme: Scheme,
      pub well_known_path: &'static str,
      pub default_port: u16,
      pub access_field: &'static str,
      pub display_name: &'static str,
  }

  impl DiscoverableService {
      pub const CALDAVS: Self = Self {
          relative_domain: match RelativeName::from_slice(b"\x08_caldavs\x04_tcp") { … },
          scheme: Scheme::HTTPS,
          well_known_path: "/.well-known/caldav",
          default_port: 443,
          access_field: "calendar-access",
          display_name: "caldavs",
      };
..
```

partim (approved):

partim:
Apologies for the ridiculous delay. Thank you for the PR – it is looking good!

PR #577, opened 2025-10-14
HACK: Serve the original TTL of the RRSIG via XFR, not the RRSIG RRSET TTL.

ximon18:
As RRSIGs should not be grouped into RRSETs with a common TTL.

### DIFF
--- a/src/net/server/middleware/xfr/responder.rs
+++ b/src/net/server/middleware/xfr/responder.rs
@@ -102,8 +102,13 @@ where
             for rr in rrset.data() {
                 last_rr_rtype = Some(rr.rtype());
 
+                let ttl = match rr {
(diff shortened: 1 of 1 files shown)


partim:
What was the reason to use the RRset’s TTL rather than the RR’s TTL? I feel like the code should not quietly manipulate things but send out what it is given?

ximon18:
The terminology (maybe the code also, I haven't looked at it) is just wrong here anyway, as RFC 4035 states:

> RRSIG RRs, unlike all other DNS RR types, do not form RRsets.

From a Cascade perspective the code in this PR is no longer used as Cascade no longer uses the patches-for-nameshed-prototype branch.

So, if it's questionable if this is ready for domain nor is it needed for Cascade, and as we intend to revisit the zone tree and net server code, perhaps we should just close this PR.

partim:
Ack. I’ll close the PR for now.

PR #578, opened 2025-10-16
Don't emit glue records while walking a zone.

ximon18:
Glue records are not actually part of the zone, they are just internal metadata used to help answer queries.

This causes unexpected additional records to be emitted by XFR out. See https://github.com/NLnetLabs/cascade/issues/222 for some context.

Note: Curretntly this PR targets the `patches-for-nameshed-prototype` branch but it should probably be merged into `main`, and then included in the patches branch afterwards (although we need to take whatever is good from the patches branch and get rid of it ASAP).


mozzieongit:
Speaking in the context of Cascade: I think glue is part of the zone in the sense that it needs to end up at the public name servers (fetched via XFR from the publication server). Does this change affect that capability or is the change only relevant for signing?

Philip-NLnetLabs:
In the ideal case, a zone has glue only for nameservers that are in the child zone. In that case duplicates are not possible because the nameservers are at the registry level associated with zone. Registries that allow sibling glue do have all kind of interesting issues.

partim:
This discussion makes it sound to me like this PR should be rewritten?

ximon18:
> This discussion makes it sound to me like this PR should be rewritten?

From a Cascade perspective the code in this PR is no longer used as Cascade no longer uses the `patches-for-nameshed-prototype` branch.

It seems to me that the current zone tree code in `domain` cannot support both (a) dynamically inserted glue to be included in the additional section of a response, and (b) accurately replicating a zone received via XFR.

So, this is neither ready for `domain` in general nor needed for Cascade. As we intend to revisit the zone tree data types in `domain` perhaps we should just close this PR.

partim:
I think the zone tree will have a full rewrite – we also have to have a discussion what exactly the zone tree in _domain_ will be for. I believe applications like Cascade will likely have to have their own implementation due to very specific requirements.

I’ll close the PR for now.

PR #597, opened 2025-11-21
Copy of example in crypto/mod.rs.

Philip-NLnetLabs:


partim:
Why do you want a copy of this as a separate example?

Philip-NLnetLabs:
This was just to test some CI thing. I'll deleted the PR.

PR #603, opened 2025-12-12
make tsig::RequestMessage implement Clone

tertsdiepraam:
This allows it to be used with `dgram_stream`


partim (approved):
I think this is okay – the alternative is to see if we can remove the `Clone` bounds on `dgram_stream` which is perhaps something for newbase?

PR #616, opened 2026-02-24
Use `jiff` instead of `time`

bal-e:
`jiff` is a popular time library which is taking over the ecosystem. It's easier to use (as you can see from the diff) and it doesn't pull in any dependencies. We're using `jiff` in other places too (e.g. Cascade), so this change helps clean up our dependency tree.

I also wanted to remove `chrono`, but we offer conversion functions with `chrono` types as part of our public stable API (under the `chrono` feature flag). I did notice, however, that `chrono/clock` was being enabled by `unstable-server-transport`, but it isn't used anywhere. So I've also removed that.


WhyNotHugo:
> I also wanted to remove chrono, but we offer conversion functions with chrono types as part of our public stable API

Perhaps you can expose a timestamp, and callers can use [chrono::DateTime::from_timestamp](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.from_timestamp).

partim:
Given that chrono is an optional dependency, I think it is okay to keep it for now.

But we should probably add `impl From<jiff::Timestamp> for crate::base::Serial` as part of this PR so people can already fully switch away from chrono.

partim (approved):

tertsdiepraam (approved):

PR #618, opened 2026-02-28
Remove unwrap() FoundSrvs::merge

WhyNotHugo:
This produces a clippy false positive but is also a bit tricky to follow. Simplify the logic.

See: https://github.com/NLnetLabs/domain/pull/602


WhyNotHugo:
v2: formatting

partim (approved):
Thank you for the PR! That’s one “XXX“ dealt with!

PR #623, opened 2026-03-20
Expand record type support in `domain::new`

bal-e:
This PR resolves various minor limitations of `domain::new`, as experienced in developing Cascade. In particular, it adds the `ZoneMD` and `Rp` record types, with extensive documentation, and adds a number of useful RTYPEs and QTYPEs (even for record types that have not been defined yet).

### DIFF
--- a/macros/src/impls.rs
+++ b/macros/src/impls.rs
@@ -57,7 +57,12 @@ impl ImplSkeleton {
                 }
 
                 GenericParam::Type(value) => {
-                    types.push(value.clone());
+                    types.push(TypeParam {
+                        // Defaults are not allowed in 'impl' blocks.
+                        eq_token: None,
+                        default: None,
+                        ..value.clone()
+                    });
                     let id = value.ident.clone();
                     let id = syn::TypePath {
                         qself: None,
--- a/src/new/base/question.rs
+++ b/src/new/base/question.rs
@@ -182,11 +182,72 @@ impl QType {
     /// The type of queries for [`Txt`](crate::new::rdata::Txt) records.
     pub const TXT: Self = Self::new(16);
 
+    /// The type of queries for [`Rp`](crate::new::rdata::Rp) records.
+    pub const RP: Self = Self::new(17);
+
     /// The type of queries for [`Aaaa`](crate::new::rdata::Aaaa) records.
     pub const AAAA: Self = Self::new(28);
 
+    /// The type of queries for [`DName`](crate::new::rdata::DName) records.
+    pub const DNAME: Self = Self::new(39);
+
+    /// The type of queries for [`Opt`](crate::new::rdata::Opt) records.
+    pub const OPT: Self = Self::new(41);
+
+    /// The type of queries for [`Ds`](crate::new::rdata::Ds) records.
+    pub const DS: Self = Self::new(43);
+
+    /// The type of queries for [`RRSig`](crate::new::rdata::RRSig) records.
+    pub const RRSIG: Self = Self::new(46);
+
+    /// The type of queries for [`NSec`](crate::new::rdata::NSec) records.
+    pub const NSEC: Self = Self::new(47);
+
+    /// The type of queries for [`DNSKey`](crate::new::rdata::DNSKey) records.
+    pub const DNSKEY: Self = Self::new(48);
+
+    /// The type of queries for [`NSec3`](crate::new::rdata::NSec3) records.
+    pub const NSEC3: Self = Self::new(50);
+
+    /// The type of queries for [`NSec3Param`](crate::new::rdata::NSec3Param) records.
+    pub const NSEC3PARAM: Self = Self::new(51);
+
+    /// The type of querios for `Cds` records.
+    pub const CDS: Self = Self::new(59);
+
+    /// The type of querios for `CDNSKey` records.
+    pub const CDNSKEY: Self = Self::new(60);
+
+    /// The type of querios for [`ZoneMD`](crate::new::rdata::ZoneMD) records.
+    pub const ZONEMD: Self = Self::new(63);
+
+    /// The type of querios for `TSig` records.
+    pub const TSIG: Self = Self::new(250);
+
+    /// The type of requests for incremental zone transfers (IXFRs).
+    pub const IXFR: Self = Self::new(251);
+
+    /// The type of requests for authoritative zone transfers (AXFRs).
+    pub const AXFR: Self = Self::new(252);
+
     /// The type of queries for all available records.
     pub const ANY: Self = Self::new(255);
+}
+
+//--- Conversion to and from 'u16'
+
+impl From<u16> for QType {
+    fn from(value: u16) -> Self {
+        Self {
+            code: U16::new(value),
+        }
+    }
+}
+
+impl From<QType> for u16 {
+    fn from(value: QType) -> Self {
+        value.code.get()
+    }
 }
 
 //--- Formatting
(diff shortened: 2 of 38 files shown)


macros/src/impls.rs
```diff
@@ -57,7 +57,12 @@ impl ImplSkeleton {
                 }
 
                 GenericParam::Type(value) => {
-                    types.push(value.clone());
+                    types.push(TypeParam {
+                        // Defaults are not allowed in 'impl' blocks.
+                        eq_token: None,
+                        default: None,
+                        ..value.clone()
+                    });
```
mozzieongit, line 65: I don't understand what this means (due to my lack of knowledge about syn/proc-macros), so can't tell if it is good or not.

src/new/base/question.rs
```diff
@@ -182,13 +182,71 @@ impl QType {
...
+
     /// The type of queries for [`Aaaa`](crate::new::rdata::Aaaa) records.
     pub const AAAA: Self = Self::new(28);
 
+    /// The type of queries for [`Opt`](crate::new::rdata::Opt) records.
+    pub const OPT: Self = Self::new(41);
+
+    /// The type of queries for [`Ds`](crate::new::rdata::Ds) records.
+    pub const DS: Self = Self::new(43);
+
```
mozzieongit, line 196: Do we have a tracking issue of types that are not yet implemented by new base, but are by old base/planned? (I'm thinking of SSHFP for example, which we use in our zone)

src/new/rdata/zonemd.rs
```diff
@@ -0,0 +1,421 @@
...
+/// unique functionality over other means of checking and authenticating DNS
+/// zones; when DNSSEC-signed, it asserts the authenticity of glue records
+/// present with the zone. This is especially important for the root zone.
+///
+/// [RFC 8976]: https://www.rfc-editor.org/rfc/rfc8976.html
+///
+/// ## Computation
+///
+/// [`ZoneMD`] specifies a [scheme] and a [hash algorithm]. The scheme decides
+/// the records in the zone will be collected, serialized, and hashed. The
```
mozzieongit, line 40:
```suggestion
/// the way records in the zone will be collected, serialized, and hashed. The
```

```suggestion
/// how the records in the zone will be collected, serialized, and hashed. The
```

src/new/rdata/zonemd.rs
```diff
@@ -0,0 +1,421 @@
...
+            .field("digest", &DigestFmt(self.digest.as_ref()))
+            .finish()
+    }
+}
+
+//--- Cloning
+
+// NOTE: We cannot implement this for all 'Digest: UnsizedCopy' as it would
+// conflict with 'impl<T: Clone> Clone for Box<T>', as 'ZoneMD<Digest>' could
+// also implement 'Clone'.
```
mozzieongit, line 238: I don't fully understand this comment. Which impl block can we not write?

src/new/rdata/zonemd.rs
```diff
@@ -0,0 +1,421 @@
...
+
+impl<Digest: ?Sized + BuildBytes + Ord> CanonicalRecordData
+    for ZoneMD<Digest>
+{
+    fn cmp_canonical(&self, other: &Self) -> Ordering {
+        u32::from(self.serial)
+            .cmp(&other.serial.into())
+            .then(self.scheme.cmp(&other.scheme))
+            .then(self.hash_alg.cmp(&other.hash_alg))
+            .then_with(|| self.digest.cmp(&other.digest))
```
mozzieongit, line 256:
```suggestion
            .then(self.digest.cmp(&other.digest))
```

Or does it need to be a function?

src/new/rdata/rp.rs
```diff
@@ -0,0 +1,232 @@
...
+    RType,
+};
+
+//----------- Rp -------------------------------------------------------------
+
+/// Identification of the person/party responsible for this domain.
+///
+/// An [`Rp`] record identifies the person or party managing a domain. It is
+/// a useful backup when standard means of communication fail, e.g. e-mails to
+/// `postmaster@` are being dropped due to mail server misconfiguration.
```
mozzieongit, line 21: Do you also want to mention that it's good in cases where the postmaster is not the server admin (or in direct contact with) and it's easier to contact the server admin directly?

src/new/rdata/rp.rs
```diff
@@ -0,0 +1,232 @@
...
+    ///
+    /// The first label here is the username (i.e. local part) of the e-mail
+    /// address, and the remaining labels make up the mail domain name. For
+    /// example, <hostmaster@sri-nic.arpa> would be represented as
+    /// `hostmaster.sri-nic.arpa`. This convention is specified in [RFC 1034,
+    /// section 3.3].
+    ///
+    /// [RFC 1034, section 3.3]: https://datatracker.ietf.org/doc/html/rfc1034#section-3.3
+    ///
+    /// This is an optional field; if this field refers to the root zone, no
```
mozzieongit, line 118:
```suggestion
    /// This is an optional field; if this field refers to the root domain name (`.`), no
```

src/new/rdata/rp.rs
```diff
@@ -0,0 +1,232 @@
...
+    /// A domain providing [`Txt`] records for human inspection.
+    ///
+    /// [`Txt`]: crate::new::rdata::Txt
+    ///
+    /// This should lie outside the name owning the record, so that it can be
+    /// used if the domain is malfunctioning. The referenced domain should
+    /// provide human-readable [`Txt`] records which explain how to reach the
+    /// responsible person/party.
+    ///
+    /// This is an optional field; if this field refers to the root zone, no
```
mozzieongit, line 131:
```suggestion
    /// This is an optional field; if this field refers to the root domain name (`.`), no
```

src/new/rdata/rp.rs
```diff
@@ -0,0 +1,232 @@
...
+    ) -> Result<&'b mut [u8], TruncationError> {
+        let bytes = self.mailbox.build_lowercased_bytes(bytes)?;
+        let bytes = self.texts.build_lowercased_bytes(bytes)?;
+        Ok(bytes)
+    }
+
+    fn cmp_canonical(&self, other: &Self) -> Ordering {
+        self.mailbox
+            .cmp_lowercase_composed(&other.mailbox)
+            .then_with(|| self.texts.cmp_lowercase_composed(&other.texts))
```
mozzieongit, line 174: Same question as https://github.com/NLnetLabs/domain/pull/623/changes#r2973740085

mozzieongit (changes requested):

src/new/rdata/rp.rs
Philip-NLnetLabs: Nobody uses RP. We need to support it because it still exists in some zones. Don't spend too much time on documenting RP.

src/new/base/record.rs
```diff
@@ -296,6 +299,18 @@ impl RType {
...
     pub const NSEC3PARAM: Self = Self::new(51);
+
+    /// The type of a `Cds` record.
+    pub const CDS: Self = Self::new(59);
+
+    /// The type of a `CDNSKey` record.
+    pub const CDNSKEY: Self = Self::new(60);
+
+    /// The type of a [`ZoneMD`](crate::new::rdata::ZoneMD) record.
+    pub const ZONEMD: Self = Self::new(63);
```
tertsdiepraam, line 310: Why does this have a link but the other ones don't? Are those records not there yet?

tertsdiepraam (commented):
Only one question in addition to Jannik's comments.

src/new/base/record.rs
bal-e: Yeah, the CDS and CDNSKEY types haven't been implemented yet, but we need their RTYPE codes in Cascade. I did hope to implement them in this PR, but I'll get to them later.

macros/src/impls.rs
bal-e: This prevents a default type parameter (e.g. `struct Foo<T = u8>`) from being copied into the impl blocks for that type (i.e. `impl<T> Foo` instead of `impl<T = u8>`).

src/new/base/question.rs
bal-e: We should make one.

src/new/rdata/zonemd.rs
bal-e: I used `.then_with()` because the digest comparison seemed a bit expensive, so it would not be necessary to compute upfront.

src/new/rdata/zonemd.rs
bal-e: `impl<Digest> Clone for Box<ZoneMD<Digest>>`, over all `Digest`. Hence the comment is under the `//--- Cloning` section. I can make it more explicit if that would help.

src/new/rdata/zonemd.rs
mozzieongit: Aahhh, of course. That makes sense.

src/new/rdata/zonemd.rs
mozzieongit: Making it more explicit might be good.

src/new/base/question.rs
bal-e: I've made #640, which I'll update once this PR is merged.

PR #637, opened 2026-04-20
Update dependencies.

partim:


partim:
I switched the resolver to version 3 so we can just delete Cargo.lock in order to get the crate versions we need.

I’m not entirely sure how well this works with caching?

bal-e (approved):
Looks good. I'll make a follow-up PR that resolves some additional issues with our caching system.

Issue #572, opened 2025-09-22
`MessageBuilder::new_vec` and friends should depend on alloc, not std


kotx:
```rs
#[cfg(feature = "std")]
impl MessageBuilder<Vec<u8>> {
    /// Creates a new message builder atop a `Vec<u8>`.
    #[must_use]
    pub fn new_vec() -> Self {
        infallible(Self::from_target(Vec::new()))
    }
}
```

`Vec` can be imported from `alloc` instead of `std`, so this code probably shouldn't be gated to `std`.

soywod:
I had the same feeling when I wanted to use `StreamTarget::new_vec()`, and I was surprised to see that it's behind the `std` feature rather than `alloc`, is there any particular reason for it?

Issue #630, opened 2026-04-16
`unstable-xfr` depends on `unstable-zonetree`


bal-e:
Enabling the `unstable-xfr` feature without `unstable-zonetree` causes compilation failures; the XFR code relies on the `ZoneUpdate` enum. At minimum, we should mark the dependency in `Cargo.toml` so `unstable-zonetree` will be enabled if `unstable-xfr` is needed.

PR #641, opened 2026-04-21
Alloc + no_std: part 1

soywod:
The `no_std` and `alloc` features lack of consistency. A user should be able to `StreamTarget::new_vec()` without `std` but `alloc` features, which is not the case at the moment. I would like to help, but there is a lot to change. I propose to go by step. Here a first PR that:

- Uses `impl core::error` instead  of `std::error`, and remove the `std` feature gate
- Replaced `extern crate std;` by a proper `cfg_attr` on `no_std`
- Removed duplicate `declare_error_trait!`

Refs: https://github.com/NLnetLabs/domain/issues/572, https://github.com/NLnetLabs/octseq/pull/68

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -178,7 +178,7 @@
 //! of unstable features. These will follow proper Semver practice but may
 //! change significantly in releases with breaking changes.
 
-#![no_std]
+#![cfg_attr(not(feature = "std"), no_std)]
 #![allow(renamed_and_removed_lints)]
 #![allow(clippy::unknown_clippy_lints)]
 #![allow(clippy::uninlined_format_args)]
@@ -186,12 +186,9 @@
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
 #[cfg(feature = "alloc")]
-extern crate alloc;
-
-#[cfg(feature = "std")]
 #[allow(unused_imports)] // Import macros even if unused.
 #[macro_use]
-extern crate std;
+extern crate alloc;
 
 // The 'domain-macros' crate introduces 'derive' macros which can be used by
 // users of the 'domain' crate, but also by the 'domain' crate itself.  Within
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -178,7 +178,7 @@
 //! of unstable features. These will follow proper Semver practice but may
 //! change significantly in releases with breaking changes.
 
-#![cfg_attr(not(feature = "std"), no_std)]
+#![no_std]
 #![allow(renamed_and_removed_lints)]
 #![allow(clippy::unknown_clippy_lints)]
 #![allow(clippy::uninlined_format_args)]
@@ -186,9 +186,12 @@
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
 #[cfg(feature = "alloc")]
+extern crate alloc;
+
+#[cfg(feature = "std")]
 #[allow(unused_imports)] // Import macros even if unused.
 #[macro_use]
-extern crate alloc;
+extern crate std;
 
 // The 'domain-macros' crate introduces 'derive' macros which can be used by
 // users of the 'domain' crate, but also by the 'domain' crate itself.  Within
--- a/src/base/charstr.rs
+++ b/src/base/charstr.rs
@@ -1112,8 +1112,7 @@ impl fmt::Display for CharStrError {
     }
 }
 
-#[cfg(feature = "std")]
-impl std::error::Error for CharStrError {}
+impl core::error::Error for CharStrError {}
 
 //------------ FromStrError --------------------------------------------------
 
@@ -1152,8 +1151,7 @@ impl fmt::Display for FromStrError {
     }
 }
 
-#[cfg(feature = "std")]
-impl std::error::Error for FromStrError {}
+impl core::error::Error for FromStrError {}
 
 //------------ PresentationError ---------------------------------------------
 
@@ -1208,8 +1206,7 @@ impl fmt::Display for PresentationError {
     }
 }
 
-#[cfg(feature = "std")]
-impl std::error::Error for PresentationError {}
+impl core::error::Error for PresentationError {}
 
 //============ Testing =======================================================
 
--- a/src/base/header.rs
+++ b/src/base/header.rs
@@ -938,8 +938,7 @@ impl fmt::Display for FlagsFromStrError {
     }
 }
 
-#[cfg(feature = "std")]
-impl std::error::Error for FlagsFromStrError {}
+impl core::error::Error for FlagsFromStrError {}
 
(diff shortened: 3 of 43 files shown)


src/lib.rs
```diff
@@ -178,20 +178,17 @@
...
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
 #[cfg(feature = "alloc")]
-extern crate alloc;
-
-#[cfg(feature = "std")]
 #[allow(unused_imports)] // Import macros even if unused.
 #[macro_use]
-extern crate std;
+extern crate alloc;
```
bal-e: Could you explain why you'd like to make this change?

There is a distinction between removing `#![no_std]` and adding `extern crate std`; the former imports the std prelude (e.g. `Box` and `Vec`), while the latter does not. One could write code implicitly referencing stuff from the std prelude, e.g. `Box`, then run `cargo check` and not notice that their code is std/alloc-dependent. They would have to run `cargo check --no-default-features` (with any other features their code is gated under) to notice that. While we check some of these cases in CI, it's not thorough, and ATM / AFAIK there's no good way to exhaustively test this. I would prefer retaining the `extern crate std` approach since it unconditionally omits the std prelude, and helps us avoid accidental std dependencies.

bal-e (changes requested):
Hi @soywod! I happen to know your work on Himalaya :)

Thanks for making a PR, we're totally on board with the direction here. And thanks for breaking it down into digestible chunks. We're happy to support this use case, and we hope that Domain can serve your needs! FYI we're slowly overhauling the API with `domain::new`, if it already has all the functionality you need then I would suggest trying it out over the new API.

FYI, we're releasing a new major version of Domain shortly (to 0.12), and this PR will likely be merged afterwards. We're happy to make a patch release afterwards if you'd like this PR (and perhaps follow-up ones) in a crates.io release; just let us know.

soywod:
> Thanks for making a PR, we're totally on board with the direction here. And thanks for breaking it down into digestible chunks. We're happy to support this use case, and we hope that Domain can serve your needs!

These are great news, happy to hear that the direction is fine. I plan to use Domain for PIM service discovery, I recently dived into and found that it plays nicely with Pimalaya I/O-free lib pattern.

> FYI we're slowly overhauling the API with `domain::new`, if it already has all the functionality you need then I would suggest trying it out over the new API.

I just tried it, it fits even better. I will go for it! Is it OK for me to prioritize the `alloc` refactor on `new`, so I can use it as possible?

> FYI, we're releasing a new major version of Domain shortly (to 0.12), and this PR will likely be merged afterwards. We're happy to make a patch release afterwards if you'd like this PR (and perhaps follow-up ones) in a crates.io release; just let us know.

Sure, a patch would be wonderful.

---

With the new API, I somehow cannot make DNS query over TCP. I suspect a misusage from my side, but I don't know enough DNS domain to understand why. I just followed the example given in new base inline doc. In the regular API, `StreamTarget` had a `as_dgram_slice` and `as_stream_slice`:

https://github.com/NLnetLabs/domain/blob/6c3f0410780a1f77e9aebe113ecfd0290174736d/src/base/message_builder.rs#L1897-L1914

Any idea?

src/lib.rs
soywod: I have no strong opinion (nor knowledge) about this. I just see that it is a common practice to put `no_std` behind `cfg_attr` instead of gating `extern crate std`. Serde or base64 are good examples. Octseq also does it. I find it just confusing to have a `no_std` (gives information that the lib supports no std) followed by a `extern crate std` (gives the opposite information). While I understand your point about dependencies, if I want to develop a feature based on alloc and no_std, I would just enable those features and it would not compile with the proper imports.

But I will of course follow your instruction as you are the maintainer, plus as I said I don't have any strong opinion about it.

bal-e: Yeah, I'd appreciate if you revert this change, and then I'll approve the PR.

> While I understand your point about dependencies, if I want to develop a feature based on alloc and no_std, I would just enable those features and it would not compile with the proper imports.

Most of the time we are developing code for `std` environments so we don't always remember to check this. There are probably several cases in the codebase where we have accidentally depended on features (e.g. #630). Having `#![no_std]` unconditionally helps us make fewer mistakes :)

bal-e:
> > Thanks for making a PR, we're totally on board with the direction here. And thanks for breaking it down into digestible chunks. We're happy to support this use case, and we hope that Domain can serve your needs!
> 
> I just tried it, it fits even better. I will go for it! Is it OK for me to prioritize the `alloc` refactor on `new`, so I can use it as possible?

Good to hear! Yes, feel free to prioritize `new`.

> Sure, a patch would be wonderful.

Great, we'll make a release soon.

> With the new API, I somehow cannot make DNS query over TCP. I suspect a misusage from my side, but I don't know enough DNS domain to understand why. I just followed the example given in new base inline doc. In the regular API, `StreamTarget` had a `as_dgram_slice` and `as_stream_slice`:
> 
> https://github.com/NLnetLabs/domain/blob/6c3f0410780a1f77e9aebe113ecfd0290174736d/src/base/message_builder.rs#L1897-L1914
> 
> Any idea?

Ah yes! When sending DNS messages over TCP, each message must be preceded by its length (big-endian `u16`, in units of bytes). `StreamTarget` does this for you automatically, but new base doesn't. I'll mention this in the docs on for the new-base `MessageBuilder`. In short, you would call `MessageBuilder::new(&mut buffer[2..], ...)`, build the message as usual, then set the length prefix as `buffer[0..2].copy_from_slice(&(message_len as u16).to_be_bytes());`. You can calculate the message length as `message.as_bytes().len()` using the `AsBytes` trait.

bal-e (approved):
LGTM :) Thanks for contributing!

soywod:
> Ah yes! When sending DNS messages over TCP, each message must be preceded by its length (big-endian `u16`, in units of bytes). `StreamTarget` does this for you automatically, but new base doesn't. I'll mention this in the docs on for the new-base `MessageBuilder`. In short, you would call `MessageBuilder::new(&mut buffer[2..], ...)`, build the message as usual, then set the length prefix as `buffer[0..2].copy_from_slice(&(message_len as u16).to_be_bytes());`. You can calculate the message length as `message.as_bytes().len()` using the `AsBytes` trait.

Thanks for your help, I made it work using the following key steps:

- Pass `&mut buf[2..]` to `MessageBuilder::new()`

- Copy message length at the begining of the buffer:

  ```rust
  let msg: &mut Message = builder.finish();
  let msg_len = msg.as_bytes().len() as u16;
  buf[0..2].copy_from_slice(&msg_len.to_be_bytes());
  ```

- Write the buffer to the stream (and not the message bytes) `stream.write_all(&buf)`

- Parse response message without length prefix `MessageParser::new(&buf[2..n])`

PR #645, opened 2026-04-29
Switch to Rust 2024

bal-e:
Now that our MSRV is past 1.85, we can switch to the 2024 edition. This changes some minor behavior in the language and exposes new lints (both from `check` and `clippy`). In particular, `unsafe` functions now require `unsafe` blocks to call other `unsafe` functions; this is quite a nice change as it exposes more places where safety comments are needed.


partim:
Did rustfmt change with the edition or why are there suddenly all these formatting changes?

(Grumbling protest against using code formatters deleted.)

tertsdiepraam:
Yes it changed a bit. Most noticeably: https://doc.rust-lang.org/edition-guide/rust-2024/rustfmt-version-sorting.html

partim (approved):

PR #648, opened 2026-04-30
Add `to_mnemonic_str` to `Rcode` and `OptRcode`

rossmacarthur:
This method is present on similar types in this crate.


withjannisNLnetLabs:
Thank you very much for the contribution @rossmacarthur.

I've continued this PR in #668. I changed the implementation a bit to avoid the `unsafe` block.
Apart from that I went with your changes.

PR #649, opened 2026-05-02
Bring SRV record to new unstable API

soywod:
This PR adds support for the SRV record to the unstable new API. I took inspiration from the Responsible Person record data type. Documentation is mostly AI-generated, as I do not know enough about the topic to redact a proper one.

### DIFF
--- a/src/new/base/record.rs
+++ b/src/new/base/record.rs
@@ -279,6 +279,9 @@ impl RType {
     /// The type of an [`Aaaa`](crate::new::rdata::Aaaa) record.
     pub const AAAA: Self = Self::new(28);
 
+    /// The type of an [`Srv`](crate::new::rdata::Srv) record.
+    pub const SRV: Self = Self::new(33);
+
     /// The type of a [`DName`](crate::new::rdata::DName) record.
     pub const DNAME: Self = Self::new(39);
 
@@ -361,6 +364,7 @@ impl RType {
                 | Self::PTR
                 | Self::MX
                 | Self::RP
+                | Self::SRV
                 | Self::DNAME
                 | Self::RRSIG
         )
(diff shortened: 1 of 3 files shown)


bal-e:
Thanks for the submission @soywod. Could I ask you to remove the AI-generated documentation? We don't mind writing all of it ourselves.

soywod:
I replaced the main comment by a `TODO` so you can track it. Ready for another review!

soywod:
Any news about this PR? I absolutely don't want to sound pushy, because I know how painful it is to receive such comments, but I just want to know if you considered this PR to be merged with a minor release in the nearest weeks. This PR is the last one remaining from my side, and it blocks my release process. This way I can adjust myself: if no merge or no release planned for now I can still point to the stable API, adjust the code and release. Let me know :slightly_smiling_face:

src/new/base/record.rs
bal-e, line 350: This reference should be upgraded into an intra-doc link; but don't worry about it, I'm happy to patch it in afterwards.

src/new/rdata/srv.rs
```diff
@@ -0,0 +1,168 @@
...
+    pub priority: U16,
+
+    /// The relative weight for selection among targets of equal priority.
+    pub weight: U16,
+
+    /// The TCP/UDP port on which the service is offered.
+    pub port: U16,
+
+    /// The domain name of the target host.
+    pub target: N,
```
bal-e, line 44: It appears the SRV RFC states that name compression should not be used for this. The conventions for such types are to use a concrete domain name instead (specifically `Name`, in place). You can see an example of this in `DName`. This will probably have large knock-on effects to the rest of the PR.

bal-e (changes requested):
I'm sorry it took so long to get back to this! Unfortunately the PR can't be merged as it stands; SRV does not allow name compression and the definition needs to be updated to reflect this. In order to unlock your use case, I'm going to open a quick new PR with the required changes.

soywod:
Superseded by https://github.com/NLnetLabs/domain/pull/688, closing.

PR #650, opened 2026-05-02
Implement missing core Error for ParseError

soywod:
I noticed that `ParseError` does not implement `core::error::Error`, which prevents the error to be chained properly (for example with `thiserror`).


partim:
Thank you for the PR! This looks like an oversight, indeed.

PR #655, opened 2026-05-07
Improve documentation with examples, clarifications and pipeline

withjannisNLnetLabs:
Improve the documentation by including
- [X] a clarification for the `new::base::name::RevName` type.
- [X] an example usage for `new::base::Record` and
- [X] `new::base::name`
- [x] cleanup of documentation warnings and
- [x] a pipeline check
- [x] remove all occurrences of `rg "\[.+\]: Self::"`

### DIFF
--- a/src/base/iana/sshfp.rs
+++ b/src/base/iana/sshfp.rs
@@ -14,12 +14,12 @@ int_enum! {
     /// SSHFP fingerprint type.
     ///
     /// This type selects the digest algorithm used for the fingerprint in the
-    /// [SSHFP] record.
+    /// [`Sshfp`] record.
     ///
     /// For the currently registered values see the [IANA registration]. This
     /// type is complete as of 2025-09-04.
     ///
-    /// [SSHFP]: ../../../rdata/sshfp/index.html
+    /// [`Sshfp`]: crate::rdata::sshfp::Sshfp
     /// [IANA registration]: https://www.iana.org/assignments/dns-sshfp-rr-parameters/dns-sshfp-rr-parameters.xhtml#dns-sshfp-rr-parameters-2
     =>
     SshfpType, u8;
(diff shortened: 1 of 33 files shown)


src/base/iana/sshfp.rs
withjannisNLnetLabs, line 1: Crate links in this file are not resolving. Maybe because they are in the macro.

src/base/zonefile_fmt.rs
withjannisNLnetLabs, line 1: @tertsdiepraam could you help me understand what `PresenationWriter` is? I couldn't figure it out to resolve the link.

src/zonetree/zone.rs
withjannisNLnetLabs, line 1: I was unable to figure out what `Any::downcast()` should point to.

src/base/zonefile_fmt.rs
tertsdiepraam: I think I meant `FormatWriter`. I changed the names and probably forgot to update it.

withjannisNLnetLabs: Thanks! I updated the docs.

src/base/iana/sshfp.rs
```diff
@@ -42,32 +46,44 @@ int_enum_zonefile_fmt_decimal!(SshfpType, "fingerprint type");
...
     (RSA => 1, "RSA")
 
     /// Specified that the DSA algorithm is used. [RFC4255]
+    ///
+    /// [RFC4255]: <https://tools.ietf.org/html/rfc4255>
     (DSA => 2, "DSA")
 
     /// Specified that the ECDSA algorithm is used. [RFC6594]
+    ///
+    /// [RFC6594]: <https://tools.ietf.org/html/rfc6594>
```
tertsdiepraam: I'm still surprised has to be enclosed with `<>`? As far as I know in the `[...]: ` syntax the right-hand side is always a URL.

src/net/server/invoker.rs
```diff
@@ -112,12 +118,16 @@ where
 
     /// Processing a single response stream item.
     ///
-    /// Calls [`process_feedback`] if necessary. Extracts any response for
+    /// Calls [`process_feedback()`] if necessary. Extracts any response for
```
tertsdiepraam: Is this a warning? I know that it's allowed but I'm surpised that rustdoc enforces it. I tend to prefer writing it without parentheses.

tertsdiepraam (approved):
I have two questions/comments but nothing blocking. Nicely done!

tertsdiepraam (approved):
Great stuff @withjannisNLnetLabs!

PR #657, opened 2026-05-15
Fix display of `SecurityAlgorithm` by splitting Display and FromStr macros

tertsdiepraam:


src/base/iana/secalg.rs
```diff
@@ -116,9 +116,13 @@ int_enum! {
...
 
-int_enum_str_with_decimal!(
+int_enum_fromstr_mnemonic!(
     SecurityAlgorithm,
     u8,
     "unknown security algorithm"
 );
+
+// We want SecurityAlgorithm to parse from a mnemonic from compatibility with
+// some zonefiles, but always print as a number.
```
Philip-NLnetLabs: Can you elaborate? Which zone files. I don't see a problem with mnemonics in user input such as config files. But I don't we should parse that in zone files.

tertsdiepraam: https://github.com/NLnetLabs/domain/pull/656 See the original PR.

Philip-NLnetLabs: I didn't realize this. I think the goal of the original PR is completely broken and this should be reverted as quickly as possible.

tertsdiepraam:
Discussed with @withjannisNLnetLabs. He's gonna fix up the docs and set up some tests for this.

PR #658, opened 2026-05-18
Update EDNS(0) Update Lease comment to reflect new standard

Skyb0rg007:
The doc-comment on the Update Lease option mentioned a previous draft standard that had since expired. RFC 9664 has been standardized and now gives official meaning to EDNS(0) option, and the comment is now updated to reflect that.


partim:
Thank you for the PR!

PR #660, opened 2026-05-21
Panic when trying to create an RRset with records that have diffferent TTLs.

Philip-NLnetLabs:
This is not ideal, but the best we can do in short time. It should be fixed by propagating the error up.

### DIFF
--- a/src/dnssec/sign/error.rs
+++ b/src/dnssec/sign/error.rs
@@ -21,6 +21,9 @@ pub enum SigningError {
     /// Cannot create an Rrset from an empty slice.
     EmptyRecordSlice,
 
+    /// Multiple TTL values in RRset.
+    MultipleTtlValues,
+
     // TODO
     Nsec3HashingError(Nsec3HashError),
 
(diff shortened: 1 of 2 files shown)


ximon18:
Works for me, as in it correctly causes my Cascade instance to panic when given the faulty input zone, rather than apparently signing it fine but then having tools like `dnssec-verify` and `ldns-verify-zone` reporting the RRSIG as bogus.

Note that I rebullt all of `domain`, `domain-kmip`, `dnst` and `cascade(d)` and made Cascade use the rebuilt `dnst` when testing this branch locally.

FYI this is the bad input zone I tested with:

```
$ORIGIN example.com.

@ IN 2000 SOA ns.example.com. some\.bloke.example.com. 4 86400 3500 3212 22

@ 3000 IN A 127.0.0.1
@ 3600 IN A 127.0.0.2
```

The panic looked like this:

```
thread 'cascade-worker' (370473) panicked at /home/ximon/.cargo/git/checkouts/domain-9d7a606d026ebb80/8db34f1/src/dnssec/sign/records.rs:441:39:
TTLs should be the same: MultipleTtlValues
```

ximon18 (approved):

PR #662, opened 2026-05-27
Fix deserialize and add testing for the `base::iana` enums

withjannisNLnetLabs:
`serde_json::from_str()` was unable to parse enum values in a JSON if they are represented as an integer instead of a string.

Modifying the `serde::de::Visitor` Trait for the deserialization solved this issue by calling `deserializer.deserialize_any(Visitor(PhantomData))` which calls the appropriate deserializer for the JSON value. In JSON the default is [u64] AFAIK. Thus the function `visit_u64` is implemented which in turn tries to convert the number into the enum associated integer type.

[u64]: https://users.rust-lang.org/t/deserialize-a-number-that-may-be-inside-a-string-serde-json/27318/3

### DIFF
--- a/src/base/iana/mod.rs
+++ b/src/base/iana/mod.rs
@@ -57,3 +57,493 @@ pub mod sshfp;
 pub mod svcb;
 pub mod tlsa;
 pub mod zonemd;
+
+#[cfg(feature = "serde")]
+#[cfg(test)]
+mod test {
+    use crate::base::iana::class::Class;
+    use crate::base::iana::digestalg::DigestAlgorithm;
+    // use crate::base::iana::exterr::ExtendedErrorCode;
+    use crate::base::iana::ipseckey::IpseckeyAlgorithm;
+    use crate::base::iana::ipseckey::IpseckeyGatewayType;
+    use crate::base::iana::nsec3::Nsec3HashAlgorithm;
+    use crate::base::iana::opcode::Opcode;
+    use crate::base::iana::opt::OptionCode;
+    // use crate::base::iana::rcode::OptRcode;
+    // use crate::base::iana::rcode::Rcode;
+    use crate::base::iana::rcode::TsigRcode;
+    use crate::base::iana::rtype::Rtype;
+    use crate::base::iana::secalg::SecurityAlgorithm;
+    use crate::base::iana::sshfp::SshfpAlgorithm;
+    use crate::base::iana::sshfp::SshfpType;
+    use crate::base::iana::svcb::SvcParamKey;
+    use crate::base::iana::tlsa::TlsaCertificateUsage;
+    use crate::base::iana::tlsa::TlsaMatchingType;
+    use crate::base::iana::tlsa::TlsaSelector;
+    use crate::base::iana::zonemd::ZonemdAlgorithm;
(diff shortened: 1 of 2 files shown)


src/base/iana/mod.rs
```diff
@@ -57,3 +57,498 @@ pub mod sshfp;
...
+    use crate::base::zonefile_fmt::ZonefileFmt;
+
+    #[track_caller]
+    fn validate_generic_representation<T>(
+        test_value: T,              // This value is used as desired value
+        display_repr: String,       // Display MUST result in this String
+        debug_repr: String,         // Debug MUST result in this String
+        fromstr_list: &[&str], // `&[&str]` `FromStr`s MUST result in `test_value`
+        zonefile_fmt_value: String, // ZonefileFmt MUST result in this String
+        serde_serialize_value: String, // ZonefileFmt MUST result in this String
```
bal-e: Perhaps move these comments to a doc comment on the whole function? Rust currently doesn't support doc comments on parameters, which is a bit unfortunate.

bal-e: Also, the docs for `serde_serialize_value` don't look right.

bal-e: Also also: maybe consume `&str` for all the arguments, so you don't need to keep calling `.into()` at the call sites?

src/base/serde.rs
```diff
@@ -119,19 +119,19 @@ where
                 write!(f, "a u32 or string")
             }
 
-            fn visit_u32<E: serde::de::Error>(self, v: u32) -> Result<T, E> {
```
bal-e, line 122: IIRC these methods _could_ be visible to the user, if they use `serde` themselves, right? Maybe we should support `visit_{u8,u16,u32,usize}()` as well. Although it would be a frustrating amount of boilerplate.

src/base/serde.rs
```diff
@@ -119,19 +119,19 @@ where
...
-                Ok(T::from(v))
+            fn visit_u64<E: serde::de::Error>(self, v: u64) -> Result<T, E> {
+                match v.try_into() {
+                    Ok(u32_value) => Ok(T::from(u32_value)),
+                    Err(e) => Err(E::custom(format!(
+                        "value too big, expected u32: {e}"
+                    ))),
```
bal-e, line 127: FYI the variable name and error message in here are now incorrect.

bal-e (approved):
Looks really good! I have a few minor notes, but I think this is a great step forward. Thanks @withjannisNLnetLabs!

I wonder if it would make sense to move these test cases to a separate text file one day. Maybe it would be worthwhile if we had more instances of this test, but that would prevent us directly referencing some `const`s.

src/base/iana/mod.rs
withjannisNLnetLabs: changed in [3677548](https://github.com/NLnetLabs/domain/pull/662/commits/3677548df8aece3f4712a6b6851cc3a925bf084b)

src/base/serde.rs
bal-e: Leaving for future work.

bal-e (approved):
LGTM :)

PR #664, opened 2026-05-28
Check remaining length when parsing open-ended RDATA.

partim:
This PR adds checks to the `parse` functions of open-ended RDATA types to make sure they will not exceed the maximum RDATA length of 65,535 bytes.

Specifically, these checks are added to all types that return an error-result of `LongRecordData` in their `new` function. I hope that covers all such types.

Fixes #621.


tertsdiepraam (approved):
Looks good! Glad you did a proper search for all of the types that were missing this. I tried to find more but couldn't find any that you missed.

PR #670, opened 2026-06-01
[new/rdata] Optimize allocation in 'BoxedRecordData::from()'

bal-e:
Ran into this TODO comment while working on something else. This will probably lead to a performance improvement for Cascade!


withjannisNLnetLabs (approved):
LGTM :)

PR #647, opened 2026-04-29
Alloc + no_std: part 2

soywod:
This part 2 is a bit bigger, but changes are small and still reviewable in my opinion. I can split if you want.

- Prevented `serde` to enable `std`, and in a more generic way made sure that `serde` and `std` are 2 separated things
- Bumped `octseq` to `0.6.1` to benefit from https://github.com/NLnetLabs/octseq/pull/68
- Replaced `std` gates by `alloc` ones, when applicable
- Used `core` and `alloc` instead of `std`, when applicable
- Fixed a typo in `src/base/header.rs` where `std` gate was used instead of `rand`

I mention that Claude Code helped me to find references and test. I reviewed every single line.

Refs: https://github.com/NLnetLabs/domain/issues/572 https://github.com/NLnetLabs/domain/pull/641 https://github.com/NLnetLabs/octseq/pull/68

PR #671, opened 2026-06-01
Better `no_std` support

bal-e:
Inspired by #647, I looked for a way to systematically use `core` and `alloc` instead of `std`. Turns out there are Clippy lints for this, so I've enabled them and resolved all instances.


Cal3165:
Hey, maybe this would be a separate pr but it would be great if the features in cargo.toml were updated to match this (std might not be a requirement for the crypto feature for example just alloc)

bal-e:
Hi @Cal3165! Thanks for your interest. Your suggestion makes sense, but our attention is primarily on the `domain::new` API now, and we're stretched a bit thin. Feel free to make a PR!

PR #672, opened 2026-06-02
Documentation with all and default features

withjannisNLnetLabs:
* added CI check to verify build with default features passes without warnings
* fixed current warnings

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
   #
   # Validate that the documentation builds without any issues.
   build-doc:
-    name: Build documentation for validation
+    name: Check docs
     strategy:
       matrix: # "" -> use default features
         features: ["", "--all-features"]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,7 +89,12 @@ jobs:
   # Validate that the documentation builds without any issues.
   build-doc:
     name: Build documentation for validation
+    strategy:
+      matrix: # "" -> use default features
+        features: ["", "--all-features", "--no-default-features"]
     runs-on: ubuntu-latest
+    env:
+      RUSTDOCFLAGS: "-D warnings"
     steps:
 
     # Load the repository.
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -104,8 +104,15 @@ jobs:
         cache: false
 
     # Do the actual formatting check.
-    # `-D warnings` promotes all warnings to errors which leads to failure
-    - name: Build documentation for validation
+    # `-D warnings` promotes all warnings to errors which leads to failure.
+    # No features explicitly enabled.
+    - name: Build documentation for validation (no feature flag)
```
bal-e: Maybe `(no feature flag)` => `(default feature flags)`?

.github/workflows/ci.yml
```diff
@@ -104,8 +104,15 @@ jobs:
...
+      run: |
+        RUSTDOCFLAGS="-D warnings" \
```
bal-e: FYI you can move this to a `env:` stanza in the YAML above. There should be other examples in this file.

bal-e (commented):
Cool! If you think we will want to try more feature flag sets, it may be good to use a matrix job here. You can see examples in other parts of the file.

.github/workflows/ci.yml
```diff
@@ -89,7 +89,12 @@ jobs:
   # Validate that the documentation builds without any issues.
   build-doc:
     name: Build documentation for validation
+    strategy:
+      matrix: # "" -> use default features
+        features: ["", "--all-features"]
```
withjannisNLnetLabs, line 94: I moved it into a matrix, not sure if `""` is the best solution here. Just the most straight forward I could come up with.

withjannisNLnetLabs:
Maybe as soon as #671 is merged we could add `--no-default-features` to the matrix aswell

.github/workflows/ci.yml
bal-e: Yeah, no worries, I think it's a great start :)

.github/workflows/ci.yml
```diff
@@ -103,12 +108,8 @@ jobs:
         toolchain: stable
         cache: false
 
-    # Do the actual formatting check.
-    # `-D warnings` promotes all warnings to errors which leads to failure
     - name: Build documentation for validation
```
bal-e, line 111: Looking at the PR's CI checklist, I think this name is a bit long. Could it be shortened to e.g. "Check docs"?

bal-e (approved):
Looks great! Just one last minor thing.

bal-e (approved):
gogogo :)

PR #676, opened 2026-06-11
Improve `new::base::Serial` with `new()`, `get()` and docs

withjannisNLnetLabs:
Improve documentation and interaction with `Serial` type.

With special emphasis on the fact that the `Serial` type is not solely representing the SOA Serial Number **BUT** all numbers that use the "Serial number arithmetics" with `SERIAL_BITS` of 32.

See more in [RFC1982].

[RFC1982]: https://datatracker.ietf.org/doc/html/rfc1982

### DIFF
--- a/src/new/base/serial.rs
+++ b/src/new/base/serial.rs
@@ -14,6 +14,7 @@ use super::wire::U32;
 ///
 /// [`Serial`] implements the "Serial number arithmetic" defined in [RFC1982]
 /// with a `SERIAL_BITS` value of 32.
+///
 /// [`Serial`] should not be used interchangably or be confused with the SOA
 /// Serial Number, which is a [`Serial`] but not the sole user of "Serial
 /// number arithmetic".
--- a/src/new/base/serial.rs
+++ b/src/new/base/serial.rs
@@ -10,7 +10,15 @@ use super::wire::U32;
 
 //----------- Serial ---------------------------------------------------------
 
-/// A serial number.
+/// Serial number arithmetic.
+///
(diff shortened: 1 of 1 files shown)


src/new/base/serial.rs
```diff
@@ -10,7 +10,15 @@ use super::wire::U32;
 
 //----------- Serial ---------------------------------------------------------
 
-/// A serial number.
+/// Serial number arithmetic.
+///
+/// [`Serial`] implements the "Serial number arithmetic" defined in [RFC1982]
+/// with a `SERIAL_BITS` value of 32.
+/// [`Serial`] should not be used interchangably or be confused with the SOA
```
bal-e, line 18: Maybe add a blank line between this paragraph and the previous? Otherwise they won't be separated in the generated HTML docs.

src/new/base/serial.rs
```diff
@@ -53,9 +73,15 @@ impl Serial {
     /// instead of a [`u32`] because it is easier to understand and implement
     /// a non-negative check versus the upper range check.
     ///
+    /// Section 7 in [RFC1982] states, in particular for the SOA Serial
+    /// Number, but for any Serial number using a "SERIAL_BITS" value of 32:
+    /// "The maximum defined increment is 2147483647 (2^31 - 1)."
```
bal-e, line 79: Nice :)

bal-e (approved):
Nice! In the future we can further elaborate on the rationale behind serial number arithmetic and its limitations (e.g. why `Serial` doesn't implement `Ord`).

PR #678, opened 2026-06-15
Add `Name` conversion and mnemonic Display for `new::base`

withjannisNLnetLabs:
This PR adds `From<>` implementations to work with `domain::base` `Name` types. Additionally easier conversion between `RevNameBuf` and `NameBuf`. And `fmt::Display`for `RType` and `RClass`.

### DIFF
--- a/src/new/base/name/absolute.rs
+++ b/src/new/base/name/absolute.rs
@@ -882,6 +882,12 @@ impl fmt::Display for NameParseError {
 
 // -- Convert from old Name to new::base::NameBuf ----------------------------
 
+/// Upgrade a [`crate::base::Name`] into a
+/// [`crate::new::base::name::absolute::NameBuf`].
+///
+/// # Panics
+///
+/// The [`crate::base::Name`] slice has to contain a valid domain.
 #[allow(unused)]
 pub fn upgrade_name<Octs>(value: &crate::base::Name<Octs>) -> NameBuf
 where
@@ -901,7 +907,7 @@ mod tests {
             crate::base::Name::from_slice(b"\x07example\x03com\x00")
                 .expect("Invalid name");
 
-        let new_name: NameBuf = upgrade_name(&old_name);
+        let new_name: NameBuf = upgrade_name(old_name);
         assert_eq!(old_name.as_slice(), new_name.as_bytes())
     }
 }
--- a/src/new/base/name/absolute.rs
+++ b/src/new/base/name/absolute.rs
@@ -23,7 +23,7 @@ use crate::{
 
 use super::{
     CanonicalName, Label, LabelBuf, LabelIter, LabelParseError,
-    NameCompressor,
+    NameCompressor, RevNameBuf,
 };
 
 //----------- Name -----------------------------------------------------------
@@ -99,6 +99,11 @@ impl Name {
     pub const fn labels(&self) -> LabelIter<'_> {
         // SAFETY: A 'Name' always contains valid encoded labels.
         unsafe { LabelIter::new_unchecked(self.as_bytes()) }
+    }
+
+    /// Covert &[`Name`] into [`RevNameBuf`]
+    pub fn to_revname(&self) -> RevNameBuf {
+        NameBuf::copy_from(self).into()
     }
 }
 
@@ -909,5 +914,14 @@ mod tests {
 
         let new_name: NameBuf = upgrade_name(old_name);
         assert_eq!(old_name.as_slice(), new_name.as_bytes())
+    }
+
+    #[test]
+    fn test_to_revname() {
+        let namebuf: NameBuf = "example.com".parse().unwrap();
+        assert_eq!(
+            (&namebuf).to_revname().as_bytes(),
+            b"\x00\x03com\x07example",
+        );
     }
 }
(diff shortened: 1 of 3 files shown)


src/new/base/record.rs
```diff
@@ -489,6 +527,22 @@ impl RClass {
...
+        }
+    }
+}
+
+impl From<RClass> for u16 {
+    fn from(value: RClass) -> Self {
+        value.code.get()
+    }
+}
+
```
withjannisNLnetLabs, line 555: This is the same as for `RType` but because it was missing I copied from there.

src/new/base/name/absolute.rs
```diff
@@ -879,3 +884,45 @@ impl fmt::Display for NameParseError {
...
 }
+
+// -- Convert from old Name to new::base::NameBuf ----------------------------
+
+/// Upgrade a [`crate::base::Name`] into a
+/// [`crate::new::base::name::NameBuf`].
+///
+/// # Panics
+///
+/// The [`crate::base::Name`] slice has to contain a valid domain.
```
bal-e: It's good to document this, but isn't this guaranteed by `crate::base::Name` already? All of its constructors verify validity or require the caller to do so.

src/new/base/record.rs
```diff
@@ -447,6 +447,44 @@ impl fmt::Debug for RType {
...
 }
 
+/// Return the associated name of [`RType`]. If [`RType`] is unknown,
+/// then the returned string contains the type in the unknown format as
+/// defined in Section 5 in [RFC3597].
+///
+/// The names are consolidated by [IANA].
+///
+/// [RFC3597]: https://datatracker.ietf.org/doc/html/rfc3597#section-5
+/// [IANA]: https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml
```
bal-e: Yay for documentation :) This is already awesome, but if you would like, feel free to add a few examples (of both mnemonic and unknown cases). They will automatically get run as tests so we don't need to worry about them getting outdated :D

src/new/base/record.rs
```diff
@@ -447,6 +447,44 @@ impl fmt::Debug for RType {
     }
 }
 
+/// Return the associated name of [`RType`]. If [`RType`] is unknown,
+/// then the returned string contains the type in the unknown format as
+/// defined in Section 5 in [RFC3597].
```
bal-e: RustDoc shows the first Markdown paragraph as a brief one-line description. To support that, could you separate the first sentence from the rest? Also the term "associated name" can be unclear ("mnemonic" would be better); to be suitably concise, it might be easiest for the one-line description to be e.g. "Format an [`RType`] in a human-readable way", and to clarify what that means in the later paragraph (and with examples).

bal-e: Likewise for `RClass`.

bal-e (commented):
I have a few comments, but overall this looks great, thanks :) And yay, tests!!

src/new/base/name/absolute.rs
withjannisNLnetLabs: No objections on removing it. Do you have a rule for when you see the need to document it and when not?

bal-e: My rule of thumb is: if the caller can cause the panic to happen, document it. In this case, `crate::base::Name` guarantees that it always holds a valid domain name; if someone tries to construct it incorrectly, they will encounter a panic already, or misuse an `_unchecked` method which says "misusing this can cause undefined behavior". So the caller's mistake would not be that they called your method incorrectly; their mistake would be calling a constructor method on `crate::base::Name` incorrectly. So you don't have to document it here.

src/new/base/name/reversed.rs
```diff
@@ -666,3 +671,47 @@ impl<'a> serde::Deserialize<'a> for std::boxed::Box<RevName> {
...
+    #[test]
+    fn test_upgrade_revnamebuf() {
+        let old_name =
+            crate::base::Name::from_slice(b"\x07example\x03com\x00")
+                .expect("Invalid name");
+
+        let new_name: RevNameBuf = old_name.into();
+        let mut buf = vec![0; new_name.built_bytes_size()];
+        new_name.build_bytes(&mut buf).unwrap();
+        assert_eq!(old_name.as_slice(), buf);
```
bal-e, line 707: FYI this code is `alloc` dependent and will break under the upcoming `no_std` PR. Could you either 1) avoid `vec!` and serialize into `[0u8; 255]`; or 2) add a `#[cfg(feature = "alloc")]` on the `fn` and use `alloc::vec![...]`?

src/new/base/record.rs
```diff
@@ -447,6 +447,54 @@ impl fmt::Debug for RType {
...
+///
+/// The mnemonics are consolidated by [IANA].
+///
+/// ```
+/// # use domain::new::base::RType;
+/// // Known RType with mnemonic
+/// assert_eq!("A", format!("{}", RType::A));
+/// // Unknown RType
+/// assert_eq!("TYPE265", format!("{}", RType::from(265)));
+/// ```
```
bal-e, line 464: Awesome!

bal-e (approved):
Great job :)

src/new/base/name/reversed.rs
withjannisNLnetLabs: I chose option 2 because I think it's more appropriate to only allocate what I really need by using the `built_bytes_size()` function.

PR #680, opened 2026-06-18
Add catalog zone support behind unstable-catalog feature

davidgroves:
This is WORK IN PROGRESS, and is TOTALLY UNTESTED outside for some synthetic unit tests and the bare minimum support for my Cascade testing. Both this and the Cascade feature need considerably more work.

Introduce a new `catalog` module implementing RFC 9432 catalog zones, gated behind the `unstable-catalog` feature (which enables `unstable-zonetree`).

This is used by https://github.com/NLnetLabs/cascade/pull/827 to implement catalog zone support in Cascade.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,6 +97,7 @@ zonefile    = ["bytes", "serde", "std"]
 
 # Unstable features
 unstable-new = []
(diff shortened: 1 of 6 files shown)


davidgroves:
Withdrawining PR as it no longer meets nlnetlabs policy

PR #682, opened 2026-06-25
[new] move DNS constants to macro implementation

withjannisNLnetLabs:
This PR simplifies the construction of constants i.e. assignments done by IANA.

It introduces a macro `known_values_define` which defines the constants and stores all values and names defined in two lists. These lists are used to implement `from_mnemonic` and `to_mnemonic`, but could be used for other functions as well.

Additionally the `known_values_from_and_to_primitive` macro which simplifies the conversion between the wrapper type and the wrapped primitive type.

The following types have been touched:
- `RType`
- `RClass`
- `QType`
- `QClass`
- `Nsec3HashAlgorithm`
- `SecAlg`
- `ZoneMDScheme`
- `ZoneMDHashAlg`

They are implemented using the above mentioned macros including sensible tests.
All these types now have a uniform `Display` and `Debug` implementation including doctests.

## Interface Changes
- [x] I decided to change the formatting of the Debug output slightly. This means specifically that the output of `RType` and others now also the number of the Type includes.
    Example: `RType::A(1)`, Before: ~`RType::A`~

*Might not be complete.*

## Known Limitation/Imperfection
- The `Debug` output for types that don't have the same constant name as the mnemonic are not copy-pastable into code.
    For example `SecAlg::DSA` (mnemonic) != `SecAlg::DSA_SHA1` (rust constant).


src/new/base/name/absolute.rs
```diff
@@ -107,6 +107,15 @@ impl Name {
...
 
+impl<T: ?Sized> AsMut<T> for Name
+where
+    [u8]: AsMut<T>,
+{
+    fn as_mut(&mut self) -> &mut T {
+        self.0.as_mut()
+    }
+}
+
```
bal-e: This is not sound. `Name` assumes that it always contains a valid domain name, and this method would allow a caller to change that with safe code.

bal-e: What was the use case for this?

src/new/base/name/mod.rs
```diff
@@ -141,6 +141,18 @@ pub trait CanonicalName: BuildBytes + Ord {
...
+    ///
+    /// Changes the characters in the domain name to the lowercased form.
+    fn make_canonical(&mut self)
+    where
+        Self: AsMut<[u8]>,
+    {
+        // A label can be at most 63 bytes. Therefore the length indicator can
+        // never be changed due to a confusion with a uppercase letter.
+        self.as_mut().make_ascii_lowercase();
+    }
```
bal-e: I don't understand the `AsMut` bound here... can't you just write `self.0.make_ascii_lowercase()`?

src/new/base/name/mod.rs
```diff
@@ -234,3 +246,34 @@ impl_canonical_name_for_deref! {
...
+#[cfg(feature = "alloc")]
+#[cfg(test)]
+mod tests {
+
+    use super::Name;
+    use super::NameBuf;
+    use crate::new::base::name::CanonicalName;
+
+    #[test]
+    fn test_make_canonical_ref_name() {
```
bal-e: A hooray for tests :)

src/new/base/record.rs
```diff
@@ -11,6 +11,85 @@ use super::wire::{
...
+/// From for Enum Type implementation
+///
+/// This macro implements conversions from the primitive type into the enum
+/// type and vice versa.
+///
+/// Uses existing struct with a `::new()` function to implement:
+/// - fn from(value: $inttype) -> $enumtype
+/// - fn from(value: $enumtype) -> $inttype
+#[macro_export]
+macro_rules! enum_type_from_and_to_primative {
```
bal-e: Typo: "primative" => "primitive"

src/new/base/question.rs
```diff
@@ -319,23 +317,55 @@ impl QClass {
...
-
+    (IN => 1, "IN")
     /// The type of queries in the CHAOS class.
-    pub const CH: Self = Self::new(3);
+    (CH => 3, "CH")
 }
 
+//--- Conversion to and from 'u16'
+
+enum_type_from_and_to_primative!(=> QClass, u16;);
```
bal-e: Missing blank line after this

src/new/base/question.rs
```diff
@@ -319,23 +317,55 @@ impl QClass {
...
-            _ => return write!(f, "QClass({})", self.code),
-        })
+        match self.get_mnemonic() {
+            Some(m) => write!(f, "QClass::{}", m),
+            None => write!(f, "QClass({})", self.code),
+        }
+    }
+}
+
+/// Format an [`QClass`] in a human-readable way
```
bal-e: Nit: "an" => "a"

src/new/base/question.rs
```diff
@@ -319,23 +317,55 @@ impl QClass {
...
+/// ```
+/// # use domain::new::base::QClass;
+/// // Known QClass with mnemonic
+/// assert_eq!("IN", format!("{}", QClass::IN));
+/// // Unknown QClass
+/// assert_eq!("CLASS42", format!("{}", QClass::from(42)));
+/// ```
+///
+/// [RFC3597]: https://datatracker.ietf.org/doc/html/rfc3597#section-5
+/// [IANA]: https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml
```
bal-e: Maybe add a URL anchor pointing to the specific section?

src/new/base/question.rs
```diff
@@ -369,4 +402,38 @@ mod test {
...
+        assert_eq!(number, 6);
+    }
+
+    #[cfg(feature = "alloc")]
+    #[test]
+    fn test_rclass_display() {
+        assert_eq!("IN", format!("{}", QClass::IN));
+        assert_eq!("CH", format!("{}", QClass::CH));
+        assert_eq!("CLASS42", format!("{}", QClass::from(42)));
+    }
```
bal-e: Technically, this is already tested by the example in the docs. I don't know whether we should duplicate it here or not...

src/new/base/record.rs
```diff
@@ -11,6 +11,85 @@ use super::wire::{
     SplitBytes, SplitBytesZC, TruncationError, U16, U32,
 };
 
+//----------- Marco ----------------------------------------------------------
```
bal-e: Typo: "Marco" => "Macro" (maybe "Macros"?)

src/new/base/record.rs
```diff
@@ -11,6 +11,85 @@ use super::wire::{
...
+/// Enum Type implementation
```
bal-e: I think this could be more specific, but we can discuss that.

src/new/base/record.rs
```diff
@@ -11,6 +11,85 @@ use super::wire::{
...
+///
+/// This Macros is used to write boilerplate `match` functions which turn the
+/// Self type into the mnemonic and vice versa.
+///
+/// Uses existing struct with a `::new()` function to implement:
+/// - get_mnemonic()
+/// - from_mnemonic()
+#[macro_export]
+macro_rules! enum_type{
```
bal-e: I'm not sure this is the right place for this macro... maybe a new module under `src/new/base`?

src/new/base/record.rs
```diff
@@ -11,6 +11,85 @@ use super::wire::{
...
+    ( $(#[$attr:meta])* =>
+      $enumtype:ident;
+      $( $(#[$variant_attr:meta])*
+    ( $variant:ident => $value:expr, $mnemonic:expr) )* ) => {
+        // create constants
```
bal-e: I think we can come up with a simpler syntax that would feel closer to normal Rust, let's workshop that together.

src/new/base/record.rs
```diff
@@ -11,6 +11,85 @@ use super::wire::{
...
+            /// Returns mnemonic representation of this type if defined.
+            #[must_use]
+            pub fn get_mnemonic(&self) -> Option<&'static str> {
+                let mnemonic = match self {
+                $(
+                    &$enumtype::$variant => Some($mnemonic),
+                )*
+                    _ => None, // default case if mnemonic is unknown
+                };
+                return mnemonic;
```
bal-e: The `let` statement could be omitted here, the `match` can be the entire function body.

src/new/base/record.rs
```diff
@@ -11,6 +11,85 @@ use super::wire::{
...
+            }
+
+            /// Returns Self if mnemonic is recognised.
+            #[must_use]
+            pub fn from_mnemonic(mnemonic: &str) -> Option<Self> {
+                let rtype = match mnemonic.to_uppercase().as_str() {
```
bal-e: Wait, is `.to_uppercase()` doing a heap allocation to a `String`? That would make this code alloc-dependent. Let's discuss how we can avoid that.

src/new/base/record.rs
```diff
@@ -11,6 +11,85 @@ use super::wire::{
...
+//----------- Marco ----------------------------------------------------------
+/// Enum Type implementation
+///
+/// This Macros is used to write boilerplate `match` functions which turn the
+/// Self type into the mnemonic and vice versa.
+///
+/// Uses existing struct with a `::new()` function to implement:
+/// - get_mnemonic()
+/// - from_mnemonic()
+#[macro_export]
```
bal-e: I don't think these macros should be exported; they are for internal use and can be scoped to the local module.

bal-e (changes requested):
I like the overall idea of using macros for deduplicating some work. I'm curious if they can be applied to more enum types in new-base so the PR removes more lines than it adds.

src/new/base/name/absolute.rs
withjannisNLnetLabs: Yes, totally agree. I used it to prototype the `make_canonical()` function. See `src/new/base/name/mod.rs`

src/new/base/name/mod.rs
withjannisNLnetLabs: No I don't think so. The trait `CanonicalName` could be implemented for everything. So therefore I don't have any guarantees on what the underlying data structure is. This is the reason I impl `AsMut` for `Name` to have an interface, that I can use.

src/new/base/record.rs
withjannisNLnetLabs: No it is not, but I wanted to prototype before I create a new file

src/new/base/record.rs
withjannisNLnetLabs: certainly. I copied the macro signature used by old base.

src/new/base/record.rs
withjannisNLnetLabs: should be fixed now

src/new/base/record.rs
withjannisNLnetLabs: I copied your behaviour from here: [(new-zonefile) record.rs#L461](https://github.com/NLnetLabs/domain/blob/new-zonefile/src/new/base/record.rs#L461)

...after I found this RFC [rust-lang/rfcs/issues/1988](https://github.com/rust-lang/rfcs/issues/1988), which ends without any real suggestion.

src/new/base/record.rs
withjannisNLnetLabs: now in new file.

src/new/base/name/mod.rs
bal-e: Oh, sorry, my bad! I misread the code and thought this method was being added to `Name` / `RevName` directly.

PR #688, opened 2026-06-29
Implement the SRV record type in `new::*` + bugfixes

bal-e:
SRV does not allow name compression, so I sought to define it with `Name` in-place and derive `{Parse,Split}BytesZC` for it. It turned out `Name` did not implement these due to small points of friction in the design; I decided to make the change anyways. I think I ended up fixing a number of silent bugs in `SplitMessageBytes` impls along the way. The SRV definition is pretty basic, and is inspired by the needs of #649.

### DIFF
--- a/src/new/base/parse/mod.rs
+++ b/src/new/base/parse/mod.rs
@@ -259,13 +259,17 @@ impl<'a> ParseMessageBytes<'a> for u8 {
     }
 }
 
-impl<'a, T: ?Sized + ParseBytesZC> ParseMessageBytes<'a> for &'a T {
-    fn parse_message_bytes(
-        contents: &'a [u8],
-        start: usize,
-    ) -> Result<Self, ParseError> {
-        T::parse_bytes_by_ref(&contents[start..])
-    }
+/// Parse bytes from a DNS message, without name compression.
+///
+/// This can be used to parse types that do not contain compressed names
+/// and do not implement [`ParseMessageBytes`]. An explicit conversion
+/// step is needed because some types implement both [`ParseBytes`] and
+/// [`ParseMessageBytes`], but with differing semantics.
+pub fn parse_without_compression<'a, T: ParseBytes<'a>>(
+    contents: &'a [u8],
+    start: usize,
+) -> Result<T, ParseError> {
+    T::parse_bytes(&contents[start..])
 }
 
 impl<'a, T: SplitMessageBytes<'a>, const N: usize> ParseMessageBytes<'a>
@@ -310,7 +314,7 @@ impl<'a> ParseMessageBytes<'a> for alloc::string::String {
         contents: &'a [u8],
         start: usize,
     ) -> Result<Self, ParseError> {
-        <&str>::parse_message_bytes(contents, start).map(|s| s.into())
+        parse_without_compression::<&str>(contents, start).map(|s| s.into())
     }
 }
 
@@ -348,14 +352,20 @@ impl<'a> SplitMessageBytes<'a> for u8 {
     }
 }
 
-impl<'a, T: ?Sized + SplitBytesZC> SplitMessageBytes<'a> for &'a T {
-    fn split_message_bytes(
-        contents: &'a [u8],
-        start: usize,
-    ) -> Result<(Self, usize), ParseError> {
-        T::split_bytes_by_ref(&contents[start..])
-            .map(|(this, rest)| (this, contents.len() - rest.len()))
-    }
+/// Parse a value from the start of a byte sequence within a DNS message,
+/// without name compression.
+///
+/// This can be used to parse types that do not contain compressed names
+/// and do not implement [`SplitMessageBytes`]. An explicit conversion
+/// step is needed because some types implement both [`SplitBytes`] and
+/// [`SplitMessageBytes`], but with differing semantics.
+pub fn split_without_compression<'a, T: SplitBytes<'a>>(
+    contents: &'a [u8],
+    start: usize,
+) -> Result<(T, usize), ParseError> {
+    let (result, rest) = T::split_bytes(&contents[start..])?;
+    let end = contents.len() - rest.len();
+    Ok((result, end))
 }
 
 impl<'a, T: SplitMessageBytes<'a>, const N: usize> SplitMessageBytes<'a>
--- a/src/new/base/question.rs
+++ b/src/new/base/question.rs
@@ -190,6 +190,9 @@ impl QType {
     /// The type of queries for [`Aaaa`](crate::new::rdata::Aaaa) records.
     pub const AAAA: Self = Self::new(28);
 
+    /// The type of queries for [`Srv`](crate::new::rdata::Srv) records.
+    pub const SRV: Self = Self::new(33);
+
     /// The type of queries for [`DName`](crate::new::rdata::DName) records.
     pub const DNAME: Self = Self::new(39);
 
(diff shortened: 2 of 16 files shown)


src/new/base/question.rs
withjannisNLnetLabs, line 200: `SRV` should probably also be included here, right?

src/new/base/parse/mod.rs
```diff
@@ -348,14 +351,19 @@ impl<'a> SplitMessageBytes<'a> for u8 {
...
+/// [`SplitMessageBytes`] is necessary. An explicit conversion step is needed
+/// because some types implement both with traits, with differing semantics.
+pub fn split_without_compression<'a, T: SplitBytes<'a>>(
+    contents: &'a [u8],
+    start: usize,
+) -> Result<(T, usize), ParseError> {
+    let (result, rest) = T::split_bytes(&contents[start..])?;
+    let end = contents.len() - rest.len();
+    Ok((result, end))
 }
```
withjannisNLnetLabs, line 369: Do I understand that correctly. That this function is in particular to split bytes which do not names or if they contain names they do not have name compression within these names?

src/new/base/record.rs
```diff
@@ -131,12 +132,12 @@ where
...
-        let (&rclass, rest) = <&RClass>::split_message_bytes(contents, rest)?;
-        let (&ttl, rest) = <&TTL>::split_message_bytes(contents, rest)?;
+        let (&rtype, rest) = split_without_compression(contents, rest)?;
+        let (&rclass, rest) = split_without_compression(contents, rest)?;
+        let (&ttl, rest) = split_without_compression(contents, rest)?;
         let rdata_start = rest;
-        let (_, rest) =
-            <&SizePrefixed<U16, [u8]>>::split_message_bytes(contents, rest)?;
+        let (_, rest): (&SizePrefixed<U16, [u8]>, _) =
+            split_without_compression(contents, rest)?;
```
withjannisNLnetLabs: So this is a little hard to read: What I understand. You need to know what the slice is in your message where the rdata is exactly is stored. Therefore you run `split_without_compression` for `SizePrefixed` to get the _offset_ to the bytes that are _after_ your slice of rdata.
But why can't you not compute that directly because. Could you not directly compute the end of the rdata by adding `rest + contents[rest] as usize` (considering to convert from U16 to u16 and then to usize).

src/new/base/question.rs
withjannisNLnetLabs, line 275: and here.

src/new/rdata/srv.rs
```diff
@@ -0,0 +1,110 @@
...
+        (*self).unsized_copy_into()
+    }
+}
+
+//--- Equality
+
+impl PartialEq for Srv {
+    fn eq(&self, other: &Self) -> bool {
+        // All elements are compared bytewise.
+        self.as_bytes() == other.as_bytes()
```
withjannisNLnetLabs: what about canonicalizing the name?

src/new/rdata/srv.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+}
+
+impl Eq for Srv {}
+
+//--- Hashing
+
+impl Hash for Srv {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        state.write(self.as_bytes())
```
withjannisNLnetLabs: same here, what about canonicalizing the name before hashing

src/new/rdata/dname.rs
```diff
@@ -150,18 +151,14 @@ impl BuildInMessage for DName {
...
-    fn parse_bytes(bytes: &'a [u8]) -> Result<Self, ParseError> {
-        <&Name>::parse_bytes(bytes).map(DName::new)
-    }
-}
+//--- Parsing from DNS messages
 
-impl<'a> SplitBytes<'a> for &'a DName {
-    fn split_bytes(bytes: &'a [u8]) -> Result<(Self, &'a [u8]), ParseError> {
-        <&Name>::split_bytes(bytes)
-            .map(|(name, rest)| (DName::new(name), rest))
```
withjannisNLnetLabs, line 164: I assume you can delete those lines because dname does not have name compression and so you can replace it with the default trait implementation and the `parse_without_compression` function

src/new/rdata/mod.rs
```diff
@@ -43,6 +43,7 @@
 //! - [`Mx`]
 //! - [`Txt`]
 //! - [`Rp`]
+//! - [`Srv`]
```
withjannisNLnetLabs, line 46: nit: It is a bit confusing that the Type is listed here under `Basic` but is not in the basic directory.

src/new/rdata/srv.rs
```diff
@@ -0,0 +1,110 @@
...
+use crate::new::base::{
+    CanonicalRecordData, ParseRecordData, ParseRecordDataBytes, RType,
+};
+use crate::utils::dst::UnsizedCopy;
+
+//----------- Srv ------------------------------------------------------------
+
+/// The locations of services associated with this domain.
+//
+// TODO: Documentation.
```
withjannisNLnetLabs, line 21: I assume this will have to wait for a future PR.

withjannisNLnetLabs (changes requested):
Minor changes needed, rather readability and nits.

src/new/base/question.rs
bal-e: Yeah, good catch! I keep forgetting about `QType`. Somebody had suggested this earlier: maybe we can find a way to share definitions between the two, or keep them in sync indirectly.

src/new/base/parse/mod.rs
bal-e: Yes! I should add an additional sentence to make this more clear in the docs.

src/new/base/record.rs
bal-e: Yeah, whether or not `SizePrefixed` makes sense here is debatable. I have some ideas for a better way to approach this, let me think a bit.

src/new/rdata/srv.rs
bal-e: That's a good question! `CanonicalRecordData`'s comparison function definitely has to canonicalize the name, and I think I forgot to implement that right now. This might be the first time a new-base record disallows name compression while requiring name canonicalization. Regardless, `PartialEq` does not _have_ to be canonical, but we can choose to make it so. Worth a discussion.

src/new/rdata/srv.rs
bal-e: Our decision for `PartialEq` will inform this, the two have to be in sync.

src/new/rdata/dname.rs
bal-e: These implementations no longer need to be explicitly stated; `DName` now implements `ParseBytesZC` and there is an `impl<T: ParseBytesZC> ParseBytes for &T`.

src/new/rdata/mod.rs
bal-e: That's fair! `Rp` isn't in the `basic` directory either. At the moment `basic` contains types defined in RFC 1035. We should decide whether the source code should be organized by these high-level categorizations, or by RFC.

src/new/rdata/srv.rs
bal-e: Yep, I want to do it properly and that will take time.

src/new/rdata/srv.rs
bal-e: Resolution: we decided to canonicalize the name.

src/new/base/record.rs
withjannisNLnetLabs: I like!

withjannisNLnetLabs (approved):
LGTM!

bal-e:
@soywod, I hope this satisfies your use case. I'm sorry it took so long!

soywod:
Thanks a lot, I will try it out and let you know.

soywod:
Works like a charm! Could you release a minor version by any chance? Thanks a lot for your time and effort :pray:

partim:
> Could you release a minor version by any chance?

Done!

PR #691, opened 2026-07-07
[new] Rework serials in `domain::new`

withjannisNLnetLabs:
TLDR; Improve existing `Serial` type with more declarative types; `Timestamp`, `SoaSerial` and `SeqNumberU32`, which represent the underlying data more appropriately.

In DNS the SOA Serial number and the RRSIG inception/expiration have a 32-bit number which represents them. The maximum number is around 4 billion. The number may overflow and therefore special consideration has to be taken to increment and compare such types. The arithmetical definition for comparison, addition, etc. is in [RFC1982].

---
**`Timestamp`** Represents the time since epoch in seconds. If the number is larger than `u32::MAX` it will wrap around. The types 'unit' is seconds because it's in the nature of the represented data. This type should _not_ be used for normale time and date operation because of its special properties for comparison and modification. For `RRSIG` the value may be represented in the `YYYYMMDDHHmmSS` format see [RFC4034].

> The Signature Expiration Time and Inception Time field values MUST be
   represented either as an unsigned decimal integer indicating seconds
   since 1 January 1970 00:00:00 UTC, or in the form YYYYMMDDHHmmSS in
   UTC, where:

Careful consideration has to be taken when defining the `Display` implementation. Another usage of the type is the `Cookie` used for DNS  messages. This type may be generated from an accurate `Timestamp` (`jiff::Timestamp`), for example after calculating the inception and expiration time of a `RRSIG` record.
- [x] what `Display` format should be used for the `Timestamp`? -> use the seconds representation, because this format is valid in all known cases.

**`SoaSerial`** Represents the version of a zone by being present in the `SOA` records serial field. The version number is mostly used in three different formats; a counter without special meaning, seconds since epoch or the date of the day including a two digit counter in the format `YYYYMMDDxx`. Additionally to the SOA record the type is used in the `ZoneMD` record. The `Display` format is simple because in any case the number representation matches the string representation.

**`SeqNumberU32`** Represents the type described in [RFC1982] with a 32-bit number. This type is used for the above types as the underlying, wrapped data structure. This type should not be used directly but may well be public for usage by the end user.
- [x] define `SeqNumberU32` as public or private? -> public
---

The types must be creatable from the primitive type `u32`. `Timestamp` and `SoaSerial` type have no logical connection and shall not be convertible. The use case for the conversion to the wrapped type is questionable.
- [x] `SoaSerial` /  `Timestamp` convertible into `SeqSequenceU32` and vise versa implemented? yes

---

Desired changes:
- **[new] Add `SeqNumberU32`, `SoaSerial`, `Timestamp` signatures with documentation.**
- **[new] Implementation of `SeqNumberU32`, `SoaSerial`, `Timestamp` functionality.**
- **[new] Add testing for `SeqNumberU32`, `SoaSerial`, `Timestamp` functionality.**
- **[new] Migration of data types to new `SoaSerial`, `Timestamp` types.**

---

> [!IMPORTANT]
> This PR requires Cascade to change, because the types in `Rrisg` and `Soa` change. 

[RFC4034]: https://datatracker.ietf.org/doc/html/rfc4034#section-3.2
[RFC1982]: https://datatracker.ietf.org/doc/html/rfc1982


src/new/base/serial.rs
```diff
@@ -2,12 +2,392 @@
...
+    }
+}
+
+//----------- SeqNumberU32 ---------------------------------------------------
+
+/// Sequence Number u32
+///
+/// This type implements the mathematical properties defined in [RFC1982].
+///
+/// The type is used as a backend for SoaSerial and Timestamp. [RFC1982]
```
tertsdiepraam:
```suggestion
/// The type is used as a backend for [`SoaSerial`] and [`Timestamp`]. [RFC1982]
```

src/new/base/serial.rs
```diff
@@ -2,12 +2,392 @@
...
+        self.0.get()
+    }
+}
+
+impl SeqNumberU32 {
+    /// Increment this by a non-negative number.
+    ///
+    /// The number must be in the range `[0, 2^31 - 1]`. An [`i32`] is used
+    /// instead of a [`u32`] because it is easier to understand and implement
+    /// a non-negative check versus the upper range check.
```
tertsdiepraam: I kind of doubt this claim? Though I guess it makes some sense.

src/new/base/serial.rs
```diff
@@ -2,12 +2,392 @@
...
+    ///
+    /// [RFC1982]: https://datatracker.ietf.org/doc/html/rfc1982#section-3.2
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        let (lhs, rhs) = (self.0.get(), other.0.get());
+
+        if lhs == rhs {
+            Some(Ordering::Equal)
+        } else if lhs.abs_diff(rhs) == 1 << 31 {
+            None
+        } else if (lhs < rhs) ^ (lhs.abs_diff(rhs) > (1 << 31)) {
```
tertsdiepraam: I think this could do with either some explanation or be written out more explicitly. Using the bitwise OR with booleans is a bit surprising, though I think it's correct! Maybe even just a small note about that you're aware that it's bitwise or and not exponentiation already helps 😄

tertsdiepraam (commented):
Just some preliminary comments. Good work!

src/new/base/serial.rs
withjannisNLnetLabs: This function was initially written by @bal-e, therefore I can't speak for the decision making.

src/new/base/serial.rs
withjannisNLnetLabs: I added some usage examples which should illustrate the underlying logic.

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+//----------- SoaSerial ------------------------------------------------------
+
+/// SoaSerial number
+///
+/// The SOA serial number is used in the SOA record defined in [RFC1035].
+///
+/// The serial is used to set the zones version number. It is commonly
+/// displayed as a datetime with counter (YYYYMMDDxx), as seconds since epoch
+/// (1609462861000, "Friday, January 1, 2021 at 1:01:01 AM") or any number
+/// without special meaning.
```
bal-e: Also note the common "counter" convention (which increments on every change).

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+#[repr(transparent)]
+pub struct SoaSerial(SeqNumberU32);
+
+impl SoaSerial {
+    /// New serial from u32
+    pub fn new(value: u32) -> Self {
+        SoaSerial(SeqNumberU32::new(value))
+    }
+
+    /// Get u32 from serial
```
bal-e: "The raw [`u32`] underlying this serial number."

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+    pub fn get(&self) -> u32 {
+        self.0.get()
+    }
+
+    /// Returns a new SoaSerial with an increased serial
```
bal-e: 1. "Increment the serial number by the specified value."
2. Reference the underlying `SeqNumberU32` method, and/or explain its semantics.
3. Does this function panic? What inputs are allowed?

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
 use super::wire::U32;
 
+//----------- SoaSerial ------------------------------------------------------
+
+/// SoaSerial number
+///
+/// The SOA serial number is used in the SOA record defined in [RFC1035].
+///
+/// The serial is used to set the zones version number. It is commonly
+/// displayed as a datetime with counter (YYYYMMDDxx), as seconds since epoch
```
bal-e: If a common representation is "seconds since epoch" which is identical to the semantics of the `Timestamp` type, do we have `From`/`Into` conversions to make that easy?

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+pub struct SoaSerial(SeqNumberU32);
+
+impl SoaSerial {
+    /// New serial from u32
+    pub fn new(value: u32) -> Self {
+        SoaSerial(SeqNumberU32::new(value))
+    }
+
+    /// Get u32 from serial
+    pub fn get(&self) -> u32 {
```
bal-e: `#[must_use] const`

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+    SplitBytes,
+    SplitBytesZC,
+    UnsizedCopy,
+)]
+#[repr(transparent)]
+pub struct SoaSerial(SeqNumberU32);
+
+impl SoaSerial {
+    /// New serial from u32
+    pub fn new(value: u32) -> Self {
```
bal-e: `#[must_use] const`

src/new/base/serial.rs
bal-e: `self` is `Copy`, take by value

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+    }
+
+    /// Get u32 from serial
+    pub fn get(&self) -> u32 {
+        self.0.get()
+    }
+
+    /// Returns a new SoaSerial with an increased serial
+    #[must_use]
+    pub fn inc(&self, value: i32) -> Self {
```
bal-e: IIRC the underlying `inc` is `const`, this can be `const` too

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+    AsBytes,
+    BuildBytes,
+    ParseBytes,
+    ParseBytesZC,
+    SplitBytes,
+    SplitBytesZC,
+    UnsizedCopy,
+)]
+#[repr(transparent)]
+pub struct SoaSerial(SeqNumberU32);
```
bal-e, line 99: Move it to the `soa.rs` file in `rdata`.

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+    AsBytes,
+    BuildBytes,
+    ParseBytes,
+    ParseBytesZC,
+    SplitBytes,
+    SplitBytesZC,
+    UnsizedCopy,
+)]
+#[repr(transparent)]
+pub struct Timestamp(SeqNumberU32);
```
bal-e, line 253: Maybe move this to a dedicated `timestamp.rs` module?

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+    #[must_use]
+    pub fn new(value: u32) -> Self {
+        Timestamp(SeqNumberU32::new(value))
+    }
+
+    /// New timestamp from [`jiff::Timestamp`]
+    #[must_use]
+    pub fn from_timestamp(value: jiff::Timestamp) -> Self {
+        Timestamp::new(value.as_second() as u32)
+    }
```
bal-e: This can be removed in favor of the `From` impl.

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+pub struct Timestamp(SeqNumberU32);
+
+///```
+/// use domain::new::base::Timestamp;
+///
+/// let jiff_ts = jiff::Timestamp::from_second(123_456_789).unwrap();
+/// let ts = Timestamp::from_timestamp(jiff_ts);
+/// assert_eq!(ts.as_secs(), 123_456_789);
+/// assert_eq!(ts, Timestamp::new(123_456_789));
+/// ```
```
bal-e: This should be moved to the type docs and should be preceded by some description of what it does / what it is supposed to be.

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+    ///
+    /// This function expects that the current system time is after UNIX_EPOCH
+    #[cfg(feature = "std")]
+    #[must_use]
+    pub fn unix_time() -> Self {
+        use std::time::SystemTime;
+
+        let time = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("The current system time is before UNIX_EPOCH.");
```
bal-e: Instead of panicking, just figure out the negative duration and use that modulo `2^32`.

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+        Timestamp::new(value.as_second() as u32)
+    }
+
+    /// Get seconds (as u32) from timestamp.
+    #[must_use]
+    pub fn as_secs(&self) -> u32 {
+        self.0.get()
+    }
+
+    /// Measure the current time (in seconds) in serial number space.
```
bal-e: "Measure time since the Unix epoch."

"Determines the current time, in seconds since the Unix epoch, modulo `2^32`."

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+        use std::time::SystemTime;
+
+        let time = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("The current system time is before UNIX_EPOCH.");
+        Self::from(time.as_secs() as u32)
+    }
+}
+
+/// Comparision is forwarded to the underlying [`SeqNumberU32`].
```
bal-e: Add a one-line description of what this impl _is_, e.g. "Ordering for [`Timestamp`]."

Explain why `Ord` is not implemented.

bal-e: Forward to the relevant docs on `SeqNumberU32` (ideally to exactly the same trait impl, for which you need to manually specify the HTML anchor).

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+impl PartialOrd for Timestamp {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        self.0.partial_cmp(&other.0)
+    }
+}
+
+impl fmt::Display for Timestamp {
```
bal-e, line 394: We should discuss what this format should be, and if there is a single "most intuitive" format to use here. In any case, the format should be documented, with usage examples i.e. doc tests.

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+        Timestamp(SeqNumberU32::new(value))
+    }
+
+    /// New timestamp from [`jiff::Timestamp`]
+    #[must_use]
+    pub fn from_timestamp(value: jiff::Timestamp) -> Self {
+        Timestamp::new(value.as_second() as u32)
+    }
+
+    /// Get seconds (as u32) from timestamp.
```
bal-e: "Seconds since the Unix epoch modulo `2^32`."

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+    }
+
+    /// Measure the current time (in seconds) in serial number space.
+    ///
+    /// # Panics
+    ///
+    /// This function expects that the current system time is after UNIX_EPOCH
+    #[cfg(feature = "std")]
+    #[must_use]
+    pub fn unix_time() -> Self {
```
bal-e: Rename to `now()`?

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+impl From<Timestamp> for crate::rdata::dnssec::Timestamp {
+    fn from(ts: Timestamp) -> Self {
+        let v: u32 = ts.into();
+        v.into()
+    }
+}
+
+//----------- SeqNumberU32 ---------------------------------------------------
+
+/// Sequence Number u32
```
bal-e: "A 32-bit unsigned integer in the [RFC1982] sequence number space."

or "A 32-bit unsigned integer using the [RFC1982] sequence number space arithmetic."

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+    AsBytes,
+    BuildBytes,
+    ParseBytes,
+    ParseBytesZC,
+    SplitBytes,
+    SplitBytesZC,
+    UnsizedCopy,
+)]
+#[repr(transparent)]
+pub struct SeqNumberU32(U32);
```
bal-e, line 499: Maybe move this to a `sequence.rs` module? `seqnum.rs`?

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+        SeqNumberU32(U32::new(value))
+    }
+    /// Get u32 from [`SeqNumberU32`]
+    pub fn get(&self) -> u32 {
+        self.0.get()
+    }
+}
+
+impl SeqNumberU32 {
+    /// Increment this by a non-negative number.
```
bal-e: Doc tests?

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+    ///
+    /// The number must be in the range `[0, 2^31 - 1]`. An [`i32`] is used
+    /// instead of a [`u32`] because it is easier to understand and implement
+    /// a non-negative check versus the upper range check.
+    ///
+    /// Section 7 in [RFC1982] states, in particular for the SOA Serial
```
bal-e: Remove reference to "SOA serial"

src/new/base/serial.rs
```diff
@@ -2,12 +2,479 @@
...
+    ///     SeqNumberU32::new(42).partial_cmp(&SeqNumberU32::new(42)),
+    ///     Some(Ordering::Equal)
+    /// );
+    /// assert_eq!(
+    ///     SeqNumberU32::new(42).partial_cmp(&SeqNumberU32::new(43)),
+    ///     Some(Ordering::Less)
+    /// );
+    ///
+    /// // Here the lower absolute number has already wrapped around and thus
+    /// // larger than the `u32::MAX`
```
bal-e: End comments with "."

bal-e (changes requested):

src/new/base/serial.rs
```diff
@@ -1,13 +1,434 @@
...
+/// The [`SoaSerial`] commonly follows one of the following strategies to
+/// declare the [`SoaSerial`]:
+///
+/// - Counter, it starts at 1 and on each change the number gets increased by
+///   one.
+/// - Seconds since Unix Epoch, when the zone is changed the number gets set
+///   to the current number of seconds since Unix Epoch.
+/// - Date including counter, on change the number gets set to the current
+///   date in the format (`YYYYMMDD00`) if the number is smaller than the
+///   previous version, the old serial gets increased by one.
```
bal-e: > if the number is smaller than the previous version, the old serial gets increased by one

I don't think this edge case needs to be mentioned in these high-level docs.

src/new/base/serial.rs
```diff
@@ -1,13 +1,434 @@
...
+}
+
+/// The raw [`u32`] underlying this [`SoaSerial`].
+///
+/// Equivalent to [`SoaSerial::new()`].
+impl From<u32> for SoaSerial {
+    fn from(_value: u32) -> Self {
+        todo!()
+    }
+}
```
bal-e: The top-level descriptions for these two impls appear to be swapped; the first one should say "the raw `u32` underlying this `SoaSerial`." Note that the references to `get()` and `new()` do *not* need to be swapped.

src/new/base/serial.rs
```diff
@@ -1,13 +1,434 @@
...
+
+//----------- Timestamp ---------------------------------------------------
+
+/// Seconds since Unix Epoch modulo 2^32.
+///
+/// The [`Timestamp`] stores the seconds since Unix Epoch modulo 2^32. It is
+/// used in the [`Rrsig`] to keep track of `inception` and `expiration` time
+/// and in the edns [`Cookie`] `timestamp`.
```
bal-e: - "edns" => "EDNS"?
- I see you're referring to the `timestamp` field of `Cookie`; maybe refer to the method directly, i.e. ```[`Cookie::timestamp()`]```?
- The same could be done for the RRSIG fields.

src/new/base/serial.rs
```diff
@@ -1,13 +1,434 @@
...
+/// The mathematical operations are done using the [`SeqNumberU32`]. The
+/// number adheres to the mathematical properties of the "Serial Number
+/// Arithmetics" see [RFC1982] with an unsigned 32 bit integer.
+///
+/// [`Cookie`]: crate::new::edns::Cookie
+/// [`Rrsig`]: domain::new::rdata::Rrsig
+///
+/// [RFC1982]: https://datatracker.ietf.org/doc/html/rfc1982
+/// [RFC4034]: https://datatracker.ietf.org/doc/html/rfc4034#section-3.1.5
+/// [RFC9018]: https://datatracker.ietf.org/doc/html/rfc9018#name-the-timestamp-sub-field
```
bal-e: I'm worried docs added here in the future might reference `[RFC4034]` or `[RFC9018]` without realizing they point to specific sections of those documents. Maybe explicitly state them as `[RFC 4034, section 3.1.5]` and `[RFC 9018, "The Timestamp Sub-Field"]`?

src/new/base/serial.rs
```diff
@@ -1,13 +1,434 @@
...
+
+impl Timestamp {
+    /// Construct a new [`Timestamp`].
+    #[must_use]
+    pub const fn new(_value: u32) -> Self {
+        todo!()
+    }
+
+    /// Underlying seconds since Unix Epoch modulo 2^32.
+    // This type has a specific unit, therefore the `get()` function was
```
bal-e, line 262: Maybe add a blank `//` line before this comment so it is more clearly distinguished from the public docs?

src/new/base/serial.rs
```diff
@@ -1,13 +1,434 @@
...
+    ///
+    /// This method may be used to sort [`Timestamp`] values or to display a
+    /// [`Timestamp`] in a date and time format.
+    ///
+    /// The returned [`SystemTime`] meets the following requirements:
+    ///
+    /// 1) The [`SystemTime`] value has a duration since `UNIX_EPOCH` that
+    ///    modulo `2**32` is equal to our [`Timestamp`] value.
+    /// 2) The time difference between the [`SystemTime`] value and the
+    ///    reference time fits in an [`i32`].
```
bal-e: Can you double-check that RustDoc's Markdown processor is okay with `1)` syntax? I know `1.` works, but I've never tried `1)`. You can try running `cargo doc` locally to check.

bal-e: > the time difference [...] fits in an `i32`

Almost exactly. `i32::MIN` (i.e. `-2^31`) is never produced. It might be worth documenting that.

src/new/base/serial.rs
```diff
@@ -1,13 +1,434 @@
...
+    /// Some(Ordering::Less) -> this < other and (other - this) < 2^31
+    ///                         // 1 < 10 and 10 - 1 < 2^31
+    ///                         or
+    ///                         this > other and (this - other) > 2^31
+    ///                         // u32::MAX > 1 and u32::MAX - 1 > 2^31
+    /// Some(Ordering::Greater) -> this > other and (this - other) < 2^31
+    ///                         // 10 > 1 and 10 - 1 < 2^31
+    ///                         or
+    ///                         this < other and (this - other) > 2^31
+    ///                         // 10 < u32::MAX and u32::MAX - 1 > 2^31
```
bal-e: This should be wrapped in a code block, e.g. ` ```text ... ``` `. Or it can be formatted like a Markdown list, where every bit of code is marked up.

src/new/base/serial.rs
```diff
@@ -1,13 +1,434 @@
...
+}
+
+/// The raw [`u32`] underlying this [`SeqNumberU32`].
+///
+/// Equivalent to [`SeqNumberU32::new()`].
+impl From<u32> for SeqNumberU32 {
+    fn from(_value: u32) -> Self {
+        todo!()
+    }
+}
```
bal-e: I think the first line of the docs are swapped between these.

src/new/base/serial.rs
```diff
@@ -1,13 +1,434 @@
-//! Serial number arithmetic.
+//! Types which operate in the modulo 2^32 number space.
```
bal-e: This description feels a bit too vague to me; maybe "Numeric types using RFC 1982 Sequence Space Arithmetic" would be more precise? I know they are all limited to 32 bits, but that is only because 32-bit implementations are the only common ones; in theory, if we needed e.g. `SeqNumberU16`, I think it would also belong in this module (as it is named right now).

src/new/base/serial.rs
```diff
@@ -37,13 +37,17 @@
...
 #[cfg(feature = "std")]
-use std::time::SystemTime;
+use std::time::{SystemTime, UNIX_EPOCH};
 
 use domain_macros::*;
 
 use super::wire::U32;
 
+/// 2^32 is used to simplify expressions.
+const POW_2_32: u64 = 0x1_0000_0000;
```
bal-e: Maybe define it as `1 << 32`? That feels more obviously correct than a hexadecimal number where you have to count the zeros and multiply by 4.

src/new/base/serial.rs
```diff
@@ -89,38 +93,45 @@ pub struct SoaSerial(SeqNumberU32);
...
     pub fn now() -> Self {
-        todo!()
+        let now = SystemTime::now();
+        let diff = match now.duration_since(UNIX_EPOCH) {
+            Ok(secs_after_epoch) => secs_after_epoch.as_secs() % POW_2_32,
+            Err(secs_before_epoch) => {
+                POW_2_32 - (secs_before_epoch.duration().as_secs() % POW_2_32)
+            }
+        };
+        SoaSerial::new(diff as u32)
```
bal-e: Technically, the `% POW_2_32` is not needed because `diff as u32` implicitly truncates to 32 bits using exactly the same logic. But I like how explicit it is, it's a good reminder that the modulo logic is intentional.

src/new/base/serial.rs
```diff
@@ -235,17 +253,54 @@ impl Timestamp {
...
+        // n * 2**32 for some unknown value of n.
+        //
+        // Epoch                                 Reference              Future
+        // |--------------------------------------- | -----------------------|
+        //                                   [  i32 range   ]
+        //
+        // [   POW_2_32   ][   POW_2_32   ][ timestamp ]
+        //
+        // The goal is to find a [`SystemTime`] which is inside the i32 range.
+        const POW_2_32: u64 = 0x1_0000_0000;
```
bal-e: This constant is already defined at the top of the module and can be removed here.

src/new/base/serial.rs
```diff
@@ -235,17 +253,54 @@ impl Timestamp {
...
+        let ref_secs =
+            reference.duration_since(UNIX_EPOCH).unwrap().as_secs();
+        let k = ref_secs / POW_2_32;
+        let ref_secs_mod = ref_secs % POW_2_32;
+        let ts_secs = self.into_int() as u64;
```
bal-e: Maybe use `as_seconds()` since `into_int()` will be deprecated?

src/new/base/serial.rs
```diff
@@ -235,17 +253,54 @@ impl Timestamp {
...
+        // Epoch                                 Reference              Future
+        // |--------------------------------------- | -----------------------|
+        //                                   [  i32 range   ]
+        //
+        // [   POW_2_32   ][   POW_2_32   ][ timestamp ]
+        //
+        // The goal is to find a [`SystemTime`] which is inside the i32 range.
+        const POW_2_32: u64 = 0x1_0000_0000;
+        let ref_secs =
+            reference.duration_since(UNIX_EPOCH).unwrap().as_secs();
```
bal-e: It looks like this function will panic if the reference time is before the Epoch? Could you document that, or just implement such support? I think the semantics of this function neatly extend for it.

src/new/base/serial.rs
```diff
@@ -404,8 +459,17 @@ impl PartialOrd for SeqNumberU32 {
...
+        let (lhs, rhs) = (self.0.get(), other.0.get());
+        if lhs == rhs {
+            Some(Ordering::Equal)
+        } else if lhs.abs_diff(rhs) == (1 << 31) {
+            None
+        } else if (lhs < rhs) == (lhs.abs_diff(rhs) < (1 << 31)) {
+            Some(Ordering::Less)
+        } else {
+            Some(Ordering::Greater)
+        }
```
bal-e, line 629: Nice, this is a lot more readable than my original code!

src/new/base/serial.rs
```diff
@@ -70,6 +70,14 @@ const POW_2_32: u64 = 0x1_0000_0000;
 /// number adheres to the mathematical properties of the "Serial Number
 /// Arithmetics" see [RFC1982] with an unsigned 32 bit integer.
 ///
+/// Basic operations performed with a [`SoaSerial`].
+/// ```
+/// # use domain::new::base::SoaSerial;
+/// let soa_serial: SoaSerial = SoaSerial::new(u32::MAX).increment();
```
bal-e: Maybe separate the `.new()` and the `.increment()` into separate statements so it's easier for the user to see what they should focus on? With the line as it is right now, the user might miss `new()` and only see `increment()`.

Also, I don't think the `: SoaSerial` type annotation is necessary. It would be if `new()` returned something weirder, but that is not the case here.

src/new/base/serial.rs
```diff
@@ -226,6 +257,15 @@ impl Timestamp {
     }
 
     /// Measure system time since Unix Epoch modulo 2^32.
+    /// ```
```
bal-e, line 278: Maybe add a blank line before this doc test?

src/new/base/serial.rs
```diff
@@ -226,6 +257,15 @@ impl Timestamp {
...
+    /// # use domain::new::base::Timestamp;
+    /// # use std::time::{SystemTime, UNIX_EPOCH};
+    /// let now = Timestamp::now();
+    /// assert_eq!(
+    ///     now.as_seconds(),
+    ///     (UNIX_EPOCH.elapsed().unwrap().as_secs() % 0x1_0000_0000) as u32
```
bal-e: Since `as u32` automatically performs the modulo operation, do you want to omit the `% 0x1_0000_0000`?

Also, this doc test is a little bit racy... is there some way to make it deterministically succeed? I really want to avoid spurious failures.

src/new/base/serial.rs
```diff
@@ -457,7 +526,41 @@ impl PartialOrd for SeqNumberU32 {
     ///                         or
     ///                         this < other and (this - other) > 2^31
     ///                         // 10 < u32::MAX and u32::MAX - 1 > 2^31
+    ///```
```
bal-e: There should be a blank line before the doc test, and a space between `///` and ` ``` `.

src/new/base/serial.rs
```diff
@@ -457,7 +526,41 @@ impl PartialOrd for SeqNumberU32 {
...
+    ///
+    /// # assert_eq!(
+    /// #     SeqNumberU32::new(0).partial_cmp(&SeqNumberU32::new((1 << 31) + 1)),
+    /// #     Some(Ordering::Greater)
+    /// # );
+    ///
+    /// # assert_eq!(
+    /// #     SeqNumberU32::new(0).partial_cmp(&SeqNumberU32::new((1 << 31))),
+    /// #     None
+    /// # );
```
bal-e: I think these belong in a dedicated test function, not in the doc test.

bal-e (changes requested):
This is really cool! Well done @withjannisNLnetLabs! I have a bunch of comments (of course :p), but overall, I really like the functionality and the way you've laid out the code. The commits were nicely organized and easy to review.

Minor nits:

- You often refer to ```the [`SoaSerial`]``` or ```the [`SeqNumberU32`]```; maybe you intend these to be read as "the SOA serial type" and "the `SeqNumberU32` type", but since the word "type" does not appear in there, IMO it's better to omit "the" and just say e.g. ```[`SoaSerial`] commonly follows one of the following strategies ...```.

- There are a few typos here and there, e.g. `Comparision`, `huaman`. I'm not sure if there's a good automated way to detect these, I suspect spellcheck will not deal well with Rust or Markdown.

Top-level feedback:

> They Timestamp and SoaSerial type have no logical connection and shall not be convertible.

I'm not sure that's true; one of the common conventions for the SOA serial is as the number of seconds since the Unix epoch modulo `2^32`, exactly like `Timestamp`. I can see some value in allowing conversions between the two, when users choose to use that convention. To avoid users relying on this convention in the wrong cases, maybe it should be implemented as inherent methods with clear names rather than `From`/`Into`.

> - define SeqNumberU32 as public or private?
> - SoaSerial / Timestamp convertible into SeqSequenceU32 and vise versa implemented?

I think `SeqNumberU32` should be public; it is an implementation of an RFC and so I'd argue it's more than an implementation detail of `SoaSerial` / `Timestamp`. Following that, I think it makes sense to have conversions to and from the wrapped types. These can be simple `From` / `Into` impls (while inherent methods that are `const` would be nice, I don't think there's too much need for them; we can add them if we have a need to).

src/new/base/serial.rs
withjannisNLnetLabs: Linking to the timestamp method is possible, but I think it is not possible to link to certain fields.

src/new/base/serial.rs
withjannisNLnetLabs: I swapped the doc comment and the comment to have the doc comment closer to the function.

src/new/base/serial.rs
withjannisNLnetLabs: Yess, I missed that during cleanup.

src/new/base/serial.rs
withjannisNLnetLabs: > Can you double-check that RustDoc's Markdown processor is okay with 1) syntax? I know 1. works, but I've never tried 1). You can try running cargo doc locally to check.

Yes it does.

> Almost exactly. i32::MIN (i.e. -2^31) is never produced. It might be worth documenting that.

I think I need a bit more insight here.

src/new/base/serial.rs
withjannisNLnetLabs: I downgraded them to code comments I don't think they add meaningful value to the documentation in the form they are at the moment. I put more effort into the though.

PR #693, opened 2026-07-08
Change a debug! log to trace!

tertsdiepraam:
This call got annoying when running Cascade under debug mode.


bal-e (approved):
Makes sense to me; get CI passing and LGTM :)

PR #702, opened 2026-07-20
Fix RP compression bug and add RFC 3597 §4 tests

SebastiaanYN:
`Rp::compose_rdata` called `append_compressed_name()` which violates RFC 3597 §4: only RFC 1035 types may compress rdata names. RP is RFC 1183. Remove the `can_compress()` branch so names always compose without compression.

Add compression tests covering every DNS record type whose rdata contains a domain name, and parser tests ensuring non-RFC 1035 types are still parsed when compressed.

### DIFF
--- a/src/rdata/mod.rs
+++ b/src/rdata/mod.rs
@@ -462,6 +462,9 @@ mod compression_tests {
         );
     }
 
+    /// Compression pointer to `example.com` at question offset 12.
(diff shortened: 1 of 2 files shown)


partim:
Thank you for the PR! You are, of course, correct, RP mustn’t compress.

Also thank for the test case!

partim (approved):

SebastiaanYN:
Thanks for taking a look! :)

PR #705, opened 2026-07-24
[ci] Properly parse Rust nightly versions

bal-e:
The job currently set `nightly-x86_64-unknown-linux-gnu` where it should have been `nightly-2026-07-24`. The former version specification changes meaning every day (night?).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,9 +84,14 @@ on:
 
   # Rebuild 'main' every week. This will account for changes to dependencies
   # and to Rust, either of which can trigger new failures. Rust releases are
-  # every 6 weeks, on a Thursday; this event runs every Friday.
+  # every 6 weeks, on a Thursday; this event runs every Friday. We use this
+  # moment to update CI caches.
   schedule:
     - cron: '0 10 * * FRI'
+
+  # Allow manual execution of CI, primarily for debugging. We consider these
+  # rare and also update CI caches when they occur.
+  workflow_dispatch:
 
 defaults:
   run:
@@ -300,7 +305,7 @@ jobs:
 
     # Restore a cached `target/`.
     - name: Restore from cache
-      if: github.event_name != 'schedule'
+      if: github.event_name != 'schedule' && github.event_name != 'workflow_dispatch'
       id: cache-restore
       uses: actions/cache/restore@v6
       with:
@@ -329,7 +334,7 @@ jobs:
 
     # Note the Rust version in the cache.
     - name: Note Rust version for the cache
-      if: github.event_name == 'schedule'
+      if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
       run: |
         # NOTE: `rustc --version --verbose` also prints a `commit-date`, which
         # is easier to parse, but it is unclear whether it perfectly matches
@@ -347,7 +352,7 @@ jobs:
     # Save to the cache every week.
     - name: Save to the cache
       uses: actions/cache/save@v6
-      if: github.event_name == 'schedule'
+      if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
       with:
         path: |
           ~/.cargo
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -342,14 +347,24 @@ jobs:
...
-        rustup toolchain list --quiet | grep nightly- | tee target/rust-version
+        # NOTE: `rustc --version --verbose` also prints a `commit-date`, which
+        # is easier to parse, but it is unclear whether it perfectly matches
+        # the date used for the nightly version by rustup. The date presented
+        # in the first line seems more likely to be the right field.
+        #
+        # Sample input:
+        # > $ rustc +nightly --version
+        # > rustc 1.98.0-nightly (df6ee909e 2026-06-28)
+        rustc --version \
```
withjannisNLnetLabs: nit: It is slightly confusing that the example includes `+nightly` but the command doesn't.

.github/workflows/ci.yml
withjannisNLnetLabs, line 384: I just noticed that in the clippy version you don't restore from cache under certain conditions (workflow/schedule), but in this version there is no condition. I assume that is missing here? Because this way you would have a clean cache, without artifacts from older versions.

withjannisNLnetLabs: I think the bare version like `2026-07-24` won't work here.

From the documentation here: https://github.com/actions-rust-lang/setup-rust-toolchain#inputs

...I can *not* tell if `nightly-2026-07-24` is accepted as it is in the toolchain file described here: https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

withjannisNLnetLabs (changes requested):
Some nit/questions, but I think the main thing is the rust version specified in `toolchain`.

.github/workflows/ci.yml
bal-e: oh, shoot, you're right. Got too excited :) will fix.

.github/workflows/ci.yml
bal-e: True, I'll remove the `+nightly` from this. We don't need it in this job because the rest setup action set the nightly as the default.

.github/workflows/ci.yml
bal-e: The unconditional restore here was intentional. When building a new cache from scratch for Clippy, we don't want to use the old cache, because it was tied to a particular Rust nightly version. Under the `schedule`/`workflow_dispatch` events, we want to use the latest Rust nightly version, not the one from the cache, so that it doesn't get stale. In that situation, we could choose to restore the cache but still use the latest Rust nightly version; but then the cache would almost certainly be wasted. For *this* job, we don't need to explicitly update the Rust compiler version, and it probably changes very rarely, so restoring the cache is probably helpful.

.github/workflows/ci.yml
```diff
@@ -310,27 +315,14 @@ jobs:
...
-      uses: actions-rust-lang/setup-rust-toolchain@v1
-      with:
-        toolchain: ${{ steps.determine-rust-version.outputs.rust-version }}
-        components: clippy
-        cache: false
-        matcher: false
+      run: |
+        # Determine the version to install. Fall back to `nightly`.
+        [[ -f target/rust-version ]] && : version=$(cat target/rust-version)
+        : version=${version:-nightly}
```
withjannisNLnetLabs: I wasn't able to confirm this logic works.

Thus, I got it to work by removing the `: `...

```bash
[[ -f version.txt ]] && version=$(cat version.txt)
version=${version:-nightly}
echo $version
```

.github/workflows/ci.yml
```diff
@@ -342,14 +334,25 @@ jobs:
...
+        # the date used for the nightly version by rustup. The date presented
+        # in the first line seems more likely to be the right field.
+        #
+        # Sample input:
+        # > $ rustc --version
+        # > rustc 1.98.0-nightly (df6ee909e 2026-06-28)
+        rustc --version \
+          | grep --extended-regexp --only-matching '[0-9]{4}-[0-9]{2}-[0-9]{2}' \
+          | sed 's/.*/nightly-&/'
+          | tee target/rust-version
```
withjannisNLnetLabs, line 350: Missing `\` after `sed 's/.*/nightly-&/'`?

.github/workflows/ci.yml
bal-e: I'll remove the `:`.

withjannisNLnetLabs (approved):

PR #706, opened 2026-07-27
Handle IpseckeyGateway::None branch in hash impl

SebastiaanYN:
Replace the hand-written hash implementation with `#[derive(Hash)]` to remove the `todo!()` from the `None` branch.


partim:
I think `IpseckeyGateway` needs to hash safely on its own without relying on information from `Ipseckey`. This means that the current implementation is also broken. I suspect we can just derive `Hash`?

SebastiaanYN:
Good point, that's a lot easier. I've updated the PR

partim (approved):

partim:
Looks good! Thank for the PR!

PR #719, opened 2026-09-07
Resolve nightly warnings

bal-e:


partim:
I believe the code diff in this PR is changing things. The original loop started as `self.current` but I don’t see a `skip(self.current)` in the new code?

